PR #4, opened 2023-12-05
Add docker-compose.yml file for database dependency

Nitemaeric:
## Context

To avoid using bare metal dependencies or dependency managers, such as `brew`, add PostgreSQL to a `docker-compose.yml` file which can be spun up in docker containers using `docker-compose up`.

## Changes

- Add `docker-compose.yml` file to configure software dependencies, such as PostgreSQL.
- Add `dotenv-rails` gem to configure environment variables in `.env` files.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,2 @@
+POSTGRES_PASSWORD=password
+DATABASE_URL=postgres://postgres:$POSTGRES_PASSWORD@localhost:5432
--- a/Gemfile
+++ b/Gemfile
@@ -75,6 +75,7 @@ end
 group :test, :development do
   gem "brakeman"
   gem "debug", platforms: %i[mri mingw x64_mingw]
+  gem "dotenv-rails"
   gem "factory_bot_rails"
   gem "pry"
   gem "pry-byebug"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -112,6 +112,10 @@ GEM
       irb (>= 1.5.0)
       reline (>= 0.3.1)
     diff-lcs (1.5.0)
+    dotenv (2.8.1)
+    dotenv-rails (2.8.1)
+      dotenv (= 2.8.1)
+      railties (>= 3.2)
     drb (2.2.0)
       ruby2_keywords
     e2mmap (0.1.0)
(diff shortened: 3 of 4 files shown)


JamieCleare2525 (approved):
LGTM!

elceebee (changes requested):
Hey, I'm just confused as to why we are adding dotenv. I'm fine with it, but we've already got a credentials file, right? I think we should chose one or the other.

Nitemaeric:
> Hey, I'm just confused as to why we are adding dotenv. I'm fine with it, but we've already got a credentials file, right? I think we should chose one or the other. (not actually requesting changes, I just want to know what the approach is before we merge.... )

The introduction of dotenv in this PR is mainly so Rails can use the same `$POSTGRES_PASSWORD` and furthermore the `$DATABASE_URL` environment variable without additional configuration, as docker compose reads the `.env` file by default.

I haven't had too much hands-on experience with rails credentials, but the standard of `dotenv` is widely adopted across multiple languages and frameworks.

I'm open to the alternative with rails credentials, but I'm far more familiar with configuration using env variables that aren't shared.

elceebee:
> > Hey, I'm just confused as to why we are adding dotenv. I'm fine with it, but we've already got a credentials file, right? I think we should chose one or the other. (not actually requesting changes, I just want to know what the approach is before we merge.... )
> 
> The introduction of dotenv in this PR is mainly so Rails can use the same `$POSTGRES_PASSWORD` and furthermore the `$DATABASE_URL` environment variable without additional configuration, as docker compose reads the `.env` file by default.
> 
> I haven't had too much hands-on experience with rails credentials, but the standard of `dotenv` is widely adopted across multiple languages and frameworks.
> 
> I'm open to the alternative with rails credentials, but I'm far more familiar with configuration using env variables that aren't shared.

I'm also very happy with dotenv -- that is my preference for the reasons you mention here. But I just want to make sure we aren't accidentally introducing a mixed economy. If we are not going to use credentials, let's delete it so no one is tempted to assume that is where things go.

ollietreend (approved):
This looks good to me.

I've just opened #5 which removes the `credentials.yml.enc` file to address @elceebee's  point.

elceebee (approved):

PR #5, opened 2023-12-05
Remove encrypted credentials.yml file

ollietreend:
This file was introduced in the initial commit (f7b293c), presumably created by the `rails new` command as part of the default Rails setup.

However we don't plan on using Rails' built in credential encryption functionality. Instead we're expecting to pass secrets in to our application using environment variables.

I've removed the file to avoid any confusion.


JamieCleare2525 (approved):
LGTM!

PR #6, opened 2023-12-06
Erb linting implementation

elceebee:
# What
- Added erb linting to the `bin/lint` script

# Change considerations
- The configuration in `.erb-lint.yml` largely mirrors that in [register](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/.erb-lint.yml) and [publish](https://github.com/DFE-Digital/publish-teacher-training/blob/main/.erb-lint.yml) 
- I've disabled `Layout/LineLength` -- I think this this is restrictive in `erb` files where there is often a lot of text and indentation. But I'm happy to change this if other disagree. Or we could change the acceptable line length for erb files to something longer than the standard 140 characters. 
- Publish enforces double quotes in `erb` files, and I've gone with that for our codebase as well for consistency. (publish does not do this)
- I haven't included `better_html` because I'm following the existing patterns in other code bases, but it is part of the recommended `erb_lint` configuration and if others agree, we could include it as well.

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -1,0 +1,26 @@
+---
+EnableDefaultLinters: true
+linters:
+  Rubocop:
+    enabled: true
+    rubocop_config:
+      inherit_from:
+        - .rubocop.yml
+      Layout/ArgumentAlignment:
+        Enabled: false
+      Layout/CommentIndentation:
+        Enabled: false
+      Layout/FirstArgumentIndentation:
+        Enabled: false
+      Layout/FirstArrayElementIndentation:
+        Enabled: false
+      Layout/FirstHashElementIndentation:
+        Enabled: false
+      Layout/InitialIndentation:
+        Enabled: false
+      Layout/LineLength:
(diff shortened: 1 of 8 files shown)


JamieCleare2525 (approved):

elceebee:
[Implement erb linting to itt-mentor-services codebase](https://trello.com/c/p0zIBVQ0/67-implement-erb-linting-to-itt-mentor-services-codebase)

CatalinVoineag:
I have not worked with `better_html`. Are you referring to this [gem](https://github.com/Shopify/better-html)? From a quick read of the readme file it looks sensible and it seems popular and well mantained. I would be OK if we use it

elceebee:
> I have not worked with `better_html`. Are you referring to this [gem](https://github.com/Shopify/better-html)? From a quick read of the readme file it looks sensible and it seems popular and well mantained. I would be OK if we use it

That's the one. I'll have a closer look at it tomorrow and implement it in a separate PR.

PR #8, opened 2023-12-07
Added mail-notify gem

CatalinVoineag:
## Context
We want to setup the `mail-notify` gem so that we can easily send emails to our users. This gem also plays nicely with Rails / ActionMailer

- Application mailer, which includes a generic template id and a basic `notify_email` method to send the email.
- TestMailer where a dummy example exists to show how we can use the `mail-notify`

## Guidance to review

To get the `mail-notify` gem to work we need a few things:
- API keys
- Accounts in https://www.notifications.service.gov.uk/ (I have invited all our team members)
- And at least one template.

I will send the API keys separately, they should be in `.env` files.

- The way this works is we setup a [email template](https://github.com/DFE-Digital/mail-notify#with-a-view) in gov notify and reference that template by its id.
- We setup some basic values like `subject` and `body`. With these variables we have control over the subject and the whole content within our rails application. And because we use erb files our content can be dynamic.

There are 3 types of api keys:
- Test (mock a sending an email)
- Team (can send actual email within the team)
- Live (send emails to anyone)

Documentation can be found on the [gem](https://github.com/DFE-Digital/mail-notify) page or here https://docs.notifications.service.gov.uk/ruby.html#ruby-client-documentation, I know it shows a different gem but I believe they work very similar

## Link to Trello card

https://trello.com/c/YZlPTCFY/28-add-gov-notify-to-allow-us-to-send-emails

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/d9a4b63c-56fc-42a1-97d9-35eee37a61db


CatalinVoineag:
> Not blocking but, we should add this new ENV variable to the `.env.example` file @CatalinVoineag

Agree

elceebee (approved):
I've tested this locally and it works as expected. Nice one!

JamieCleare2525 (approved):

PR #11, opened 2023-12-08
Add documentation for our data model

ollietreend:
## Context

I've taken a first stab at adding documentation for our intended data model. At this point it time, the services encapsulated by this application are in their formative stages, so this data model will undoubtedly change over time.

Right now I consider this to be a 'draft' document, which shows the intended data model for the application.

## Guidance to review

✨ [View the rendered document](https://github.com/DFE-Digital/itt-mentor-services/blob/add-draft-erd/docs/data-model.md) to see the diagram

## Link to Trello card

https://trello.com/c/XH4JTWDv/4-add-the-draft-erd-to-our-code-repo

## Things to check


docs/data-model.md
```diff
@@ -0,0 +1,76 @@
...
+    uuid Mentor_ID PK
+    string TRN FK
+  }
+
+  Provider {
+    uuid Provider_ID PK
+    string Accredited_Provider_ID FK
+  }
+
+  TrainingCompleted {
```
Nitemaeric, line 44: `TrainingCompleted` infers a state of "completed". Do we want to close off this model to that extent right now? Would `Training` with a `completed_at` timestamp be more extendable?

docs/data-model.md
```diff
@@ -0,0 +1,76 @@
...
+There are a few things to bear in mind when reading this:
+
+- This diagram attempts to bridge the gap between a 'high level' list of entities, and a 'low level' database schema. It sits somewhere in between.
+- Attribute names have been written in `Pascal_Case_With_Underscores`. This is to aid readability because spaces are not allowed in Mermaid JS ERD attribute names. It does not represent the format of eventual database column names.
+- It is incomplete. As we continue developing our services, this diagram will undoubtedly change and grow.
+
+```mermaid
+erDiagram
+  Placement {
+    uuid Placement_ID PK
```
Nitemaeric, line 21: Just as a sanity check, are we planning to use `id` as the PK? Is there a requirement to prefix?

docs/data-model.md
```diff
@@ -0,0 +1,76 @@
...
+This application will power two user-facing services: 'Manage school placements' and 'Claim funding for mentors' (a.k.a. Track & Pay). Some data entities will be used by both services – for example, Mentors, Providers and Schools. Others will only be relevant to one service – for example, Funding Claims and School Placements.
+
+## Entity Relationship Diagram (ERD)
+
+This diagram represents our current understanding of the data models that will exist within this application.
+
+There are a few things to bear in mind when reading this:
+
+- This diagram attempts to bridge the gap between a 'high level' list of entities, and a 'low level' database schema. It sits somewhere in between.
+- Attribute names have been written in `Pascal_Case_With_Underscores`. This is to aid readability because spaces are not allowed in Mermaid JS ERD attribute names. It does not represent the format of eventual database column names.
```
Nitemaeric, line 15: Could we use `snake_case` instead? This mirrors Rails' defaults.

docs/data-model.md
elceebee: I tend to agree that `Training` would be a better name -- it works better a plural, too, ie a mentor has many `trainings` rather than `training_completeds`.

docs/data-model.md
SocFoot, line 15: I would rather use rails defaults for naming conventions `Placement_ID`-> `placement_id`. 
`Placement{Placement_ID ...}` feels like it should be `Placement{id}`.

Is the relationship between school and claim not a many to many? Seems like school would be making a claim every year.
Can mentors work for different schools? If so it's a many to many relationship

docs/data-model.md
SocFoot: same

SocFoot (changes requested):
Can you check the relationships?

docs/data-model.md
ollietreend: Yep – good point. The naming of this entity was something we collectively struggled with. I agree that "Completed" is a bit of a clunky work and doesn't really represent the meaning of this data.

I wanted to avoid the ambiguity of using the name "Training" on its own. My concern is that, if you picked this app up without prior context, you could easily mistake "Training" as meaning "Teacher Training". I feel like this is especially likely given we operate within ITT (Initial Teacher Training).

What do you think about "Mentor Training" as a model name, to avoid any ambiguity?
@Nitemaeric @elceebee @SocFoot

elceebee (approved):

Nitemaeric (approved):

docs/data-model.md
elceebee: I think Ollie is right that a Claim will only have one school, but a school can have claims. Even if the mentor moves to a different school, the claim is for to compensate the school for the time they lost when the mentor was in training. Only the one school is relevant to the claim. 

From a user perspective, the school user is the person filling in this form, making a claim on behalf of the school. They won't be making a claim for many schools.

SocFoot (approved):

JamieCleare2525 (approved):

PR #13, opened 2023-12-11
Add host-based routing constraints

Nitemaeric:
## Context

We want to split routing between the "School Placements" and "Track & Pay" services within the Rails application. In production, these URLs are likely to be `manage-school-placements.service.gov.uk` and `claim-funding-for-general-mentors.service.gov.uk`.

- School Placements have confirmed that "Placements" is their desired top-level namespace.
- Track & Pay have confirmed that "Claims" is their desired top-level namespace.

- Add 2 routing files for the separate services, `config/routes/claims.rb` and `config/routes/placements.rb`
- Update `app/views/layouts/application.html.erb` to use service name header based on `current_service` helper.

## Guidance to review

By updating your local `/etc/hosts` file with the following lines, you can access both services without running a proxy server:

```
127.0.0.1 claims.localhost
127.0.0.1 placements.localhost
```

## Link to Trello cards

- https://trello.com/c/Mvow5DdE/50-add-route-constraint-for-track-pay
- https://trello.com/c/US9I2gnJ/49-add-route-constraint-for-school-placements

## Things to check

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,3 +2,6 @@ POSTGRES_PASSWORD=password
 DATABASE_URL=postgres://postgres:$POSTGRES_PASSWORD@localhost:5432
 GOVUK_NOTIFY_API_KEY=
 SIGN_IN_METHOD=persona
+
+PLACEMENTS_HOST=placements.localhost
+CLAIMS_HOST=claims.localhost
--- a/app/controllers/claims/pages_controller.rb
+++ b/app/controllers/claims/pages_controller.rb
@@ -1,0 +1,2 @@
+class Claims::PagesController < ApplicationController
+end
--- a/app/controllers/placements/pages_controller.rb
+++ b/app/controllers/placements/pages_controller.rb
@@ -1,0 +1,2 @@
+class Placements::PagesController < ApplicationController
+end
--- a/spec/requests/pages_spec.rb
+++ b/spec/requests/pages_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 
 RSpec.describe "Pages", type: :request do
   describe "GET /home" do
-    it "returns http success" do
+    xit "returns http success" do
       get "/"
       expect(response).to have_http_status(:success)
     end
--- a/spec/system/home_page_spec.rb
+++ b/spec/system/home_page_spec.rb
@@ -1,0 +1,57 @@
+require "rails_helper"
+
+RSpec.feature "Home Page" do
+  after do
+    Capybara.app_host = nil
+  end
+
+  scenario "User visits the claims homepage" do
+    given_i_am_on_the_claims_site
+    and_i_am_on_the_start_page
+    i_can_see_the_claims_service_name_in_the_header
+    and_i_can_see_the_claims_service_value_in_the_page
+  end
+
+  scenario "User visits the placements homepage" do
+    given_i_am_on_the_placements_site
+    and_i_am_on_the_start_page
+    i_can_see_the_placements_service_name_in_the_header
+    and_i_can_see_the_placements_service_value_in_the_page
+  end
+
+  private
+
+  def given_i_am_on_the_claims_site
+    ENV["CLAIMS_HOST"] = "claims.localhost"
+    Capybara.app_host = "http://#{ENV["CLAIMS_HOST"]}"
+  end
+
+  def given_i_am_on_the_placements_site
+    ENV["PLACEMENTS_HOST"] = "placements.localhost"
+    Capybara.app_host = "http://#{ENV["PLACEMENTS_HOST"]}"
+  end
+
+  def and_i_am_on_the_start_page
+    visit "/"
+  end
+
+  def i_can_see_the_claims_service_name_in_the_header
+    within(".govuk-header") do
+      expect(page).to have_content("Claim Funding for General Mentors")
+    end
+  end
+
+  def i_can_see_the_placements_service_name_in_the_header
+    within(".govuk-header") do
+      expect(page).to have_content("Manage School Placements")
+    end
+  end
+
+  def and_i_can_see_the_claims_service_value_in_the_page
+    expect(page).to have_css(".govuk-summary-list__value", text: "Claims")
+  end
+
+  def and_i_can_see_the_placements_service_value_in_the_page
+    expect(page).to have_css(".govuk-summary-list__value", text: "Placements")
+  end
+end
--- a/spec/system/home_page_spec.rb
+++ b/spec/system/home_page_spec.rb
@@ -22,12 +22,10 @@ RSpec.feature "Home Page" do
   private
 
   def given_i_am_on_the_claims_site
-    ENV["CLAIMS_HOST"] = "claims.localhost"
     Capybara.app_host = "http://#{ENV["CLAIMS_HOST"]}"
   end
 
   def given_i_am_on_the_placements_site
-    ENV["PLACEMENTS_HOST"] = "placements.localhost"
     Capybara.app_host = "http://#{ENV["PLACEMENTS_HOST"]}"
   end
 
--- a/.env.test
+++ b/.env.test
@@ -1,0 +1,2 @@
+PLACEMENTS_HOST=placements.localhost
+CLAIMS_HOST=claims.localhost
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,6 @@
 class ApplicationController < ActionController::Base
+  include ApplicationHelper
+
   default_form_builder(GOVUKDesignSystemFormBuilder::FormBuilder)
 
   helper_method :current_user
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,2 +1,10 @@
 module ApplicationHelper
+  def current_service
+    case request.host
+    when ENV["CLAIMS_HOST"]
+      :claims
+    when ENV["PLACEMENTS_HOST"]
+      :placements
+    end
+  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -7,4 +7,13 @@ module ApplicationHelper
       :placements
     end
   end
+
+  def root_path
+    case current_service
+    when :claims
+      claims_root_path
+    when :placements
+      placements_root_path
+    end
+  end
 end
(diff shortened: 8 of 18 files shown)


.env.example
```diff
@@ -2,3 +2,6 @@ POSTGRES_PASSWORD=password
 DATABASE_URL=postgres://postgres:$POSTGRES_PASSWORD@localhost:5432
 GOVUK_NOTIFY_API_KEY=
 SIGN_IN_METHOD=persona
+
+PLACEMENTS_HOST=
+CLAIMS_HOST=
```
CatalinVoineag, line 7: Should we populate these with something like `placements.localhost:3000`?

app/controllers/claims/pages_controller.rb
```diff
@@ -0,0 +1,2 @@
+class Claims::PagesController < ApplicationController
+end
```
CatalinVoineag, line 2: Don't we need an index controller method here?

app/controllers/placements/pages_controller.rb
```diff
@@ -0,0 +1,2 @@
+class Placements::PagesController < ApplicationController
```
CatalinVoineag, line 1: Same here

spec/system/home_page_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "rails_helper"
```
CatalinVoineag, line 1: Would it be useful to have something like this to test the claims and placements index pages?

CatalinVoineag (commented):
Can you include a video or something of how this should work? I couldn't get it working on my local. It doesn't hit the individual placements and claims controllers and I'm not sure if [this](https://github.com/DFE-Digital/itt-mentor-services/blob/ec841334e7879a1d677e0145f2913b4e808e74a1/config/routes/claims.rb#L1) is correct, depending on what we want the host to be I guess.

Can jump on a call if it's easier

elceebee (approved):
Nice one, I got it to work locally as expected, see images. Good to see some tests as well so we can start establishing some patterns / good practice for that. 

I agree that env vars values should go in the .env.example file unless they are secret. Just helps a new dev get started more quickly. 

image

image

app/controllers/claims/pages_controller.rb
Nitemaeric: No. Rails will default to rendering the view if it finds one. :)

spec/system/home_page_spec.rb
CatalinVoineag: Do you still think there is value in this? ☝🏻

ollietreend (commented):
@Nitemaeric I get three failures when running the test suite locally.

```
$ rake
[...]
Failures:

  1) Pages GET /home returns http success
     Failure/Error: get "/"
     
     ActionController::RoutingError:
       No route matches [GET] "/"
     # ./spec/requests/pages_spec.rb:6:in `block (3 levels) in <main>'

  2) Home Page User visits the claims homepage
     Failure/Error: visit "/"
     
     ActionController::RoutingError:
       No route matches [GET] "/"
     # ./spec/system/home_page_spec.rb:31:in `and_i_am_on_the_start_page'
     # ./spec/system/home_page_spec.rb:6:in `block (2 levels) in <main>'

  3) Home Page User visits the placements homepage
     Failure/Error: visit "/"
     
     ActionController::RoutingError:
       No route matches [GET] "/"
     # ./spec/system/home_page_spec.rb:31:in `and_i_am_on_the_start_page'
     # ./spec/system/home_page_spec.rb:12:in `block (2 levels) in <main>'
```

I know we don't have CI hooked up to the repo yet, but we should try and avoid committing failing tests to main.

spec/system/home_page_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    allow_any_instance_of(ApplicationHelper).to receive(
+      :current_service
+    ).and_return(:claims)
+  end
+
+  def given_i_am_on_the_placements_site
+    allow_any_instance_of(ApplicationHelper).to receive(
+      :current_service
+    ).and_return(:placements)
+  end
```
ollietreend, line 28: Can I suggest something like this instead?

```suggestion
  def given_i_am_on_the_claims_site
    Capybara.app_host = "http://#{ENV["CLAIMS_HOST"]}"
  end

  def given_i_am_on_the_placements_site
    Capybara.app_host = "http://#{ENV["PLACEMENTS_HOST"]}"
  end
```

And then reset the `app_host` to the default value of `nil` in the test teardown:

```ruby
after do
  Capybara.app_host = nil
end
```

[Inspired by StackOverflow](https://stackoverflow.com/a/7050306), this seems to work as expected and allows the tests to pass.

I think this provides a more truthful system test because it covers the route config, and removes the need to stub the `current_service` helper method.

Nitemaeric:
Turns out I was setting `ENV["..._HOST"]` to `http://...` :/

spec/requests/pages_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.describe "Pages", type: :request do
   describe "GET /home" do
-    it "returns http success" do
+    xit "returns http success" do
```
Nitemaeric, line 5: Skipping this test for now. I'm not sure if we'll need to drop it entirely. Might be the case.

spec/system/home_page_spec.rb
Nitemaeric: I'll add in a case to check the page content.

Nitemaeric: Added a temp case while we don't have the login page setup.

spec/system/home_page_spec.rb
```diff
@@ -0,0 +1,53 @@
+require "rails_helper"
+
+RSpec.feature "Home Page" do
+  scenario "User visits the claims homepage" do
+    given_i_am_on_the_claims_site
+    and_i_am_on_the_start_page
```
SocFoot, line 6: Can you remove these type of private methods. I guess there is a config aspect to set the host, maybe use something like
```
RSpec.describe "Widget management", type: :request do
  before { host! "api.example.com" }

  it "creates a Widget" do
    headers = { "ACCEPT" => "application/json" }
    post "/widgets", :params => { :widget => { :name => "My Widget" } }, :headers => headers

    expect(response.content_type).to start_with("application/json")
    expect(response).to have_http_status(:created)
  end
end
```
, or a global config

If you want your test to be more expressive you can make a use of `describe|context|scenario|it`.

Good work

SocFoot (changes requested):

spec/system/home_page_spec.rb
Nitemaeric: I've referenced the existing BAT apps for the system specs.

[Apply example](https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/spec/system/candidate_interface/candidate_content_spec.rb)
[Register example](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/spec/features/landing_page_spec.rb)

That being said, I'm not against the format you're proposing. What do you guys think? @DFE-Digital/track-pay @DFE-Digital/school-placements

elceebee: Weighing in --- I actually prefer the way this is currently written. When we have good ACs on tickets, this pattern maps straight to the ACs, but without the nightmare of using something like cucumber. And if it's the pattern used more widely across the DfE, it makes sense to continue it here. 

Happy to go with the majority if other don't agree, of course.

SocFoot: I didn't realise other repo where doing this. Looks a bit crazy but my main concern is this:

- Usually you want very low amount coding in your test and accept some level of duplication. If you code in your test you now need to test this code... So it's important(to me at least) to be able to spot quickly where is the code(aka magic) which sometimes is ok. This will create a situation where a LOT of functions may or may not do very similar things. 

But if the team feel like it's ok to go with a lot of small functions I will follow.

ollietreend: I think we'd be best to stick with the convention that's already established here in DfE.

For anyone who hasn't seen this style of feature spec before, it's worth pointing out that it does have a small but established following. It's not invented here at DfE.

Here's a blog post describing the style and the reasoning behind it:
_(via the Wayback Machine because the live URL seems to be broken)_
https://web.archive.org/web/20181125211832/https://about.futurelearn.com/blog/how-we-write-readable-feature-tests-with-rspec/

This seems like a nice way to ensure we create readable feature specs.

spec/system/home_page_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def given_i_am_on_the_claims_site
+    ENV["CLAIMS_HOST"] = "claims.localhost"
+    Capybara.app_host = "http://#{ENV["CLAIMS_HOST"]}"
+  end
+
+  def given_i_am_on_the_placements_site
+    ENV["PLACEMENTS_HOST"] = "placements.localhost"
+    Capybara.app_host = "http://#{ENV["PLACEMENTS_HOST"]}"
+  end
```
ollietreend, line 28: ### 1. Redefining environment variable

Do we need to set `ENV["CLAIMS_HOST"] =` and `ENV["PLACEMENTS_HOST"] =` here?

I see you've updated the `.env.example` file too – so I'd expect these environment variables to already be present, without needing to re-define them here.

### 2. Polluting global scope

I think we need a teardown to reset `Capybara.app_host` to `nil` again. Right now it looks like this will leak out and persist in the global scope.

We may want to consider a global setup/teardown and helper methods to set & reset the `Capybara.app_host` value, because this is likely going to be something we use across most feature specs.

elceebee: You may recall that I raised the `.env.test` issue a couple days ago. I sorted it locally by renaming my `.env.local` file to just `.env`. This makes it accessible to both the local and test environments. I think eventually we will want a `.env.test` file as there will be environment that is not shared across test and local, but this works for now.

SocFoot (approved):

spec/system/home_page_spec.rb
ollietreend: Ah – I've been using a `.env` file all along. I didn't realise others had named theirs `.env.local`. 🙈

Yep – all makes sense. It would definitely be nice to have environment variables for tests defined in one place, rather than manually setting them within the test scenarios.

But all good if you'd rather keep it this way for now.

Nitemaeric:
> It appears this breaks logging in as a persona -- `root_path` is not defined.
> 
>  no_root_path.mov

😅 Yup. It's fixed now :)

elceebee (approved):

PR #14, opened 2023-12-11
For discussion: flipflop as a feature flag tool

elceebee:
## Context

This is a basic implementation of the [Flipflop gem](https://github.com/voormedia/flipflop), which we could use as a Feature Flag tool. 

This is for discussion and further work is required for actual implementation.

### Advantages
- The gem is updated regularly, actively maintained
- Very easy to get started and create features
- Very little for developers to maintain
- If we don't want to use the UI at all, we can use rake tasks
- Custom Strategies available -- ie, we could make features available to specific user groups. See the Flipflop documentation. 

### Disadvantages
- UI is not in keeping with the rest of the application (no branding, cannot guarantee accessibility standards) -- not sure this is much of an issue though as it will only be used by members of the product team, mostly developers.
- The other services appear to have home-rolled Feature Flags, so it is possible that users may have different expectations in terms of UI. Again, I'm not sure this is massively important given our user cohort. Also, we could override the default UI if we find it's a problem. 

## Guidance to review

- Review the documentation and advantages / disadvantages to form an opinion about whether or not this would be a good tool to use. 
- I've used Redis as a strategy rather than Active Record / Database as that seems to be the strategy used in other Feature Flags around DfE. But if people feel strongly in another direction, it would be good to hear about it. 

## Link to Trello card

[Feature Flag Investigation](https://trello.com/c/j2HF4dBc)

## Things to check

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/44073106/e6c9a7cf-50fa-46f4-9a40-bf8123515190

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ gem "view_component"
 gem "mail-notify"
 
 # Use Redis adapter to run Action Cable in production
-# gem 'redis', '~> 4.0'
+gem "redis", "~> 4.0"
 
 # Use Kredis to get higher-level data types in Redis [https://github.com/rails/kredis]
 # gem 'kredis'
@@ -44,6 +44,8 @@ gem "bootsnap", require: false
 
 # Use Active Storage variants [https://guides.rubyonrails.org/active_storage_overview.html#transforming-images]
 # gem 'image_processing', '~> 1.2'
+
+gem "flipflop" # Feature flag
 
 gem "govuk-components"
 gem "govuk_design_system_formbuilder", "~> 5.0.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -149,6 +149,9 @@ GEM
     faraday-follow_redirects (0.3.0)
       faraday (>= 1, < 3)
     faraday-net_http (3.0.2)
+    flipflop (2.7.1)
+      activesupport (>= 4.0)
+      terminal-table (>= 1.8)
     globalid (1.2.1)
       activesupport (>= 6.1)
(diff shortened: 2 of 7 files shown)


CatalinVoineag:
This looks good, nice work.

I don't have a lot of opinions on this. The only questions I have are:

Do you know why there's a preference for Redis over the database? Feels much easier to interact with if it's in the DB. Probably because I'm not an expert in Redis.
Can we change the feature flag in the rails console?

Nitemaeric:
I have the same comments as Ollie. The tool looks great! To avoid data loss, let's use Postgres for now.

We'll need to monitor the database performance once this goes live though, as we may be querying multiple flags per page render.

elceebee:
[Feature Flag investigation](https://trello.com/c/j2HF4dBc/32-feature-flag-investigation)

SocFoot:
@elceebee let's not be judgmental, this is probably WAY better than a custom made tool that i know nothing about ;). Still having different way of managing an app in production can add overhead for the team doing it. I am unsure if we will be in charge of turning these feature flags on and off.

- If we are not I think the ultimate decision should go with the team managing them.
- If we are let's document this(even if it's trivial tool) so the next people looking at the project can get up to speed quickly without killing themselves.

### On the use of Redis
Redis is an CP type database and a key value in MEMORY data store. Almost none of these attributes are valuable for feature flags. We want CA on hard drive. The only fact that at scale, if everyone at dfe uses it we might have less than a thousand feature flags, postgres seems like the go to answer.

PR #24, opened 2023-12-12
Add lint and test to CI

SocFoot:
## Context

We currently don't have CI available. This pull request goal is to add a lint step and test step to our CI via github action.

NONE

## Guidance to review

CI should run be green.

## Link to Trello card

https://trello.com/c/G2AoTkQi/52-setup-github-actions-and-basic-ci

## Things to check

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,101 @@
+name: Build and Deploy
+concurrency: build_and_deploy_${{ github.ref_name }}
+
+on: push
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        tests: [rubocop, erblint, prettier]
+        include:
+          - tests: rubocop
+            command: bundle exec rubocop
+          - tests: erblint
+            command: bundle exec erblint app
+          - tests: prettier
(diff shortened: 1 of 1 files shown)


CatalinVoineag (approved):
Nice work, we really need this. Do you think we can print the seed in the logs when we run rspec? If we ever get any flaky tests it's easier to replicate it on local with the seed

JamieCleare2525 (approved):
Nice work!

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,100 @@
...
+          DATABASE_URL: postgres://postgres:password@localhost:5432
+          GOVUK_NOTIFY_API_KEY: fake_key
+          SIGN_IN_METHOD: persona
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    services:
+        postgres:
+          image: postgres:13.5
```
elceebee, line 56: Is there a reason we are using a different version of postgres here? In the dockerfile, we have it is at 16.

I'm sure it won't make a difference to the tests, but I think it would be good to make it consistent if possible.

elceebee (commented):
This is great, nice work! 
Not a blocker from me, but would be good to use the same version of postgres as we have defined in the dockerfile (16) unless there is some reason not to.

PR #27, opened 2023-12-13
Import Gias

CatalinVoineag:
## Context

We need to import GIAS data into our app. The GIAS service does not have
an official API. What they do have is a CSV that they publish every day
and we can download this CSV and parse the data.

To do this, we have introduced a rake task `rake gias_update` which will
download the CSV file for that day in the tmp folder and upsert the data
into our DB. All in all we have about 24k GiasSchool records created
containing schools and MATS that are still open. The file will be removed
after the task is done.

- New rake task
- New ENV variable to hold the GIAS URL
- CSV Import
- Generic Service pattern

## Guidance to review

To run this on local:
1. bundle install
2. copy `GIAS_CSV_BASE_URL` to your env file
3. rake gias_update

## Link to Trello card

https://trello.com/c/5t52nESP/53-import-gias-schools

## Things to check

## Change considerations
- The task takes about 20 seconds on my machine, so if we put this into production hopefully it won't take too long and there's only 1 db call.
- Rerunning the job after you import the data will not update any records in your db that have not changed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,6 +2,6 @@ POSTGRES_PASSWORD=password
 DATABASE_URL=postgres://postgres:$POSTGRES_PASSWORD@localhost:5432
 GOVUK_NOTIFY_API_KEY=
 SIGN_IN_METHOD=persona
-
 PLACEMENTS_HOST=placements.localhost
 CLAIMS_HOST=claims.localhost
+GIAS_CSV_BASE_URL=https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public
--- a/app/models/gias_school.rb
+++ b/app/models/gias_school.rb
@@ -1,0 +1,27 @@
+# == Schema Information
+#
+# Table name: gias_schools
+#
+#  id         :uuid             not null, primary key
+#  address1   :string
+#  address2   :string
+#  address3   :string
+#  name       :string           not null
+#  postcode   :string
+#  telephone  :string
+#  town       :string
+#  ukprn      :string
+#  urn        :string           not null
+#  website    :string
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+# Indexes
+#
+#  index_gias_schools_on_urn  (urn) UNIQUE
+#
+class GiasSchool < ApplicationRecord
+  validates :urn, presence: true
+  validates :urn, uniqueness: { case_sensitive: false }
+  validates :name, presence: true
+end
--- a/app/services/service_pattern.rb
+++ b/app/services/service_pattern.rb
@@ -1,0 +1,16 @@
+module ServicePattern
+  def self.included(base)
+    base.extend ClassMethods
+    base.class_eval { private_class_method :new }
+  end
+
+  def call
+    raise NotImplementedError("#call must be implemented")
+  end
+
+  module ClassMethods
+    def call(...)
+      new(...).call
+    end
+  end
+end
--- a/db/migrate/20231213110952_create_gias_schools.rb
+++ b/db/migrate/20231213110952_create_gias_schools.rb
@@ -1,0 +1,18 @@
+class CreateGiasSchools < ActiveRecord::Migration[7.1]
+  def change
+    create_table :gias_schools, id: :uuid do |t|
+      t.string :urn, null: false, index: { unique: true }
+      t.string :name, null: false
+      t.string :postcode
+      t.string :town
+      t.string :ukprn
+      t.string :telephone
+      t.string :website
+      t.string :address1
+      t.string :address2
+      t.string :address3
+
+      t.timestamps
+    end
+  end
+end
--- a/lib/tasks/gias_update.rake
+++ b/lib/tasks/gias_update.rake
@@ -1,0 +1,15 @@
+desc "Upsert GIAS data into GiasSchool"
+task gias_update: :environment do
+  logger = Logger.new($stdout)
+  today = Time.zone.today.strftime("%Y%m%d")
+  gias_filename = "edubasealldata#{today}.csv"
+
+  logger.info "Downloading the new gias file for #{Time.zone.today}"
+  tempfile = Down.download("#{ENV["GIAS_CSV_BASE_URL"]}/#{gias_filename}")
+  logger.info "Done!"
+
+  logger.info "Importing data"
+  GiasCsvImporter.call(tempfile.path)
+
+  tempfile.unlink
+end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -20,7 +20,7 @@ RSpec.describe User, type: :model do
 
   describe "validations" do
     it { is_expected.to validate_presence_of(:email) }
-    it { is_expected.to validate_presence_of(:email) }
+    it { is_expected.to validate_uniqueness_of(:email).case_insensitive }
     it { is_expected.to validate_presence_of(:first_name) }
     it { is_expected.to validate_presence_of(:last_name) }
   end
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .envrc
 .idea
 .vscode
+.byebug_history
 config/master.key
 db/*.sqlite3
 node_modules
--- a/Gemfile
+++ b/Gemfile
@@ -58,6 +58,9 @@ gem "omniauth-rails_csrf_protection"
 # Decorators
 gem "draper"
 
+# Download files safely
+gem "down", "~> 5.0"
+
 group :development do
   gem "annotate", require: false
   gem "prettier_print", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -130,6 +130,8 @@ GEM
     dotenv-rails (2.8.1)
       dotenv (= 2.8.1)
       railties (>= 3.2)
+    down (5.4.1)
+      addressable (~> 2.8)
     draper (4.0.2)
       actionpack (>= 5.0)
       activemodel (>= 5.0)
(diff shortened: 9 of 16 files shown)


.env.example
```diff
@@ -2,3 +2,4 @@ POSTGRES_PASSWORD=password
 DATABASE_URL=postgres://postgres:$POSTGRES_PASSWORD@localhost:5432
 GOVUK_NOTIFY_API_KEY=
 SIGN_IN_METHOD=persona
+GIAS_DOWNLOAD_LINK=https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public
```
ollietreend, line 5: Is this URL correct? It returns a 404.

```
$ curl -I https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public
HTTP/2 404
```

CatalinVoineag: Well it works only if you put the correct file in there,  Something like `https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public/edubasealldata20231213.csv`

Which is dependent on `Day.today`, so there might not be too much value putting it in here as it doesn't really function on it's own it needs more [ruby logic ](https://github.com/DFE-Digital/itt-mentor-services/blob/cced3e2b6aece6c68fc518900181f9105f83269a/lib/tasks/gias_update.rake#L10)

app/services/csv_imports/gias_import.rb
```diff
@@ -0,0 +1,44 @@
+require "csv"
+
+module CsvImports
+  class GiasImport
```
ollietreend, line 4: I'm pretty confident that GIAS will be the only CSV we need to import. So I expect it'll be the only class that ever exists in the `CsvImports` namespace (I hope!)

Not a strong opinion, but I could imagine a class name of `GiasCsvImporter` would be just as meaningful, and wouldn't need the `CsvImports` namespace.

app/services/service_pattern.rb
```diff
@@ -0,0 +1,16 @@
...
+  def call
+    raise NotImplementedError("#call must be implemented")
+  end
+
+  module ClassMethods
+    def call(...)
+      new(...).call
+    end
+  end
+end
```
ollietreend, line 16: This is nice 😍

lib/tasks/gias_update.rake
```diff
@@ -0,0 +1,17 @@
...
+task gias_update: :environment do
+  logger = Logger.new($stdout)
+  today = Time.zone.today.strftime("%Y%m%d")
+  gias_filename = "edubasealldata#{today}.csv"
+  csv_path = Rails.root.join("tmp/#{gias_filename}")
+
+  logger.info "Downloading the new gias file for #{Time.zone.today}"
+  Down.download(
+    "#{ENV["GIAS_DOWNLOAD_LINK"]}/#{gias_filename}",
+    destination: csv_path
```
ollietreend, line 11: Do we need to specify a destination? It looks like the [Down.download](https://github.com/janko/down#downloading) method will automatically save to a [Tempfile](https://rubyapi.org/o/Tempfile) – i.e. a File in a temporary location.

lib/tasks/gias_update.rake
```diff
@@ -0,0 +1,17 @@
...
+  )
+  logger.info "Done!"
+
+  logger.info "Importing data"
+  CsvImports::GiasImport.call(csv_path)
+end
```
ollietreend, line 17: We should also consider deleting the downloaded file at the end of this task. Otherwise we'll litter the disk with single-use CSV files.

One nice advantage of using the [Tempfile](https://rubyapi.org/o/Tempfile) created by `Down.download` (rather than specifying a location for a File) is that Tempfiles are automatically deleted when the Ruby script finishes. However the [recommended best practice](https://rubyapi.org/o/Tempfile#class-Tempfile-label-Good+practices) is to explicitly close and delete the file anyway.

spec/services/csv_imports/gias_import_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "rails_helper"
+
+RSpec.describe CsvImports::GiasImport do
+  subject { described_class.call("spec/fixtures/test_gias_import.csv") }
+
+  it "upserts the correct schools" do
+    expect { subject }.to change(GiasSchool, :count).from(0).to(1)
+  end
```
ollietreend, line 8: Is it worth also testing that it handles the 'update' path rather than only 'insert'?

This test only proves that we insert new schools.

ollietreend (commented):
This is looking really good ✨

app/models/gias_school.rb
```diff
@@ -0,0 +1,21 @@
...
+#  urn        :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+# Indexes
+#
+#  index_gias_schools_on_urn  (urn) UNIQUE
+#
+class GiasSchool < ApplicationRecord
+  validates :urn, presence: true
```
JamieCleare2525, line 18: Is it worth replicating the validations used in Publish?
https://github.com/DFE-Digital/publish-teacher-training/blob/main/app/models/gias_school.rb

They appear to include `:address1`, `:town`, `:postcode`

CatalinVoineag: Probably, I'll make the change

db/migrate/20231213110952_create_gias_schools.rb
```diff
@@ -0,0 +1,12 @@
+class CreateGiasSchools < ActiveRecord::Migration[7.1]
+  def change
+    create_table :gias_schools, id: :uuid do |t|
```
JamieCleare2525, line 3: Looking at @simonwhatley's prototype is it worth storing more information about the school? 
https://itt-mentoring-beta-94341ca35a2c.herokuapp.com/support/organisations/049fc44e-6876-4884-8a77-d363173ce8d2/details 
Publish store a lot more information. https://github.com/DFE-Digital/publish-teacher-training/tree/main/db

spec/models/user_spec.rb
```diff
@@ -20,7 +20,7 @@
 
   describe "validations" do
     it { is_expected.to validate_presence_of(:email) }
-    it { is_expected.to validate_presence_of(:email) }
+    it { is_expected.to validate_uniqueness_of(:email).case_insensitive }
```
JamieCleare2525, line 23: Cheers for catching this.

db/migrate/20231213110952_create_gias_schools.rb
CatalinVoineag: That's a good shout, I'll include the columns in that prototype

lib/tasks/gias_update.rake
CatalinVoineag: Yeah, that's a good shout

app/services/csv_imports/gias_import.rb
```diff
@@ -0,0 +1,44 @@
...
+      rows = 0
+      errors = []
+
+      CSV
+        .foreach(csv_path, headers: true, encoding: "iso-8859-1:utf-8")
+        .with_index(2) do |school, row_number|
+          next if school["CloseDate"].present?
+
+          begin
+            GiasSchool.find_or_initialize_by(urn: school["URN"]).update!(
```
SocFoot, line 25: I leave it to you but you may want to do a bulk update here and not loop through item.

SocFoot (changes requested):

app/services/csv_imports/gias_import.rb
```diff
@@ -0,0 +1,44 @@
...
+              name: school["EstablishmentName"],
+              town: school["Town"].presence,
```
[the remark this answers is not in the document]
elceebee: This is ok, it returns nil or the value if present. See https://apidock.com/rails/Object/presence

`school["Town"].presence` is the same as `school["Town"].present? ? school["Town"] : nil`

app/services/csv_imports/gias_import.rb
elceebee: Will this take four minutes every time it is run (and we'll be running it daily?) If so, I think there is an argument to be made for running it as a series of smaller jobs, maybe batching 1000 rows at a time. But perhaps that is beyond the scope of this ticket / irrelevant given the relatively small amount of traffic we anticipate on the site.

elceebee (approved):
Locally this works exactly as described. Nice one. 

I think there is a question about performance, particularly if it's going to run every day, but it's not a blocker.

app/models/gias_school.rb
```diff
@@ -0,0 +1,21 @@
+# == Schema Information
```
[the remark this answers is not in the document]
CatalinVoineag: I think this is just the annotate gem https://github.com/ctran/annotate_models, seems to be standard practice of DFE repos

app/models/gias_school.rb
CatalinVoineag: So, I think we don't want to validate too much of the gias data. Not yet at least. When it comes to address/postcode, we might use the `Easting` and `Northing` coordinates, not imported just yet(but in the CSV gias file) to determine the exact location of a school. 

So at the moment we are fine with importing schools that don't have a postcode or town or address.

db/migrate/20231213110952_create_gias_schools.rb
CatalinVoineag: There is no email address in the Gias CSV file, I've added the rest though.

app/services/csv_imports/gias_import.rb
CatalinVoineag: I've gone with a bulk update. On my machine importing almost 24k schools took about 20 seconds, so we should be fine now 🤞🏻

JamieCleare2525 (approved):
Ran locally and everything worked as expected.

SocFoot (approved):

PR #28, opened 2023-12-14
Add update service page for both claims and placements

gms-gs:
## Context

We are making this change to allow us to update users for each service with news

Add MVC for this

## Guidance to review

Run tests and or run app loccally

## Link to Trello card

https://trello.com/c/1qgE9sRc/44-news-and-updates-list-page

## Things to check

## Screenshots
Screenshot 2023-12-14 at 15 02 11
Screenshot 2023-12-14 at 15 18 44

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,6 +46,7 @@ gem "bootsnap", require: false
 # gem 'image_processing', '~> 1.2'
 
 gem "flipflop"
+gem "redcarpet", "~> 3.6"
 
 gem "govuk-components"
 gem "govuk_design_system_formbuilder", "~> 5.0.0"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -79,7 +79,7 @@ GEM
       minitest (>= 5.1)
       mutex_m
       tzinfo (~> 2.0)
-    addressable (2.8.5)
+    addressable (2.8.6)
       public_suffix (>= 2.0.2, < 6.0)
     aes_key_wrap (1.1.0)
     annotate (3.2.0)
@@ -165,16 +165,16 @@ GEM
       terminal-table (>= 1.8)
     globalid (1.2.1)
       activesupport (>= 6.1)
-    govuk-components (4.1.2)
+    govuk-components (5.0.0)
       html-attributes-utils (~> 1.0.0, >= 1.0.0)
       pagy (~> 6.0)
-      view_component (>= 3.3, < 3.7)
+      view_component (>= 3.3, < 3.8)
     govuk_design_system_formbuilder (5.0.0)
       actionview (>= 6.1)
       activemodel (>= 6.1)
       activesupport (>= 6.1)
       html-attributes-utils (~> 1)
-    haml (6.2.3)
+    haml (6.3.0)
       temple (>= 0.8.2)
       thor
       tilt
@@ -183,7 +183,7 @@ GEM
       activesupport (>= 6.1.4.4)
     i18n (1.14.1)
       concurrent-ruby (~> 1.0)
-    io-console (0.7.0)
+    io-console (0.6.0)
     irb (1.10.1)
       rdoc
       reline (>= 0.3.8)
@@ -225,7 +225,7 @@ GEM
     minitest (5.20.0)
     msgpack (1.7.2)
     mutex_m (0.2.0)
-    net-imap (0.4.7)
+    net-imap (0.4.8)
       date
       net-protocol
     net-pop (0.1.2)
@@ -235,7 +235,7 @@ GEM
     net-smtp (0.4.0)
       net-protocol
     nio4r (2.7.0)
-    nokogiri (1.15.5-arm64-darwin)
+    nokogiri (1.15.5-x86_64-darwin)
       racc (~> 1.4)
     nokogiri (1.15.5-x86_64-linux)
       racc (~> 1.4)
@@ -290,7 +290,7 @@ GEM
     puma (6.4.0)
       nio4r (~> 2.0)
     racc (1.7.3)
-    rack (2.2.8)
+    rack (3.0.8)
     rack-oauth2 (2.2.0)
       activesupport
       attr_required
@@ -300,13 +300,13 @@ GEM
       rack (>= 2.1.0)
     rack-protection (3.0.6)
       rack
-    rack-session (1.0.2)
-      rack (< 3)
+    rack-session (2.0.0)
+      rack (>= 3.0.0)
     rack-test (2.1.0)
       rack (>= 1.3)
-    rackup (1.0.0)
-      rack (< 3)
-      webrick
+    rackup (2.1.0)
+      rack (>= 3)
+      webrick (~> 1.8)
     rails (7.1.2)
       actioncable (= 7.1.2)
       actionmailbox (= 7.1.2)
@@ -350,6 +350,7 @@ GEM
     rbs (2.8.4)
     rdoc (6.6.1)
       psych (>= 4.0.0)
+    redcarpet (3.6.0)
     regexp_parser (2.8.3)
     reline (0.4.1)
       io-console (~> 0.5)
@@ -475,7 +476,7 @@ GEM
     validate_url (1.0.15)
       activemodel (>= 3.0.0)
       public_suffix
-    view_component (3.6.0)
+    view_component (3.7.0)
       activesupport (>= 5.2.0, < 8.0)
       concurrent-ruby (~> 1.0)
       method_source (~> 1.0)
@@ -501,6 +502,7 @@ GEM
 PLATFORMS
   arm64-darwin-22
   arm64-darwin-23
+  x86_64-darwin-22
   x86_64-linux
 
 DEPENDENCIES
@@ -533,6 +535,7 @@ DEPENDENCIES
   rails (~> 7.1.2)
   rails-controller-testing
   rails-erd
+  redcarpet (~> 3.6)
   rladr
   rspec
   rspec-rails
--- a/app/assets/components/service_update/view.rb
+++ b/app/assets/components/service_update/view.rb
@@ -1,0 +1,26 @@
+class ServiceUpdate::View < GovukComponent::Base
+  attr_reader :service_update
+
+  delegate :title, :content, :date, to: :service_update
+
+  TITLE_CLASS = "govuk-heading-m govuk-!-margin-bottom-2"
+
+  def initialize(service_update:)
+    @service_update = service_update
+  end
+
+  def title_element
+    tag.h2(title, class: TITLE_CLASS)
+  end
+
+  def date_pretty
+    date.to_date.strftime("%e %B %Y")
+  end
+
+  def content_html
+    custom_render =
+      Redcarpet::Render::HTML.new(link_attributes: { class: "govuk-link" })
+    markdown = Redcarpet::Markdown.new(custom_render)
+    markdown.render(content).html_safe
+  end
+end
--- a/app/models/service_update.rb
+++ b/app/models/service_update.rb
@@ -1,0 +1,34 @@
+class ServiceUpdate
+  CLAIMS_SERVICE_UPDATES_YAML_FILE =
+    Rails.root.join("db/claims_service_updates.yml").freeze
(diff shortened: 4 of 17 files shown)


Gemfile
```diff
@@ -100,3 +100,5 @@ group :test, :development do
   gem "rspec"
   gem "rspec-rails"
 end
+
+gem "redcarpet", "~> 3.6"
```
JamieCleare2525, line 104: Not a blocker, but could you move this higher up the Gemfile?
Just so there is a grouping of "Gems used in all environments", "Gems used in Development", "Gems used in Test", etc.

JamieCleare2525 (approved):
Ran locally and everything working as expected.

db/placements_service_updates.yml
```diff
@@ -0,0 +1,6 @@
+# Example structure for adding service updates
+
+# - date: '2023-11-29'
```
SocFoot, line 3: same

db/claims_service_updates.yml
```diff
@@ -0,0 +1,6 @@
+# Example structure for adding service updates
+
+# - date: '2023-11-29'
```
SocFoot, line 3: Can we have a structure of what the yaml look with type rather than example.

spec/models/service_update_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+        updates = ServiceUpdate.where(service: :placements)
+        expect(updates).to eq({})
+      end
+    end
+  end
+
+  describe ".yaml_file" do
+    it "returns claims YAML file path" do
+      file_path = ServiceUpdate.yaml_file(service: :claims)
+      expect(file_path).to eq(Rails.root.join("db/claims_service_updates.yml"))
```
SocFoot, line 57: You probably want to read from a fixture rather than a file that actually do something

spec/models/service_update_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    it "returns placements YAML file path" do
+      file_path = ServiceUpdate.yaml_file(service: :placements)
+      expect(file_path).to eq(
+        Rails.root.join("db/placements_service_updates.yml")
```
SocFoot, line 63: Same

app/assets/components/service_update/view.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class ServiceUpdate::View < GovukComponent::Base
```
CatalinVoineag, line 3: I'm not that familiar with these Gov components but why do we put them in the assets folder? Why can't we follow the `view component` pattern and have them in `app/components`?

config/routes.rb
```diff
@@ -14,6 +14,11 @@
   get("/auth/developer/sign-out", to: "sessions#signout")
   post("/auth/developer/callback", to: "sessions#callback")
 
+  resources :service_updates,
+            only: %i[index],
+            path: "service-updates",
+            controller: "service_updates"
```
CatalinVoineag, line 20: Not sure if the controller param is necessary if the resource is named the same as the controller 🤷🏻

app/assets/components/service_update/view.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  TITLE_TAGS = %w[h1 h2 h3 h4 h5 h6].freeze
+  TITLE_CLASS = "govuk-heading-m govuk-!-margin-bottom-2"
+
+  def initialize(service_update:, title_tag: "h2")
+    @service_update = service_update
+    @title_tag = title_tag
+  end
+
+  def render?
```
SocFoot, line 16: Seems that you want to use `alias`

app/assets/components/service_update/view.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def date_pretty
+    date.to_date.strftime("%e %B %Y")
+  end
+
+  def content_html
+    custom_render =
+      Redcarpet::Render::HTML.new(link_attributes: { class: "govuk-link" })
+    Redcarpet::Markdown.new(custom_render).render(content).html_safe
```
SocFoot, line 35: you can pass `custom_render` directly

app/models/service_update.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    return {} unless updates
+
+    updates.map { |params| new(params) }
+  end
+
+  def self.yaml_file(service:)
+    case service
+    when :claims
+      Rails.root.join(CLAIMS_SERVICE_UPDATES_YAML_FILE)
```
SocFoot, line 29: Is the `Rails.root.join` useful here? Seems like `CLAIMS_SERVICE_UPDATES_YAML_FILE` already used it

app/models/service_update.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  include ActiveModel::Model
+
+  attr_accessor :date, :title, :content
+
+  def id
+    title.parameterize
+  end
+
+  def self.where(service:)
```
SocFoot, line 17: In rails where is used for return value that would be a list. I think what you want is a get/find

app/models/service_update.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class ServiceUpdate
+  CLAIMS_SERVICE_UPDATES_YAML_FILE =
+    Rails.root.join("db/claims_service_updates.yml").freeze
+  PLACEMENTS_SERVICE_UPDATES_YAML_FILE =
+    Rails.root.join("db/placements_service_updates.yml").freeze
+
+  include ActiveModel::Model
```
SocFoot, line 9: Do you need this? Doesn't seems that you are using active model.

app/models/service_update.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  attr_accessor :date, :title, :content
```
SocFoot, line 11: But if you want to use it you probably want to remove attr_accessor.
```suggestion
class ServiceUpdate
  include ActiveModel::Attributes

  attribute :date, :string
  ...
end
```

app/models/service_update.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class ServiceUpdate
```
SocFoot, line 3: The name update will probably be confusing down the line. Can you use a more specific name like user feed(probably ask product people)

SocFoot (changes requested):
Good work a lot of files a lot of comments

app/assets/components/service_update/view.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
```
elceebee, line 1: I don't think we need this -- it was discussed in an earlier PR.

app/models/service_update.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def id
+    title.parameterize
+  end
+
+  def self.where(service:)
+    path = yaml_file(service:)
+    updates = YAML.load_file(path) if path
+
+    return {} unless updates
```
Nitemaeric, line 21:
```suggestion
    return [] unless updates
```

app/models/service_update.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    updates.map { |params| new(params) }
```
Nitemaeric, line 23: As `updates` should always be an array, mapping over it should be fine. No need to early return.

config/routes/placements.rb
```diff
@@ -1,3 +1,7 @@
-scope module: :placements, as: :placements, constraints: { host: ENV["PLACEMENTS_HOST"] } do
+scope module: :placements,
+      as: :placements,
+      constraints: {
+        host: ENV["PLACEMENTS_HOST"]
+      } do
```
Nitemaeric, line 5: We should avoid adding any changes that aren't relevant to this specific PR objective.

app/models/service_update.rb
elceebee, line 23: You are coding very defensively, which is a very good thing when you are consuming something changeable or depending on asynchronous jobs. But we don't want to code defensively against things we are responsible for (ie, our file structure). Our defence against someone deleting a file should be the tests. In this situation, if someone deletes the file, the tests would pass (because of the early returns), but the feature would be broken. 

And a warning, I'm pedantic about naming -- I don't think these are `params`, they are individual updates. And the `yaml_file` function returns a path, not a `yaml_file`

There are lots of ways to write this, but one alternative would be:

```suggestion
    path = file_path(service:)
    updates = YAML.load_file(path) || []

    updates.map { |service_update| new(service_update) }
```

app/models/service_update.rb
elceebee: Agreed.

spec/models/service_update_spec.rb
elceebee: I think this is fine, it's not actually reading the file, just checking that the function returns the the expected value.

spec/system/home_page_spec.rb
```diff
@@ -1,9 +1,7 @@
 require "rails_helper"
 
 RSpec.feature "Home Page" do
-  after do
-    Capybara.app_host = nil
-  end
+  after { Capybara.app_host = nil }
```
elceebee, line 4: Did the linter do this? If so, fine (odd it wasn't picked up before). 

If it didn't, I don't think we should introduce changes to unrelated files in a PR. Both syntaxes are valid and the change doesn't have anything to do with the task, so there is no need to change it.

config/routes.rb
```diff
@@ -14,6 +14,11 @@
   get("/auth/developer/sign-out", to: "sessions#signout")
   post("/auth/developer/callback", to: "sessions#callback")
 
+  resources :service_updates,
+            only: %i[index],
+            path: "service-updates",
```
elceebee, line 19: Is this the pattern we will want to use for all paths (ie, hyphen instead of snake case?). If so, we should change it at the top of the app in an initialiser (`ActionDispatch`) rather than for each individual route.

elceebee (changes requested):
Nice work, just a few comments. If anything is unclear, let me know, happy to discuss.

Gemfile.lock
```diff
@@ -499,8 +500,7 @@ GEM
     zeitwerk (2.6.12)
 
 PLATFORMS
-  arm64-darwin-22
-  arm64-darwin-23
+  x86_64-darwin-22
```
elceebee, line 503: Please don't delete `arm64-darwin-23` -- that's my M3.

elceebee: How odd! We could get into a weird back and forth where every time I run bundle, it adds it back in, every time you run bundle it removes it :)

app/assets/components/service_update/view.rb
ollietreend, line 35: I think a lot of the heavy lifting can be done using this DfE gem:
https://github.com/DFE-Digital/govuk-markdown

It uses Redcarpet configured to produce HTML with the Design System class names.

For example:

```ruby
GovukMarkdown.render("[A link](/foo)")
```

Will output:

```html
<p class="govuk-body-m">
  <a href="/foo" class="govuk-link">A link</a>
</p>
```

It seems like we could avoid a lot of the manual HTML tag config done within this ViewComponent if we used that instead.

ollietreend (changes requested):

app/helpers/application_helper.rb
```diff
@@ -1,5 +1,5 @@
 module ApplicationHelper
-  def current_service
+  def current_service(request)
```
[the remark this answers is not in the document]
elceebee: The controller has access to the request, so I think you must mean that you need to pass it when calling `current_service` in the view file? There are other ways to access this in views -- You could set a constant like `@service_name ||= current_service.to_s` in the controller and passing that to the view. Or add a view helper that takes the request. 

Making the change like this means we will have to unnecessarily pass the request every time we use `current_service` in a controller -- which will probably be pretty frequently.

ollietreend (approved):

elceebee (approved):

db/claims_service_updates.yml
SocFoot: Is `price:1.5` a float or a decimal? That's why you need the type.

SocFoot (approved):

PR #30, opened 2023-12-14
Update ERD for service split

ollietreend:
## Context

This PR updates the data model to document how we could split Organisations (Schools or Providers) and Users between the different services provided by our app.

## Guidance to review

✨ [View the rendered document](https://github.com/DFE-Digital/itt-mentor-services/blob/update-erd-for-service-split/docs/data-model.md) to see the diagram

## Link to Trello card

https://trello.com/c/qHWnixO0/71-update-the-erd-to-reflect-our-understanding-of-user-and-organisation-onboarding

### DIFF
--- a/docs/data-model.md
+++ b/docs/data-model.md
@@ -28,6 +28,8 @@ erDiagram
   School {
     uuid id PK
     string urn FK "Primary key for schools in GIAS"
+    bool placements_enabled "Indicates if the School has been onboarded into the School Placements service"
+    bool claims_enabled "Indicates if the School has been onboarded into the Track & Pay service"
   }
 
   Mentor {
@@ -60,6 +62,14 @@ erDiagram
     string trn FK "Primary key for people in DQT"
   }
 
+  GiasSchool {
+    uuid id PK
+    string urn UK "Unique identifier for Schools"
+    string name
+  }
+
+  School ||--|| GiasSchool : "belongs to"
+
   School ||--|{ Mentor : "has many"
   Provider ||--|{ School : "has many"
   Trainee }|--|| Provider : "belongs to"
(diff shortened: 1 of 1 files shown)


CatalinVoineag (commented):
This looks sensible to me. I guess we'll have to also store some app id or something from the DFE sign in, I can see this being in the `memebership` table and a user having multiple memberships. 👍🏻

CatalinVoineag (approved):

docs/data-model.md
```diff
@@ -28,6 +28,8 @@ erDiagram
   School {
     uuid id PK
     string urn FK "Primary key for schools in GIAS"
+    bool placements_enabled "Indicates if the School has been onboarded into the School Placements service"
```
JamieCleare2525, line 31: Rather than a `boolean` should these be `datetime` then we can also track when the onboarding happened.

JamieCleare2525 (changes requested):

docs/data-model.md
ollietreend: This could be useful, and it's a convenient way to record when a school was onboarded. But I don't necessarily want to make any assumptions about what we should track yet, nor how exactly to track it.

For example, there may be a need for an audit trail which shows the date/time a school was onboarded but also who did it. Or perhaps it's possible to 'offboard' a school from the service temporarily, and we'd need the audit trail to show a complete onboard/offboard history for schools. In that case, we would probably model the data slightly differently.

I'm also conscious that some of these needs may also potentially be met by the [DfE Analytics gem](https://github.com/DfE-Digital/dfe-analytics) and platform. The approach is to send granular event data to BigQuery for later filtering and analysis – so I could imagine us being able to source the date/time a school was enabled via that too.

So for that reason, I'd probably be keen to stick with a `boolean` field. That seems to be the simplest thing we can do to meet the immediate need of our service.

JamieCleare2525: Na. I'm happy to concede!

JamieCleare2525 (approved):

docs/data-model.md
```diff
@@ -60,6 +62,27 @@ erDiagram
...
+    uuid id PK
+    string email UK
+  }
+
+  Membership {
+    uuid id PK
+    uuid user_id FK
+    string organisation_type FK "Polymorphic association with School or Provider"
+    string organisation_id FK "Polymorphic association with School or Provider"
+    enum service "Placements or Claims"
```
Nitemaeric, line 75: After our discussion, I think it's best if we move this enum into the `User` table.

Once a user signs in, we need to know if the user is logged in from the Claims DSI or Placements DSI.

With a split User on both sides, we can easily find our user count.

Nitemaeric (approved):
1 comment. Overall looks great! Thanks!

docs/data-model.md
CatalinVoineag: I might not have all the context here but having the 'service enum' on the membership will allow 1 user access to both services through multiple memberships. If we need to know from which service the user logs in we can use some other column? Just a thought

Nitemaeric: We've considered the likelihood that a user may have no memberships when they first log in.

ollietreend: I think either approach could be made to work. I've been on the fence about whether 'service' should live on the User or the Membership. But having given it some discussion and thought, I think User is probably the right place for it.

This would support User models in each service which look something like this (pseudo code):

```ruby
class User {
  validates :email, uniqueness: { scope: :service }
}

class Claims::User < User {
  default_scope { where service: "claims" }
}

class Placements::User < User {
  default_scope { where service: "placements" }
}
```

Then we'd easily be able to do things like this, without having to think about the service:

```ruby
Claims::User.where(email: "john.smith@example.com").exists?
Claims::User.count
# etc.
```

Since we're also going to need different DfE Sign-in client IDs, it seems reasonable to carry this separation through to the User model.

ollietreend: For what it's worth, I think the same approach could be carried forward for Schools too.

```ruby
class School {
  validates :urn, presence: true
}

class Claims::School {
  default_scope { where claims_enabled: true }
}

class Placement::School {
  default_scope { where placements_enabled: true }
}
```

And then when we need to list all schools within our services, we can simply use:

```ruby
Claims::School.all
```

ollietreend: @Nitemaeric are you happy with this? (Will squash before merge)

https://github.com/DFE-Digital/itt-mentor-services/pull/30/commits/21b2fbf25e89eb1cc6a4646d90c28411e49ca37a

docs/data-model.md
```diff
@@ -65,14 +65,14 @@ erDiagram
   User {
     uuid id PK
     string email UK
+    enum service "Placements or Claims"
```
elceebee, line 68: I'm happy to go this way -- but I wonder if it's useful to use the same pattern we're using above for schools (placements_enabled, claims_enabled). It would be consistent, but also, the unique constraint would then just be on the `email`, rather than `email`, `service` combination. 

Perfectly happy to go with the enum approach. Just putting this out there for a point of discussion.

docs/data-model.md
Nitemaeric: Yeah, I like the idea of domain scoped models either pointing to the shared model or applying a scope on the shared model.

PR #33, opened 2023-12-18
Add healthcheck endpoint

CatalinVoineag:
## Context

Add a simple healthcheck page so kubernetes can monitor the container.

New healthcheck route

## Guidance to review

Just visit `/healthcheck`

- 2 file changes are because of the linter.

## Link to Trello card

https://trello.com/c/z21wL0OH/77-add-healthcheck-endpoint

## Things to check

## Screenshots

Screenshot from 2023-12-18 13-58-41

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,4 +16,6 @@ Rails.application.routes.draw do
 
   draw :placements
   draw :claims
+
+  get :healthcheck, controller: :heartbeat
 end
--- a/app/controllers/heartbeat_controller.rb
+++ b/app/controllers/heartbeat_controller.rb
@@ -1,0 +1,20 @@
+class HeartbeatController < ApplicationController
+  def healthcheck
+    checks = { database: database_alive? }
+    status = checks.values.all? ? :ok : :service_unavailable
+
+    render(status:, json: { checks: })
+  end
+
+  private
+
+  def database_alive?
+    ActiveRecord::Base.connection_pool.with_connection do |conn|
+      conn.execute("SELECT 1")
+    end
+
+    ActiveRecord::Base.connected?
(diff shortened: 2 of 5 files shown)


elceebee (approved):

config/routes.rb
```diff
@@ -16,4 +16,6 @@
 
   draw :placements
   draw :claims
+
+  get "/healthcheck", to: proc { [200, {}, ["OK"]] }
```
Nitemaeric, line 20: Maybe https://edgeapi.rubyonrails.org/classes/Rails/HealthController.html could be used?

```suggestion
  get "/healthcheck", to: "health#show"
```

Rails supposedly defaults this to `/up`, but locally it isn't in there for me.

Nitemaeric (commented):
Nitpick, but could you split out the lint changes into a separate commit please?

config/routes.rb
CatalinVoineag: I've started reading that and figure out that we don't check if the DB is up or if we add sideqkiq and redis later on, we don't check the health of those. Refactored this into a `Heartbeat` controller, similar to the one in register. To check for these things

JamieCleare2525 (approved):

config/routes.rb
```diff
@@ -16,4 +16,6 @@
 
   draw :placements
   draw :claims
+
+  get :healthcheck, controller: :heartbeat
```
Nitemaeric, line 20: Logically, this is all good 👍 Out of curiousity, why is the controller "Heartbeat" and the action "healthcheck"? Are you planning to add additional heartbeats?

CatalinVoineag: I don't like having methods that are named the same as the class they are in, and this way we can add additional heartbeats if we need to

PR #39, opened 2023-12-19
Claims / Sign in page should redirect to persona select

Nitemaeric:
## Context

We want to lay down the foundations of user pages and flows. There is no data added, just _blank_ pages to highlight routing structures.

- Remove "University" persona from Claims personas
- Redirect Claims users to sign-in page when landing on service root url
- Add "account" and "sign out" navigation items to navbar when signed in

## Guidance to review

- Head to the Claims service
- Be redirected to the personas page
- Login as any of the personas
- Be redirected to their internal pages (General users pending)

## Link to Trello card

https://trello.com/c/ow40pfAm/75-update-landing-page-of-app-to-sign-in-or-persona-select

## Things to check

## Screenshots / Videos

https://github.com/DFE-Digital/itt-mentor-services/assets/4231530/eb2fd6d8-ea62-4f6b-b7c7-90aef4f3016e

### DIFF
--- a/app/controllers/claims/pages_controller.rb
+++ b/app/controllers/claims/pages_controller.rb
@@ -1,2 +1,3 @@
 class Claims::PagesController < ApplicationController
+  before_action :authenticate_user!
 end
--- a/app/controllers/claims/pages_controller.rb
+++ b/app/controllers/claims/pages_controller.rb
@@ -1,3 +1,2 @@
 class Claims::PagesController < ApplicationController
-  before_action :authenticate_user!
 end
--- a/app/controllers/claims/support/application_controller.rb
+++ b/app/controllers/claims/support/application_controller.rb
@@ -1,13 +1,11 @@
 class Claims::Support::ApplicationController < ApplicationController
-  before_action :authenticate_support_user!
+  before_action :authenticate_user!, :authorize_user!
 
   private
 
-  def authenticate_support_user!
-    authenticate_user!
+  def authorize_user!
+    return if current_user.support_user?
 
-    unless current_user.support_user?
-      redirect_to claims_root_path, alert: "You cannot perform this action"
-    end
+    redirect_to claims_root_path, alert: "You cannot perform this action"
   end
 end
--- a/app/controllers/claims/support/application_controller.rb
+++ b/app/controllers/claims/support/application_controller.rb
@@ -1,0 +1,13 @@
+class Claims::Support::ApplicationController < ApplicationController
+  before_action :authenticate_support_user!
+
+  private
+
+  def authenticate_support_user!
+    authenticate_user!
+
+    unless current_user.support_user?
+      redirect_to claims_root_path, alert: "You cannot perform this action"
+    end
+  end
+end
--- a/app/controllers/personas_controller.rb
+++ b/app/controllers/personas_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class PersonasController < ApplicationController
+  skip_before_action :authenticate_user!, only: %i[index]
+
   def index
     @personas = Persona.public_send(current_service).decorate
   end
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -66,6 +66,18 @@ RSpec.configure do |config|
   config.filter_rails_from_backtrace!
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
+
+  config.around(:each, type: :request) do |example|
+    host! ENV["CLAIMS_HOST"]
+    example.run
+    host! nil
+  end
+
+  config.around(:each, type: :system) do |example|
+    Capybara.app_host = "http://#{ENV["CLAIMS_HOST"]}"
+    example.run
+    Capybara.app_host = nil
+  end
 end
 
 Shoulda::Matchers.configure do |config|
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -35,7 +35,10 @@ begin
 rescue ActiveRecord::PendingMigrationError => e
   abort e.to_s.strip
 end
+
 RSpec.configure do |config|
+  config.include ServicesTestHelper, type: :system
+
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
   # config.fixture_paths = [Rails.root.join("spec/fixtures")]
 
--- a/spec/requests/sessions_spec.rb
+++ b/spec/requests/sessions_spec.rb
@@ -1,27 +1,23 @@
 require "rails_helper"
 
 RSpec.describe "Sessions", type: :request do
-  around do |example|
-    host! ENV["PLACEMENTS_HOST"]
-    example.run
-    host! nil
-  end
-
   describe "POST /auth/developer/callback" do
     it "returns http success" do
-      placements_user = create(:placements_user)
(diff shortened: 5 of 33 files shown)


app/controllers/claims/pages_controller.rb
```diff
@@ -1,2 +1,3 @@
 class Claims::PagesController < ApplicationController
+  before_action :authenticate_user!
```
CatalinVoineag, line 2: Should we just put this in the ApplicationController? And if any controllers don't need it we can skip it in there.

app/controllers/claims/schools_controller.rb
```diff
@@ -0,0 +1,2 @@
+class Claims::SchoolsController < ApplicationController
```
CatalinVoineag, line 1: Related to the comment above, don't we need to auth the user here as well?

app/controllers/personas_controller.rb
```diff
@@ -2,6 +2,6 @@
 
 class PersonasController < ApplicationController
   def index
-    @personas = Persona.all.decorate
+    @personas = Persona.where(service: current_service).decorate
```
elceebee, line 5: You can use the scope on the model for each service.

```suggestion
    @personas = Persona.public_send(current_service).decorate
```

app/controllers/claims/pages_controller.rb
elceebee: I agree with Catalin that this would be better on the ApplicationController and we skip it when not needed. It is easier to require auth by default since that is what we will need most often.

elceebee (approved):
Would be good to merge [this PR](https://github.com/DFE-Digital/itt-mentor-services/pull/40). it will fix the broken tests and adds a few more tests for returning the correct user.

I agree with Catalin about requiring authentication by default and skipping as an exception rather than the other way around.

app/controllers/claims/pages_controller.rb
Nitemaeric: I think I'd propose something slightly different.

```
class Claims::PagesController < Claims::Base::ApplicationController # Base / Users / Main any of these namespaces work for me
end

class Claims::Base::ApplicationController < ApplicationController
  before_action :authenticate_user!
end
```

This way, we have a base controller for general user routes and one for support. We also have the top-level `ApplicationController`'s action hooks unmodified and likely the path of least surprise. Anything inheriting from `ApplicationController` should have access to all helpers and custom controller methods, but without any additional behaviour added.

What do you both think?

Nitemaeric: I think auth by default is also viable now that I think about it though 🤔 .

Nitemaeric: Yeah, scratch the above suggestion. Keep it simple.

CatalinVoineag: I guess I don't really see the value of having ` Claims::Base::ApplicationController` at the moment, feels like we are abstracting too soon.

I like the idea of auth by default and if you need to skip it for the support controller you can skip it or override the method from the `ApplicationController`.

In the end, I don't really have a strong opinion on this. Happy with both approaches

app/controllers/claims/support/application_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Claims::Support::ApplicationController < ApplicationController
+  before_action :authenticate_support_user!
```
CatalinVoineag, line 2: Don't we need `skip_before_action :authenticate_user!` here? I think it will go through the application controller twice now.

We could also just override the `authenticate_user` method in here if you want.

CatalinVoineag (commented):
Pulled this on local, works as expected 👍🏻 Just one comment that I think we should address.

app/controllers/claims/support/application_controller.rb
ollietreend: The structure of the equivalent `Placements::Support::ApplicationController` class changed before PR #42 was merged, so you'll probably want to update this file to match.

https://github.com/DFE-Digital/itt-mentor-services/blob/a485beef8f5ddedbf32f3484d749162214958fed/app/controllers/placements/support/application_controller.rb#L1-L11

In addition to that, both `[SERVICE]::Support::ApplicationController` classes will now need to update the `before_action` line so it doesn't re-declare `:authenticate_user!` since that's now in the base `ApplicationController` class.

In other words, both classes should now look like this:

```diff
class [SERVICE]::Support::ApplicationController < ApplicationController
-  before_action :authenticate_user!, :authorize_user!
+  before_action :authorize_user!

  private

  def authorize_user!
    return if current_user.support_user?

    redirect_to placements_root_path, alert: "You cannot perform this action"
  end
end
```

spec/rails_helper.rb
```diff
@@ -63,6 +66,18 @@
...
+    host! ENV['CLAIMS_HOST']
+    example.run
+    host! nil
+  end
+
+  config.around(:each, type: :system) do |example|
+    Capybara.app_host = "http://#{ENV['CLAIMS_HOST']}"
+    example.run
+    Capybara.app_host = nil
+  end
```
ollietreend, line 80: It looks like there's duplication here between this config and the config manually applied on system specs in @CatalinVoineag's PR.

https://github.com/DFE-Digital/itt-mentor-services/blob/1f7ff126fafb8160301460e2fba88a2d49720da6/spec/system/claims/view_organisations_spec.rb#L4-L8

Also: how will this config know which hostname to use – `CLAIMS_HOST` or `PLACEMENTS_HOST`?

spec/requests/sessions_spec.rb
```diff
@@ -1,27 +1,23 @@
...
-    host! ENV["PLACEMENTS_HOST"]
-    example.run
-    host! nil
-  end
-
   describe "POST /auth/developer/callback" do
     it "returns http success" do
-      placements_user = create(:placements_user)
-      post auth_developer_callback_path,
+      claims_user = create(:claims_user)
```
ollietreend, line 6: Any reason why this was changed from `placements_user` to `claims_user`? Was this a meaningful change?

ollietreend (dismissed):
This is looking good ✨

I've added one change request, a comment, and a question.

spec/support/capybara.rb
ollietreend, line 1: Does this branch need rebasing on `main` now that #51 has been merged in, or is this introducing duplicate config?

spec/rails_helper.rb
Nitemaeric: This is the default, it sets the host to `CLAIMS` by default. I've made the change as I saw a previous PR sets `CLAIMS` to be the default host.

For system tests, I've added the `given_i_am_on_the_xxxxx_site` helper method, which is the only time we should really need to think about the site we're on. In unit tests, we shouldn't need to think about it.

Alternatively, we _could_ pass in a `service: :claims` or `service: :placements` in the spec describe options,

```ruby
RSpec.describe "Home Page", type: :system, service: :claims do
```

spec/requests/sessions_spec.rb
Nitemaeric: Changing the default service to `:claims` means the service will look up a `Claims::User` by default.

Nitemaeric:
[Update landing page of app to Sign In or Persona select](https://trello.com/c/ow40pfAm/75-update-landing-page-of-app-to-sign-in-or-persona-select)

CatalinVoineag (approved):
Tested this on local, works as expected. Good work

PR #40, opened 2023-12-19
Lb/current user method

elceebee:
## Context

See slack thread: https://ukgovernmentdfe.slack.com/archives/C06962474AC/p1702994098513929

We need the `current_user` method to return a Service::User object, not a `DfeSignInUser` instance. 

I did this as part of the support_user journey work, it makes sense to implement it earlier as needed in other places. 

When a user is logged in a service, the `current_user` method returns the appropriate user object - `Placements::User` for the placements service, `Claims::User` for the claims service

No UI changes. 

## Things to check

- [NA] This code does not rely on migrations in the same Pull Request

### DIFF
--- a/spec/features/personas/sign_in_as_persona_spec.rb
+++ b/spec/features/personas/sign_in_as_persona_spec.rb
@@ -4,7 +4,7 @@ require "rails_helper"
 
 feature "Sign In as Persona" do
   around do |example|
-    Capybara.app_host = "https://#{ENV["PLACEMENTS_HOST"]}"
+    Capybara.app_host = "https://#{ENV["CLAIMS_HOST"]}"
     example.run
     Capybara.app_host = nil
   end
--- a/spec/models/provider_spec.rb
+++ b/spec/models/provider_spec.rb
@@ -18,6 +18,8 @@ RSpec.describe Provider, type: :model do
     subject { create(:provider) }
 
     it { is_expected.to validate_presence_of(:provider_code) }
-    it { is_expected.to validate_uniqueness_of(:provider_code).case_insensitive }
+    it do
+      is_expected.to validate_uniqueness_of(:provider_code).case_insensitive
+    end
   end
 end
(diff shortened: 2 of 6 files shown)


spec/features/personas/sign_in_as_persona_spec.rb
```diff
@@ -4,7 +4,7 @@
 
 feature "Sign In as Persona" do
   around do |example|
-    Capybara.app_host = "https://#{ENV["PLACEMENTS_HOST"]}"
+    Capybara.app_host = "https://#{ENV["CLAIMS_HOST"]}"
```
JamieCleare2525, line 7: Why is this being changed?

elceebee, line 7: this is a quick fix for the persona test -- I'm actually splitting it into two different tests (one for claims, one for placements in the next PR, but I just want to get the current_user stuff in so we can all use it for now.

spec/models/provider_spec.rb
```diff
@@ -18,6 +18,8 @@
     subject { create(:provider) }
 
     it { is_expected.to validate_presence_of(:provider_code) }
-    it { is_expected.to validate_uniqueness_of(:provider_code).case_insensitive }
+    it do
+      is_expected.to validate_uniqueness_of(:provider_code).case_insensitive
+    end
```
elceebee, line 23: Linting, not sure how this snuck in...

spec/features/personas/sign_in_as_persona_spec.rb
elceebee: This is a good question! The default user is a claims user, so all the tests fail unless we are in the claims context (or change all the users to be placement users.

In the fuller PR (for the support_user journey), I've split this into two different tests. Personal logins for each service, as they will diverge as the user journeys become more specific to the service context. Happy to walk through it if that doesn't make sense.

JamieCleare2525 (approved):
Resolves the issue of `User` being assigned to `:current_user`

Nitemaeric (approved):

PR #41, opened 2023-12-19
Add phase banner to the app

CatalinVoineag:
## Context

We want to add a phase banner that will indicate to our users that this app is in Beta. The banner title is dynamic based on the environment we are in and both claims and placements can control the content of it.

- Phase banner Styles
- Translations
- HostingEnvironment module

## Guidance to review

- The link in the banner only redirects to the current page, we need to talk to product about where we want that link to go. It was not in the scope of this story

Run the app on local and you should see the `Development` phase banner
Screenshot from 2023-12-20 09-55-24

## Link to Trello card

https://trello.com/c/ivbin1aW/73-add-beta-banner-to-app

## Things to check

## Screenshots

Development:
Screenshot from 2023-12-20 09-55-24

Production:
Screenshot from 2023-12-20 10-01-23

### DIFF
--- a/app/assets/stylesheets/components/_page_banner.scss
+++ b/app/assets/stylesheets/components/_page_banner.scss
@@ -1,0 +1,22 @@
+.app-phase-banner .govuk-tag {
+  color: govuk-colour("white");
+  text-transform: uppercase;
+  letter-spacing: 1px;
+  font-weight: $govuk-font-weight-bold;
+}
+
+.app-phase-banner__env--development .govuk-tag {
+  background: govuk-colour("dark-grey");
+}
+
+.app-phase-banner__env--qa .govuk-tag {
+  background-color: govuk-colour("orange");
+}
+
+.app-phase-banner__env--staging .govuk-tag {
+  background-color: govuk-colour("red");
+}
+
+.app-phase-banner__env--beta .govuk-tag {
+  background: govuk-colour("blue");
+}
--- a/config/locales/placements/en/models/hosting_environment.yml
+++ b/config/locales/placements/en/models/hosting_environment.yml
@@ -1,0 +1,6 @@
+en:
+  models:
+    hosting_environment:
+      placements:
+        banner:
+          description: Make a complaint or give feedback
--- a/app/assets/stylesheets/application.sass.scss
+++ b/app/assets/stylesheets/application.sass.scss
@@ -3,3 +3,5 @@ $govuk-new-link-styles: true;
 $govuk-assets-path: "";
 
 @import "govuk-frontend/dist/govuk/all";
+
+@import "components/all";
--- a/app/assets/stylesheets/components/_all.scss
+++ b/app/assets/stylesheets/components/_all.scss
@@ -1,0 +1,1 @@
+@import "page_banner";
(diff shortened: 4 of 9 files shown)


CatalinVoineag:
> @CatalinVoineag Could you add a screenshot please. Looks good other than that! 🥳 (edit: PR is in draft when I commented so it may not actually be ready yet)

Yeah, it's still in draft. I'll add everything in a sec

config/locales/placements/en/models/hosting_environment.yml
```diff
@@ -0,0 +1,7 @@
+en:
```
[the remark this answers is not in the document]
CatalinVoineag: I think they will look the same for both in the beginning but according to the assumption on the ticket they might look different later

`Track & Pay and School Placements may move out of BETA at different times. We’ll likely need a way of configuring these independently.`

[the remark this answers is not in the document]
elceebee: I think it makes sense for them to be separate because they aren't necessarily on the same time line -- one might stop being a 'beta' service before the other.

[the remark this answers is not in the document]
CatalinVoineag: In that article, it also suggests to move to rack for running the system specs when you don't have javascript and use selenium when you have.

This might speed up our specs a bit considering we are quite light on javascript

app/assets/stylesheets/components/_page_banner.scss
```diff
@@ -0,0 +1,22 @@
+.app-phase-banner .govuk-tag {
+  color: govuk-colour("white");
+  text-transform: uppercase;
+  letter-spacing: 1px;
+  font-weight: 700;
```
elceebee, line 5: I always get a bit anxious when I see specific values in css. Generally difficult to maintain.

I think for the font-weight, you can use:
`$govuk-font-weight-bold`, but I don't see a variable for letter spacing.

elceebee (approved):
Testing locally and it works as expected. Would be good to use defined styles where possible - so changing the font-weight from `700` to `$govuk-font-weight-bold`, but it's not a blocker.

app/assets/stylesheets/components/_page_banner.scss
CatalinVoineag: Good point, I'll change it

app/models/hosting_environment.rb
```diff
@@ -0,0 +1,13 @@
+module HostingEnvironment
```
Nitemaeric, line 1: Hmm, would this be better suited to go in `lib` or `services`? As is, it doesn't seem to model any data, but functions as a service.

config/locales/placements/en/models/hosting_environment.yml
```diff
@@ -0,0 +1,7 @@
...
+  models:
+    hosting_environment:
+      placements:
+        name: beta
```
Nitemaeric, line 5: Hmm, this feels like we're using this dictionary for config. Can I suggest an env var as we don't use `Settings`?

app/models/hosting_environment.rb
CatalinVoineag: I'll put it in `lib`

config/locales/placements/en/models/hosting_environment.yml
CatalinVoineag: The think with using env vars though is that you don't allow each service do define their hosting envname

Nitemaeric: Sure. In that case, should this be a constant in the `HostingEnvironment` module instead?

Nitemaeric (approved):

PR #34, opened 2023-12-18
Initial draft for testing practices ADR

Nitemaeric:
## Context

We want to align on our testing practices to make testing both easier to write and to review.

## Changes proposed in this pull request

- Add ADR related to Testing Practices

## Guidance to review

- There's a lot of other testing practices we may want to agree on. Feel free to suggest additions.
- Click [here](https://github.com/DFE-Digital/itt-mentor-services/blob/adr/testing-practices/adr/00003-testing-practices.md) to preview the ADR.

## Link to Trello card

https://trello.com/c/jzu6Vgel/69-write-up-adr-for-testing-practices

## Things to check

- [ ] If the code removes any existing feature flags, a data migration has also been added to delete the entry from the database
- [ ] This code does not rely on migrations in the same Pull Request
- [ ] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [ ] If this code adds a column to the DB, decide whether it needs to be in analytics yml file or analytics blocklist
- [ ] API release notes have been updated if necessary
- [ ] If it adds a significant user-facing change, is it documented in the [CHANGELOG](CHANGELOG.md)?
- [ ] Required environment variables have been updated or [added to the Azure KeyVault](/docs/environment-variables.md#deploy-pipeline)

PR #42, opened 2023-12-20
Lb/feature/login as support user

elceebee:
## Context

Support users for the placements service should see a list of Organisations after they login

- Placements support user model
- Logic for determining if a user is a support user
- views and redirects for placement support users
- Changes to personas controller to only show the personas for a particular service
- Takes some o the authentication logic from a [yet-to-be-merged in PR](https://github.com/DFE-Digital/itt-mentor-services/pull/39) 

## Guidance to review

Login as a support user in the placements service. 

## Link to Trello card

https://trello.com/c/9QbPxRYT

## Things to check

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/44073106/b7a1ff2e-4271-4ad6-aba6-9c7b8a2a810b


app/controllers/sessions_controller.rb
```diff
@@ -7,7 +7,7 @@ def callback
     if current_user
       # DfESignInUsers::Update.call(user: current_user, sign_in_user: sign_in_user)
 
-      redirect_to(root_path)
+      redirect_to after_sign_in_path
```
elceebee, line 10: This is from Dan's PR

app/helpers/routes_helper.rb
```diff
@@ -0,0 +1,19 @@
+module RoutesHelper
```
elceebee, line 1: This file is from Dan's PR

app/models/school.rb
```diff
@@ -16,10 +16,12 @@
 #  index_schools_on_urn         (urn) UNIQUE
 #
 class School < ApplicationRecord
-  has_one :gias_school, foreign_key: :urn, primary_key: :urn
+  belongs_to :gias_school, foreign_key: :urn, primary_key: :urn
```
elceebee, line 19: This is how it should have been implemented when I merged in this code earlier in the week.

JamieCleare2525 (approved):
Tested locally. All looks good

spec/requests/personas_spec.rb
```diff
@@ -1,10 +1,33 @@
 require "rails_helper"
 
 RSpec.describe "Personas", type: :request do
-  describe "GET /personas" do
-    it "returns http success" do
-      get personas_path
-      expect(response).to have_http_status(:success)
+  context "placements" do
```
CatalinVoineag, line 4: I thought we are not doing these request specs for controller that are not API as per https://github.com/DFE-Digital/itt-mentor-services/pull/34

CatalinVoineag (approved):
Looks good, just one comment from me but it's not a blocker. Well done

spec/requests/personas_spec.rb
elceebee: You are right of course! I will delete it.

elceebee:
[Login as a support user and see organisations](https://trello.com/c/9QbPxRYT/65-login-as-a-support-user-and-see-organisations)

app/controllers/application_controller.rb
```diff
@@ -15,4 +16,26 @@ def sign_in_user
   def current_user
     @current_user ||= sign_in_user&.user
   end
+
+  def support_user
+    @support_user ||= sign_in_user&.support_user
+  end
```
ollietreend, line 22: Do we need a separate `support_user` method here? It feels a bit weird that the controller will have both a `current_user` and a `support_user` which will represent the same thing in the case of a support user being logged in.

I feel like it'd probably be sufficient to use the existing `current_user` object and authorise access with:

```ruby
if current_user.support_user?
  # user is allowed access to support things
else
  # 🙅🏻‍♂️
end
```

And, as a bit of an aside – not necessarily something to change in this PR:

I'm now wondering if we really need the `class Placements::SupportUser < Placements::User` at all. 🤔 It's effectively just a scope. A `Placements::SupportUser` is functionally the same as a `Placements::User` at this point.

This may change in a future refactor if we decide that the organisations list should be generated by the model as `user.organisations` or `support_user.organisations` because that behaviour would be different.

app/views/placements/support/organisations/index.html.erb
```diff
@@ -0,0 +1,23 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t("organisations").capitalize %></h1>
```
ollietreend, line 3: Not a strong opinion, but would it make more sense for the locale file to define the capital "O" on this, rather than us post-processing the translation?

```suggestion
    <h1 class="govuk-heading-xl"><%= t("organisations") %></h1>
```

```yaml
en:
  organisations: Organisations
```

Or if the key seems too clashy, I guess it could be `organisations_heading`?

spec/factories/users.rb
```diff
@@ -34,4 +34,10 @@
   factory :placements_user, class: "Placements::User", parent: :user do
     service { "placements" }
   end
+
+  factory :placements_support_user,
+          class: "Placements::SupportUser",
+          parent: :user do
+    service { "placements" }
+  end
```
ollietreend, line 42: If we do stick with `Placements::SupportUser` being its own entity, then we probably want to add:

```ruby
support_user { true }
```

to make it more true to reality.

ollietreend (changes requested):
This is looking good @elceebee ✨

I'd be interested to talk through the choice of `current_user` vs `support_user` with you.

app/views/placements/support/organisations/index.html.erb
ollietreend: (Feel free to disagree and mark this as 'resolved')

app/controllers/application_controller.rb
elceebee: I suppose I was assuming there would be methods on the `support_user` model that would be distinct from the `current_user` (eg, like the organisations you mention). I'm happy to go either way.

app/views/placements/support/organisations/index.html.erb
elceebee: Sure!

spec/factories/users.rb
elceebee: Because `Placements::SupportUser` is scoped to `support_user == true`, this isn't necessary. I might be misunderstanding...

CatalinVoineag:
Sorry to add a comment after my review. Do you think it's worth updating the seeds to always have organisations for a support user?

app/controllers/application_controller.rb
ollietreend: I think this might give us a more elegant way forward, where we don't need to define a secondary `support_user` method.

PR: https://github.com/DFE-Digital/itt-mentor-services/pull/45

spec/factories/users.rb
ollietreend: I'd assumed `default_scope` only filtered 'find' queries, so it wasn't possible to find a non-support user from the database. I didn't realise it also automatically sets `support_user = true` in the model.

```ruby
Placements::User.new.attributes["support_user"] # => false
Placements::SupportUser.new.attributes["support_user"] # => true
```

Very cool! 🤓 Every day's a school day!

JamieCleare2525 (approved):
LGTM! Tested since @ollietreend changes!

CatalinVoineag (approved):
Shipit 😄

ollietreend (approved):

PR #45, opened 2023-12-20
Make `current_user` return a `Placements::SupportUser` for support users

ollietreend:
## Context

The School Placements service has two user models: `Placements::User` and `Placements::SupportUser < Placements::User`. We decided that support users will need slightly different behaviour from 'regular' users – for example, they'll need access to _all_ Organisations rather than only those they have Memberships to.

I've updated the `current_user` method so that it returns the most appropriate User type for the circumstances. If the user is in the Placements service and is a Support User, then we'll return a `Placements::SupportUser` object.

I'm implementing in the hope that it'll simplify the implementation of PR #42 which needs to distinguish between regular users and support users. For more context, see this comment thread: https://github.com/DFE-Digital/itt-mentor-services/pull/42#discussion_r1432836796

## Guidance to review

I'd recommend reviewing commit-by-commit (and focussing primarily on the final commit).

1. The first commit was an unrelated change that `bundle` decided to make to `Gemfile.lock`. It seemed insistent on making this change, so I've just rolled with it.
2. The second commit is a micro-optimisation, but one which I thought I'd make while I was in the area. It should avoid unnecessary updates to the session object when calling `sign_in_user`.
3. This is the meat of the PR. Hopefully the tests explain what's changed.

## Link to Trello card

https://trello.com/c/9QbPxRYT/65-login-as-a-support-user-and-see-organisations

### DIFF
--- a/app/models/dfe_sign_in_user.rb
+++ b/app/models/dfe_sign_in_user.rb
@@ -47,7 +47,16 @@ class DfESignInUser
 
   def user
     # TODO: When dfe sign-in is fully implemented, we will be able to find the user by the id == id_token.
-    @user ||= user_klass.find_by(email:)
+    @user ||=
+      begin
+        user = user_klass.find_by(email:)
+
+        if service == :placements && user&.support_user?
+          user = user.becomes(Placements::SupportUser)
+        end
+
+        user
+      end
   end
 
   def self.end_session!(session)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -237,6 +237,8 @@ GEM
     net-smtp (0.4.0)
       net-protocol
     nio4r (2.7.0)
+    nokogiri (1.15.5-arm64-darwin)
+      racc (~> 1.4)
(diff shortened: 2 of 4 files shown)


app/models/dfe_sign_in_user.rb
```diff
@@ -47,7 +47,15 @@ def self.load_from_session(session)
 
   def user
     # TODO: When dfe sign-in is fully implemented, we will be able to find the user by the id == id_token.
-    @user ||= user_klass.find_by(email:)
+    @user ||= begin
+      user = user_klass.find_by(email:)
+
+      if service == :placements && user&.support_user?
+        user = user.becomes(Placements::SupportUser)
+      end
```
ollietreend, line 55: It feels _slightly weird_ to be doing this here. But I figured the `DfeSignInUser` model already has an awareness of the split between services and how that affects the type of User class returned. So this seems like the most obvious place to put it.

JamieCleare2525 (approved):

PR #47, opened 2023-12-21
Organisation index per service

CatalinVoineag:
## Context

We want an organisation index for each service.
This index page will be shown to the user after he sigin in as a
persona only if the user has multiple organisations.

An organisation can be a school or a provider, the Membership
polymorphic table handles this.

The Placements service will show both schools and providers, the claims
only schools.

## Changes

Migrations to create the memberships table
Separate organisations controllers
Separate organisation index views
Update seed file

## Guidance for review

- The links on the org index pages don't go anywhere, they will probably be implemented when we have show pages for the organisations.
- The placements organisation index page shows providers, not sure if this is correct, someone from placements should tell us.
- If you want to pull this on local please run `bin/setup` or `rake db:seed` after.

## Link to Trello card

https://trello.com/c/1Wl6xPgw/76-if-a-user-belongs-to-multiple-organisations-they-should-be-redirected-to-a-list-of-their-organisations-after-sign-in

## Things to check

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/b9d79e61-637b-4238-99fc-e4829528a678

### DIFF
--- a/app/views/claims/organisations/index.html.erb
+++ b/app/views/claims/organisations/index.html.erb
@@ -1,0 +1,15 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t("organisations") %></h1>
+
+    <% if @schools.any? %>
+      <ul class="govuk-list">
+        <% @schools.each do |school| %>
+          <li>
+           <%= govuk_link_to school.name, request.env["PATH_INFO"], no_visited_state: true %>
+          </li>
+        <% end %>
+      </ul>
+    <% end %>
+  </div>
+</div>
--- a/app/views/placements/organisations/index.html.erb
+++ b/app/views/placements/organisations/index.html.erb
@@ -1,0 +1,27 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t("organisations") %></h1>
+
+    <% if @schools.any? %>
+      <h2 class="govuk-heading-l"><%= t("schools") %></h2>
+      <ul class="govuk-list">
+        <% @schools.each do |school| %>
+        <li>
+          <%= govuk_link_to school.name, request.env["PATH_INFO"], no_visited_state: true %>
+        </li>
+        <% end %>
+      </ul>
+    <% end %>
+
+    <% if @providers.any? %>
+      <h2 class="govuk-heading-l"><%= t("providers") %></h2>
+      <ul class="govuk-list">
+        <% @providers.each do |provider| %>
+          <li>
+            <%= govuk_link_to provider.provider_code, request.env["PATH_INFO"], no_visited_state: true %>
+          </li>
+        <% end %>
+      </ul>
+    <% end %>
+  </div>
+</div>
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,6 +1,10 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the bin/rails db:seed command (or created alongside the database with db:setup).
-#
+SCHOOLS = [
+  { claims: true },
+  { placements: true },
+  { claims: true, placements: true },
+  { claims: true, placements: true },
+].freeze
+
 # Persona Creation (Dummy User Creation)
 Rails.logger.debug "Creating Personas"
 
@@ -12,3 +16,30 @@ Rails.logger.debug "Creating Personas"
 end
 
 Rails.logger.debug "Personas successfully created!"
+
+Rake::Task["gias_update"].invoke unless GiasSchool.any?
+gias_scools = GiasSchool.last(SCHOOLS.count)
+
+SCHOOLS.each.with_index do |school_attributes, index|
+  school = School.find_or_initialize_by(**school_attributes)
+  school.urn = gias_scools[index].urn
+
+  school.save!
+end
+
+User
+  .where(first_name: %w[Anne Patricia])
+  .each do |user|
+    school = School.public_send(user.service).first
+    user.memberships.find_or_create_by!(organisation: school)
+  end
+
+User
+  .where(first_name: %w[Mary Colin])
+  .each do |user|
+    schools = School.where("#{user.service}": true)
+
+    schools.each do |school|
+      user.memberships.find_or_create_by!(organisation: school)
+    end
+  end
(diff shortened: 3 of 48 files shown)


app/views/claims/organisations/index.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t("organisations") %></h1>
+
+    <% if @schools.any? %>
+      <ul class="govuk-list">
+        <% @schools.each do |school| %>
+          <li>
+            <%= govuk_link_to school.name, request.env["PATH_INFO"], class: "govuk-link--no-visited-state" %>
```
ollietreend, line 9: [From the `govuk_link_to` docs](https://govuk-components.netlify.app/helpers/link/#other-link-styles), it looks like we can pass an option to turn off visited state. This might be a cleaner way to do it rather than manually setting the class attribute.

```suggestion
            <%= govuk_link_to school.name, request.env["PATH_INFO"], no_visited_state: true %>
```

app/views/placements/organisations/index.html.erb
```diff
@@ -0,0 +1,27 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= t("organisations") %></h1>
+
+    <% if @schools.any? %>
+      <h2 class="govuk-heading-l"><%= t("schools") %></h2>
+      <ul class="govuk-list">
+        <% @schools.each do |school| %>
+        <li>
+          <%= govuk_link_to school.name, request.env["PATH_INFO"], class: "govuk-link--no-visited-state" %>
```
ollietreend, line 10: Ditto

app/views/placements/organisations/index.html.erb
```diff
@@ -0,0 +1,27 @@
...
+        <% end %>
+      </ul>
+    <% end %>
+
+    <% if @providers.any? %>
+      <h2 class="govuk-heading-l"><%= t("providers") %></h2>
+      <ul class="govuk-list">
+        <% @providers.each do |provider| %>
+          <li>
+            <%= govuk_link_to provider.provider_code, request.env["PATH_INFO"], class: "govuk-link--no-visited-state" %>
```
ollietreend, line 21: Ditto

config/initializers/school.rb
```diff
@@ -0,0 +1,6 @@
+SCHOOLS = [
+  { claims: true },
+  { placements: true },
+  { claims: true, placements: true },
+  { claims: true, placements: true }
+].freeze
```
ollietreend, line 6: I think this might be better located inside the `seeds.rb` file directly. From what I can tell, that's the only place this constant gets used?

I don't think this serves a clear purpose as a global constant, nor does it feel meaningful as a standalone file.

What do you think?

db/seeds.rb
```diff
@@ -12,3 +12,30 @@
 end
 
 Rails.logger.debug "Personas successfully created!"
+
+Rake::Task["gias_update"].invoke unless GiasSchool.any?
+gias_scools = GiasSchool.last(SCHOOLS.count)
+
+SCHOOLS.each.with_index do |school_attributes, index|
```
ollietreend, line 19: For what it's worth, I think the `PERSONAS` constant is more meaningful as a global constant because it is also used for validation in the `Persona` model:

https://github.com/DFE-Digital/itt-mentor-services/blob/a485beef8f5ddedbf32f3484d749162214958fed/app/models/persona.rb#L18-L21

However I think `SCHOOLS` only serves a purpose specifically within this `seeds.rb` file, and doesn't need to be used elsewhere. So I feel like it makes the most sense for it to be located directly where it's used – i.e. in the same file.

spec/system/claims/view_organisations_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  scenario "I sign in as persona Marry" do
+    persona = given_the_claims_persona("Mary")
+    and_persona_has_multiple_organisations(persona)
+    when_i_visit_claims_personas
+    when_i_click_sign_in(persona)
+    i_can_see_a_list_marys_claims_organisations(persona)
+  end
+end
```
ollietreend, line 17: It would be good to also cover the scenario of users with only one organisation. Or is that already covered somewhere else?

```ruby
feature "View organisations " do
  scenario "when the user belongs to multiple organisations" do
    persona = given_the_claims_persona("Mary")
    and_persona_has_multiple_organisations(persona)
    when_i_visit_claims_personas
    when_i_click_sign_in(persona)
    i_can_see_a_list_marys_claims_organisations(persona)
  end

  scenario "when the user only has one organisation" do
    # ...
  end
end
```

spec/system/placements/view_organisations_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  scenario "I sign in as persona Marry" do
+    persona = given_the_placements_persona("Mary")
+    and_persona_has_multiple_organisations(persona)
+    when_i_visit_placements_personas
+    when_i_click_sign_in(persona)
+    i_can_see_a_list_marys_placements_organisations(persona)
+  end
+end
```
ollietreend, line 17: Ditto

spec/system/claims/view_organisations_spec.rb
```diff
@@ -0,0 +1,44 @@
+require "rails_helper"
+
+feature "View organisations as a multi org user" do
```
ollietreend, line 3: Should this be a **system test** rather than a **feature test**?

```suggestion
RSpec.describe "View organisations", type: :system do
```

ollietreend (changes requested):
@CatalinVoineag this is looking great – really nice work. It's exciting to see the User -> Membership -> Organisation association taking shape.

I've left a few comments of things that I think might need some more attention. I'm very happy to talk any of them through if you'd like to. 🙇🏻

spec/system/claims/view_organisations_spec.rb
```diff
@@ -0,0 +1,60 @@
+require "rails_helper"
+
+RSpec.describe "View organisations", type: :system do
+  around do |example|
+    Capybara.app_host = "https://#{ENV["CLAIMS_HOST"]}"
+    example.run
+    Capybara.app_host = nil
+  end
+
+  scenario "I sign in as persona Marry with multiple organistions" do
```
ollietreend, line 10:
```suggestion
  scenario "I sign in as persona Mary with multiple organistions" do
```

ollietreend (changes requested):
One teeny tiny nitpick!

PR #48, opened 2023-12-21
Placements Support Add Providers

JamieCleare2525:
## Context

Support users should be able to "Onboard" Providers from the Publish API
Based on https://sp-beta-pr-27.herokuapp.com/support/organisations/new/provider

- API service for collecting Provider data from Publish API
- Controllers/Views for New/Check/Create Providers

## Guidance to review

- run `bundle install`
- run `yarn install`

- Sign in as Colin
- Visit `http://placements.localhost:3000/support/providers/new`
- Type a few letters into the text input (e.g "CT")
- Select an option from the dropdown list.
- Click "Continue"
- _This should take you to the "Check your answers" page_
- Click "Add organisation"

## Link to Trello card

https://trello.com/c/9W5FDi7E/67-add-a-new-provider

## Things to check

## Screenshots
_Happy Path(Provider doesn't exist)_

https://github.com/DFE-Digital/itt-mentor-services/assets/17162488/3ab112c6-2ed7-47cd-8e6c-a69d78650bba

_Unhappy Path(Provider code already exists)_

https://github.com/DFE-Digital/itt-mentor-services/assets/17162488/abdf6638-2dd6-4d8a-9a44-30101c2a8647

### DIFF
--- a/app/controllers/placements/support/provider_suggestions_controller.rb
+++ b/app/controllers/placements/support/provider_suggestions_controller.rb
@@ -1,0 +1,37 @@
+class Placements::Support::ProviderSuggestionsController < Placements::Support::ApplicationController
+  def index
+    accredited_providers = AccreditedProviderApi.call
+    filtered_providers = filter_providers(accredited_providers, query_params)
+    providers =
+      filtered_providers.map { |provider| formatted_provider(provider) }
+    render json: providers
+  end
+
+  private
+
+  def query_params
+    params.require(:query)
+  end
+
+  def filter_providers(providers, query)
+    return providers if query.blank?
+
+    downcase_query = query.downcase
+    providers.select do |provider|
+      [
+        provider.dig("attributes", "name"),
+        provider.dig("attributes", "postcode"),
+        provider.dig("attributes", "urn"),
+        provider.dig("attributes", "ukprn"),
+      ].any? { |attribute| attribute&.downcase&.include?(downcase_query) }
+    end
+  end
+
+  def formatted_provider(provider)
+    {
+      id: provider.fetch("id"),
+      name: provider.dig("attributes", "name"),
+      code: provider.dig("attributes", "code"),
+    }
+  end
+end
--- a/app/controllers/placements/support/providers_controller.rb
+++ b/app/controllers/placements/support/providers_controller.rb
@@ -1,0 +1,28 @@
+class Placements::Support::ProvidersController < Placements::Support::ApplicationController
+  def new
+    @provider = Provider.new
+  end
+
+  def create
+    @provider = Provider.new(provider_params)
+    if @provider.save
+      redirect_to placements_support_organisations_path
+    else
+      render :new
+    end
+  end
+
+  def check
+    @provider = Provider.new(provider_code: params[:accredited_provider_id])
+    @provider_details = AccreditedProviderApi.call(@provider.provider_code)
+    return if @provider.valid? && @provider_details.present?
+
+    render :new
+  end
+
+  private
+
+  def provider_params
+    params.require(:provider).permit(:provider_code)
+  end
+end
--- a/app/javascript/utils/request_helper.js
+++ b/app/javascript/utils/request_helper.js
@@ -1,0 +1,32 @@
+const getPath = (endpoint, query) => {
+  return `${endpoint}?query=${query}`;
+};
+
+const request = (endpoint) => {
+  let xhr = null; // Hoist this call so that we can abort previous requests.
+
+  return (query, callback) => {
+    if (xhr && xhr.readyState !== XMLHttpRequest.DONE) {
+      xhr.abort();
+    }
+    const path = getPath(endpoint, query);
+
+    xhr = new XMLHttpRequest();
+    xhr.addEventListener("load", (evt) => {
+      let results = [];
+      try {
+        results = JSON.parse(xhr.responseText);
+      } catch (err) {
+        console.error(
+          `Failed to parse results from endpoint ${path}, error is:`,
+          err,
+        );
+      }
+      callback(results);
+    });
+    xhr.open("GET", path);
+    xhr.send();
+  };
+};
+
+export { getPath, request };
--- a/app/services/accredited_provider_api.rb
+++ b/app/services/accredited_provider_api.rb
@@ -1,0 +1,47 @@
+class AccreditedProviderApi
+  include ServicePattern
+
+  def initialize(code = nil)
+    @code = code
+  end
+
+  attr_reader :code
+
+  def call
+    code.present? ? provider_list(code) : provider_details
+  end
+
+  private
+
+  def provider_list(code)
+    Rails
+      .cache
+      .fetch("accredited_provider_details_#{code}", expires_in: 24.hours) do
+        response = HTTParty.get(provider_details_url(code))
+        response = JSON.parse(response.to_s)
+        response["data"]
+      end
+  end
+
+  def provider_details
+    Rails
+      .cache
+      .fetch("all_accredited_providers", expires_in: 24.hours) do
+        response = HTTParty.get(all_providers_url)
+        response = JSON.parse(response.to_s)
+        response["data"]
+      end
+  end
+
+  def all_providers_url
+    "#{ENV["PUBLISH_BASE_URL"]}/api/public/v1/recruitment_cycles/#{next_year}/providers?filter[is_accredited_body]=true"
+  end
+
+  def provider_details_url(code)
+    "#{ENV["PUBLISH_BASE_URL"]}/api/public/v1/recruitment_cycles/#{next_year}/providers/#{code}"
+  end
+
+  def next_year
+    Time.current.next_year.year
+  end
+end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -36,6 +36,7 @@
     </div>
 
     <div class="govuk-width-container">
+      <%= yield :before_content %>
       <main class="govuk-main-wrapper" id="main-content" role="main">
         <%= yield %>
       </main>
--- a/spec/system/placements/providers/support_user_adds_a_provider_spec.rb
+++ b/spec/system/placements/providers/support_user_adds_a_provider_spec.rb
@@ -1,0 +1,151 @@
+require "rails_helper"
+
+# Changing type from feature => system stops the javascript working.
+# Will attempt to resolve in another PR.
+RSpec.describe "Placements / Providers / Support User adds a Provider",
+               type: :feature do
+  before do
+    next_year = Time.current.next_year.year
+
+    stub_request(
+      :get,
+      "https://www.publish-teacher-training-courses.service.gov.uk/api/public/v1/recruitment_cycles/#{next_year}/providers?filter%5Bis_accredited_body%5D=true",
+    ).to_return(
+      status: 200,
+      body: {
+        "data" => [
+          {
+            "id" => 123,
+            "attributes" => {
+              name: "Provider 1",
+              code: "Prov1",
+            },
+          },
+          {
+            "id" => 234,
+            "attributes" => {
+              name: "Provider 2",
+              code: "Prov2",
+            },
+          },
+        ],
+      }.to_json,
+    )
+
+    stub_request(
+      :get,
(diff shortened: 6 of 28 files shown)


app/services/accredited_provider_api.rb
```diff
@@ -0,0 +1,20 @@
...
+        response = JSON.parse(response.to_s)
+        response["data"]
+      end
+  end
+
+  private
+
+  def provider_url
+    year = Time.current.next_year.year
+    "https://www.publish-teacher-training-courses.service.gov.uk/api/public/v1/recruitment_cycles/#{year}/providers?filter[is_accredited_body]=true"
```
CatalinVoineag, line 18: Can the base URL be an ENV variable?

app/services/accredited_provider_api.rb
```diff
@@ -0,0 +1,20 @@
+class AccreditedProviderApi
+  include ServicePattern
+
+  def call
```
CatalinVoineag, line 4: Do we need to handle a 500 or any other response that it's not 200 here? What happens if publish is down?

app/controllers/placements/support/provider_suggestions_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def formatted_provider(provider)
+    {
+      id: provider["id"],
+      name: provider["attributes"]["name"],
+      code: provider["attributes"]["code"]
+    }
+  end
+end
```
CatalinVoineag, line 33: This is more of a personal opinion but I think we should use {}.fetch here as it offers more descriptive errors when keys are missing, also it doesn't fail silently if the keys are missing. It's not really a blocker though.

app/controllers/placements/support/providers_controller.rb
```diff
@@ -0,0 +1,33 @@
+class Placements::Support::ProvidersController < Placements::Support::ApplicationController
+  def new
+    @provider = Provider.new
+  end
+
+  def create
+    @provider = Provider.find_or_initialize_by(provider_params)
```
CatalinVoineag, line 7: Not sure if we need find or initialize here. I feel like we always want to create a new record in the `create` controller method

app/controllers/placements/support/providers_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+      redirect_to placements_support_organisations_path
+    else
+      render :new
+    end
+  end
+
+  def check
+    @provider = Provider.new(provider_code: params[:accredited_provider_id])
+    if @provider.valid?
+      accredited_providers = AccreditedProviderApi.call
```
CatalinVoineag, line 18: I assume Publish doesn't have an endpoint to get only one provider by an id? I'm wondering if we need to get all of them again

app/controllers/placements/support/provider_suggestions_controller.rb
```diff
@@ -0,0 +1,33 @@
+class Placements::Support::ProviderSuggestionsController < Placements::Support::ApplicationController
+  def index
+    accredited_providers = AccreditedProviderApi.call
+    filtered_providers = filter_providers(accredited_providers, params[:query])
```
CatalinVoineag, line 4: Is it worth doing `params.require(:query)`?

CatalinVoineag (commented):
Just a few comments. I pulled this on my local and it does work as expected. Good work 👍🏻

app/services/accredited_provider_api.rb
ollietreend: @CatalinVoineag I'm going to be following up with a more comprehensive implementation of the Provider API ([Trello card](https://trello.com/c/3iEk7hif/78-retrieve-providers-from-the-teacher-training-courses-api)) which will supersede this. But we went ahead with this implementation for now to avoid my card being a blocker.

It's worth noting that the response is cached, so we'll be somewhat protected from downtime just because we won't need to regularly hit the endpoint.

CatalinVoineag: Yeah, that makes sense. Thank you 👍🏻

app/javascript/utils/request_helper.js
```diff
@@ -0,0 +1,32 @@
...
+          `Failed to parse results from endpoint ${path}, error is:`,
+          err,
+        );
+      }
+      callback(results);
+    });
+    xhr.open("GET", path);
+    xhr.send();
+  };
+};
```
ollietreend, line 30: Any reason why we aren't using the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) instead of creating our own wrapper for XMLHttpRequest?

I don't think browser support should be a problem for us. Especially because we're already using ES6 syntax, and I don't think `esbuild` [has been configured](https://github.com/DFE-Digital/itt-mentor-services/blob/f42e69303f7bd251f41ca38048d4af2cb5010cb9/package.json#L11) to transpile that to ES5 for older browsers.

app/models/provider.rb
```diff
@@ -13,5 +13,9 @@
 #
 class Provider < ApplicationRecord
   validates :provider_code, presence: true
-  validates :provider_code, uniqueness: { case_sensitive: false }
+  validates :provider_code,
+            uniqueness: {
+              case_sensitive: false,
+              message: "Provider already exists!"
```
ollietreend, line 19: I'm sure our (soon to join) content designer will want to configure custom/meaningful error messages.

However, as a general rule of thumb:

1. I don't think we should use exclamation marks in error messages. It feels a bit shouty!!! 🙃
2. I think we should aim to [use locale files to set custom validation messages](https://guides.rubyonrails.org/i18n.html#error-message-scopes) rather than hardcoding them in the models. I believe this same message could be set using the locale key `en.activerecord.errors.models.provider.attributes.provider_code.taken`.

Localised error messages give us more flexibility if we want to [customise the full message](https://edgeguides.rubyonrails.org/configuring.html#configuring-active-model) further down the line. For example, as was needed in [a previous Rails project I worked on](https://github.com/alphagov/whitehall/pull/8361).

app/views/layouts/application.html.erb
```diff
@@ -36,6 +36,7 @@
     </div>
 
     <div class="govuk-width-container">
+      <%= yield :before_content %>
```
ollietreend, line 39: I wonder if we could call this content block `:back_link`. I suspect that'll be our only use for it.

What do you think?

ollietreend (commented):
@JamieCleare2525 this is only a partial review – lunch is calling! 🙈

But I'm submitting it now to give you the feedback as soon as possible.

Nothing in here is a necessity – so I'm just leaving them as "comments" rather than "requested changes".

app/javascript/utils/request_helper.js
JamieCleare2525: This is extracted directly from https://github.com/DFE-Digital/publish-teacher-training/blob/main/app/javascript/utils/request_helper.js
As this was what Publish was doing, I thought it best to copy their implementation.

app/views/layouts/application.html.erb
JamieCleare2525: I've once again stolen this from Publish https://github.com/search?q=repo%3ADFE-Digital%2Fpublish-teacher-training%20before_content&type=code
As they use it for breadcrumbs (not sure yet if we are using breadcrumbs) I didn't think it was worth changing it.

app/services/accredited_provider_api.rb
JamieCleare2525: resolved in [abdfae3](https://github.com/DFE-Digital/itt-mentor-services/pull/48/commits/abdfae3e4ab05b2c4cf0a45a0c634fec1fbf6545)

app/controllers/placements/support/provider_suggestions_controller.rb
JamieCleare2525: resolved in [abdfae3](https://github.com/DFE-Digital/itt-mentor-services/pull/48/commits/abdfae3e4ab05b2c4cf0a45a0c634fec1fbf6545)

app/controllers/placements/support/providers_controller.rb
JamieCleare2525: resolved in [abdfae3](https://github.com/DFE-Digital/itt-mentor-services/pull/48/commits/abdfae3e4ab05b2c4cf0a45a0c634fec1fbf6545)

app/controllers/placements/support/provider_suggestions_controller.rb
JamieCleare2525: resolved in [abdfae3](https://github.com/DFE-Digital/itt-mentor-services/pull/48/commits/abdfae3e4ab05b2c4cf0a45a0c634fec1fbf6545)

app/models/provider.rb
JamieCleare2525: resolved in [3b78a77](https://github.com/DFE-Digital/itt-mentor-services/pull/48/commits/3b78a77c049f64ecf0a81a96fc412ccf8260cce1)

app/javascript/utils/request_helper.js
ollietreend: Seems sensible. Let's use that for now then – it's not doing any harm.

package-lock.json
ollietreend, line 1: Ooops – I don't think this file should exist.

Did you run `npm install` instead of `yarn install`?

ollietreend (changes requested):
Sorry – a few more comments!

app/services/accredited_provider_api.rb
```diff
@@ -0,0 +1,47 @@
...
+    "#{ENV["PUBLISH_BASE_URL"]}/api/public/v1/recruitment_cycles/#{next_year}/providers?filter[is_accredited_body]=true"
+  end
+
+  def provider_details_url(code)
+    "#{ENV["PUBLISH_BASE_URL"]}/api/public/v1/recruitment_cycles/#{next_year}/providers/#{code}"
+  end
+
+  def next_year
+    Time.current.next_year.year
+  end
```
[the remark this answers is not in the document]
JamieCleare2525: This is best guess based on the conversations on Slack. We can change it should this be incorrect.

lib/hosting_environment.rb
```diff
@@ -1,7 +1,7 @@
 module HostingEnvironment
   PRODUCTION_BANNER_NAME = {
     claims: "beta",
-    placements: "beta"
+    placements: "beta",
```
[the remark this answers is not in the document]
JamieCleare2525: This appears to be the way of things.
https://ukgovernmentdfe.slack.com/archives/C06962474AC/p1703259713095079

bin/bundle
```diff
@@ -102,7 +102,7 @@ m =
       require_error = activation_error_handling { require "bundler/version" }
       if require_error.nil? &&
            Gem::Requirement.new(bundler_requirement).satisfied_by?(
-             Gem::Version.new(Bundler::VERSION)
+             Gem::Version.new(Bundler::VERSION),
```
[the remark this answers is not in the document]
JamieCleare2525: This appears to be the way of things.
https://ukgovernmentdfe.slack.com/archives/C06962474AC/p1703259713095079

config/initializers/persona.rb
```diff
@@ -6,15 +6,15 @@
   {
     first_name: "Patricia",
     last_name: "Adebayo",
-    email: "patricia@example.com"
+    email: "patricia@example.com",
```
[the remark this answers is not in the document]
JamieCleare2525: This appears to be the way of things.
https://ukgovernmentdfe.slack.com/archives/C06962474AC/p1703259713095079

spec/system/placements/providers/support_user_adds_a_provider_spec.rb
```diff
@@ -0,0 +1,148 @@
+require "rails_helper"
+
+RSpec.feature "Placements / Providers / Support User adds a Provider" do
```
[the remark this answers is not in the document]
JamieCleare2525: Turns out changing this to a system stops the javascript from working. I'll try and resolve this in a separate PR.

ollietreend (approved):

ollietreend:
@JamieCleare2525 can you make the commit history pretty before merging this please? 🙏🏻 Thank you.

[robinsons-squash.jpg]

PR #53, opened 2023-12-22
template without links

SocFoot:
model and template

template model and spec

## Things to check


app/assets/stylesheets/application.sass.scss
```diff
@@ -3,5 +3,86 @@ $govuk-new-link-styles: true;
 $govuk-assets-path: "";
 
 @import "govuk-frontend/dist/govuk/all";
-
 @import "components/all";
+// // @import "govuk-frontend/govuk/base";
```
JamieCleare2525, line 7: Can we get rid of all the commented out imports?

app/controllers/claims/pages_controller.rb
```diff
@@ -1,2 +1,5 @@
 class Claims::PagesController < ApplicationController
+  def index
+    @school = GiasSchool.first
```
JamieCleare2525, line 3: I think this should be `current_user.schools`

app/helpers/navigation_bar_helper.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module NavigationBarHelper
+  def render_navigation_bar?(provider)
+    request.path.exclude?("support") && provider && !current_page?(root_path) &&
+      !current_page?(publish_provider_path(provider.provider_code)) &&
+      provider.recruitment_cycle
```
JamieCleare2525, line 7: `Provider` in our app doesn't have `recruitment_cycle`. So I'm assuming this should break.

spec/factories/gias_schools.rb
```diff
@@ -24,5 +24,11 @@
   factory :gias_school do
     sequence(:urn) { _1 }
     name { "Hogwarts" }
+    ukprn { "fake_uprn" }
```
JamieCleare2525, line 27: At this point I think we should use Faker or sequence. Otherwise if someone uses `create_list(:gias_school, 5)` we'll end up with 5 GiasSchools with the same attributes.

app/components/navigation_bar.html.erb
```diff
@@ -0,0 +1,22 @@
+<div class="app-primary-navigation">
+    <div class="app-primary-navigation govuk-!-display-none-print">
+    <div class="govuk-width-container app-primary-navigation--justify-between">
+        <nav class="app-primary-navigation__nav" aria-label="main menu">
+        <ul class="app-primary-navigation__list">
+            <li class="app-primary-navigation__item">
+                <a class="app-primary-navigation__link" href="#">Claims</a>
```
elceebee, line 7: I believe we agreed to use localisation for strings, so this should be
```suggestion
                <a class="app-primary-navigation__link" href="#"><%= t("claims") %></a>
```

or similar.

Also, if this is meant to be a claims-specific component, it should before be named `claims_navigation_bar.html.erb`. If it's meant to be generic, it should probably take the primary nav title as an argument.

app/views/claims/pages/index.html.erb
```diff
@@ -1,26 +1,49 @@
...
-    <h1 class="govuk-heading-xl">It works! 🎉</h1>
+<%= render(NavigationBar.new(
+    items: navigation_items(@provider),
+    current_path: request.path,
+    current_user: User.first
+  )) %>
+    <div class="govuk-grid-row">
+    <br><br>
+      <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">Organisation details</h1>
```
elceebee, line 9:
```suggestion
      <h1 class="govuk-heading-l"><%= t("organisation_details") %></h1>
```

or similar, with corresponding translation.

spec/features/personas/sign_in_as_a_claims_user_persona_spec.rb
```diff
@@ -50,6 +54,10 @@ def given_there_is_an_existing_claims_persona_for(persona_name)
   create(:persona, persona_name.downcase.to_sym, service: "claims")
 end
 
+def and_given_i_have_a_school
+  create(:gias_school)
+end
```
elceebee, line 59: I think users will have a `School`, not a `GiasSchool`. You are likely to have started this before the `School` model and relationship with ` `GiasSchool` was in place.

spec/system/home_page_spec.rb
```diff
@@ -27,6 +28,10 @@ def given_i_am_on_the_placements_site
     Capybara.app_host = "http://#{ENV["PLACEMENTS_HOST"]}"
   end
 
+  def and_given_i_have_a_school
+    create(:gias_school)
+  end
```
elceebee, line 33: As above, I think the claims users will have a `School`, not a `GiasSchool`.

elceebee (changes requested):
A few translation things which aren't a big deal, but I definitely think the organisation needs to be the `School` model, not the `GiasSchool`. 

I've tested it locally and it works as expected.

app/views/claims/pages/index.html.erb
```diff
@@ -1,26 +1,49 @@
...
+        <%= govuk_summary_list do |summary_list|
+          summary_list.with_row do |row|
+          row.with_key(text: 'Email address')
+          row.with_value do %>
+            <a href="contact@example.com">contact@example.com</a>
+          <% end;row.with_action(text: 'Change', visually_hidden_text: 'email address', href: '/contact-details')
+          end;summary_list.with_row do |row|
+            row.with_key(text: 'Telephone number')
+            row.with_value(text: @school.telephone)
+            row.with_action(text: 'Change', visually_hidden_text: 'email address', href: '/contact-details')
```
CatalinVoineag, line 32: This is more of a personal opinion but I feel like we should not redirect a user to a non-existent path. I usually just redirect to the current path until the path that we want is implemented.

app/views/claims/pages/index.html.erb
```diff
@@ -1,26 +1,49 @@
...
-    ) %>
-  </div>
-</div>
+        <h2 class="govuk-heading-m govuk-!-margin-top-9">Contact details</h2>
+        <%= govuk_summary_list do |summary_list|
+          summary_list.with_row do |row|
+          row.with_key(text: 'Email address')
+          row.with_value do %>
+            <a href="contact@example.com">contact@example.com</a>
+          <% end;row.with_action(text: 'Change', visually_hidden_text: 'email address', href: '/contact-details')
```
CatalinVoineag, line 28: Can we use translations for the text and visually hidden text? Same with the other actions

app/views/claims/pages/index.html.erb
```diff
@@ -1,26 +1,49 @@
...
+    current_path: request.path,
+    current_user: User.first
+  )) %>
+    <div class="govuk-grid-row">
+    <br><br>
+      <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l">Organisation details</h1>
+      <%= govuk_summary_list do |summary_list|
+        summary_list.with_row do |row|
+        row.with_key(text: "Organisation name")
```
CatalinVoineag, line 12: Can we use translations for these texts?

CatalinVoineag (commented):
Good work. I'm a bit confused on the user flow we need here. Maybe you didn't pull the latest version of main. But I expected to have a `organisations#show` controller and route here so we can redirect the user to the organisations details from [here](https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/views/claims/organisations/index.html.erb#L9) if he has multiple organisations or from [here](https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/controllers/application_controller.rb#L23C7-L23C25) if he has only one organisation.

Happy to jump on a call for more detail.

app/controllers/claims/pages_controller.rb
SocFoot: My take on this ticket is that it is mainly a "front end" ticket so i tried to not touch business logic and the association between schools and user being absent I decided to use static data. I guess we could decide to add the business logic here and increase the size of the PR or to have a second ticket. I don't mind

elceebee: Perhaps it's something to discuss as a group -- we don't have properly written up tickets / ACs / requirements, so it's difficult to know where to draw the lines. My view is that seeing the relevant school is key to the feature, so should be implemented here.

app/controllers/claims/pages_controller.rb
```diff
@@ -1,2 +1,5 @@
 class Claims::PagesController < ApplicationController
+  def index
+    @school = User.first&.schools&.first&.gias_school
```
JamieCleare2525, line 3:
```suggestion
    @school = current_user.schools
```
If the school needs additional information from the GiasSchool association delegate it.

app/helpers/navigation_bar_helper.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module NavigationBarHelper
+  def navigation_items(_provider)
+    [
+      {
+        name: t("navigation_bar.users"),
+        url: "#",
+        additional_url:
+          "request_access_publish_provider_path(provider.provider_code)",
```
JamieCleare2525, line 10: should this be in a string?

app/views/claims/pages/index.html.erb
```diff
@@ -1,28 +1,51 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">It works! 🎉</h1>
+<%= render(NavigationBar.new(
+    items: navigation_items(@provider),
+    current_path: request.path,
+    current_user: User.first,
```
JamieCleare2525, line 4: `current_user` is a helper method which can be used in the views and controllers.

app/views/claims/pages/index.html.erb
```diff
@@ -1,28 +1,51 @@
...
+  )) %>
```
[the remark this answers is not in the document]
Nitemaeric: I think we can also adopt the pattern established with Gov UK of ...

```erb
<%= navigation_bar do |navbar| %>
  <% navbar.with_navigation_item(text: "Text", url: "/") %>
<% end %>
```

app/views/claims/pages/index.html.erb
```diff
@@ -1,28 +1,51 @@
...
+              <% end;row.with_action(text: t('change'), visually_hidden_text:  t('email_address'), href: '/contact-details')
+            end;summary_list.with_row do |row|
+              row.with_key(text: t('address'))
+              row.with_value do %>
+                <p class="govuk-body">
+                  <%= @school.address1 %><br>
+                  <%= @school.address2 %><br>
+                  <%= @school.address3 %><br>
+              </p>
+              <% end;row.with_action(text:  t('change'), visually_hidden_text:  t('email_address'), href: '/contact-details')
```
Nitemaeric, line 47: Hmm, are we sure this is linted correctly? Many lines start with `end;`.

app/views/layouts/application.html.erb
```diff
@@ -25,7 +25,9 @@
 
     <%= govuk_skip_link %>
 
-    <%= govuk_header(homepage_url: "https://www.gov.uk", service_name: t(".#{current_service}.header.service_name"), service_url: "/") %>
+    <%= govuk_header(homepage_url: "https://www.gov.uk") do |header| %>
+      <%= header.with_product_name(name: t(".#{current_service}.header.service_name")) %>
+    <% end %>
```
Nitemaeric, line 30: This was addressed in a recent PR.

spec/features/personas/sign_in_as_a_claims_user_persona_spec.rb
```diff
@@ -46,8 +46,8 @@
 
 private
 
-def given_there_is_an_existing_claims_persona_for(persona_name)
-  create(:persona, persona_name.downcase.to_sym, service: "claims")
+def given_there_is_an_existing_claims_persona_for_with_a_school(persona_name)
```
Nitemaeric, line 49:
```suggestion
def given_there_is_an_existing_claims_persona_with_a_school_for(persona_name)
```

spec/system/home_page_spec.rb
```diff
@@ -44,7 +44,7 @@ def i_can_see_the_placements_service_name_in_the_header
   end
 
   def and_i_can_see_the_claims_service_value_in_the_page
-    expect(page).to have_css(".govuk-summary-list__value", text: "Claims")
+    expect(page).to have_css(".app-primary-navigation__link", text: "Claims")
   end
 
   def and_i_can_see_the_placements_service_value_in_the_page
```
Nitemaeric, line 50: This may mean we don't need this spec anymore, right? As there is no longer a "home page".

app/controllers/claims/pages_controller.rb
```diff
@@ -1,2 +1,5 @@
...
+  end
```
Nitemaeric, line 4: If this isn't a meaningful lookup, let's use something more straight-forward with a comment.

```suggestion
  def index
    @school = Claims::School.first # Placeholder
  end
```

app/components/application_component.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
elceebee, line 1:
```suggestion

```

We agreed in an earlier PR that we don't need the frozen_string_literal comment at the top of each class.

app/views/claims/pages/index.html.erb
elceebee: I am finding this difficult to read -- not sure why the linter allowed this. And sometimes we have `end;row` and sometimes we have `end; row`.

CatalinVoineag:
> Good work. I'm a bit confused on the user flow we need here. Maybe you didn't pull the latest version of main. But I expected to have a `organisations#show` controller and route here so we can redirect the user to the organisations details from [here](https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/views/claims/organisations/index.html.erb#L9) if he has multiple organisations or from [here](https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/controllers/application_controller.rb#L23C7-L23C25) if he has only one organisation.
> 
> Happy to jump on a call for more detail.

This has been changed to `schools#index` now for the claims domain

app/controllers/claims/pages_controller.rb
```diff
@@ -1,2 +1,5 @@
 class Claims::PagesController < ApplicationController
+  def index
+    @user = User.first
```
Nitemaeric, line 3: I've just realised. Is this the correct page?

The organisation details should be for the `/schools/:school_id` page, right?

app/views/claims/pages/index.html.erb
```diff
@@ -1,29 +1,56 @@
...
+<%= navigation_bar %>
+<% if @user&.schools&.first&.gias_school %>
+    <div class="govuk-grid-row">
+    <br><br>
+      <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l"><%= t("organisation_details") %></h1>
+      <%= govuk_summary_list do |summary_list|
+            summary_list.with_row do |row|
+              row.with_key(text: t("organisation_name"))
+              row.with_value(text: @user&.schools&.first&.gias_school&.name)
```
Nitemaeric, line 10: To respect the Law of Demeter , we should delegate the attributes we want to access in the `gias_school` model - I've just checked, we already do this.

To also avoid `nil` errors, we can add the following:

```ruby
  belongs_to :gias_school, foreign_key: :urn, primary_key: :urn

  delegate :name, to: :gias_school, allow_nil: true # Add allow_nil
end
```

This will allow you to do `school.name` and it will return `nil` when the `gias_school` association is missing.

As for `@user`, shouldn't that always be the `current_user` in this case?

Also, if you're using the `/schools/:school_id` path, you should be able to do:

```ruby
class Claims::SchoolsController < ApplicationController
  before_action :set_school

  private

  def set_school
    current_user.schools.find(params[:id])
  end
end
```

If the User does not belong to the School, the `#find` call will raise a not found error.

app/views/claims/pages/index.html.erb
```diff
@@ -1,29 +1,56 @@
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
-    <h1 class="govuk-heading-xl">It works! 🎉</h1>
+<%= navigation_bar %>
+<% if @user&.schools&.first&.gias_school %>
+    <div class="govuk-grid-row">
+    <br><br>
+      <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l"><%= t("organisation_details") %></h1>
```
Nitemaeric, line 6: Using view-scoped keys prevents namespace and key conflicts.
```suggestion
      <h1 class="govuk-heading-l"><%= t(".heading") %></h1>
```

app/views/claims/pages/index.html.erb
```diff
@@ -1,29 +1,56 @@
...
-      ],
-    ) %>
-  </div>
-</div>
+        <h2 class="govuk-heading-m govuk-!-margin-top-9">Contact details</h2>
+        <%= govuk_summary_list do |summary_list|
+          summary_list.with_row do |row|
+          row.with_key(text: t("email_address"))
+          row.with_value do %>
+            <a href="contact@example.com"><%= @user.email %></a>
```
Nitemaeric, line 27: The indentation here seems off?

spec/system/home_page_spec.rb
```diff
@@ -15,6 +15,18 @@
 
   private
 
+  def given_i_am_on_the_claims_site
+    Capybara.app_host = "http://#{ENV["CLAIMS_HOST"]}"
+  end
+
+  def given_i_am_on_the_placements_site
+    Capybara.app_host = "http://#{ENV["PLACEMENTS_HOST"]}"
+  end
```
Nitemaeric, line 24: These methods are already defined in the `spec/support` directory.

Nitemaeric (changes requested):

app/views/claims/pages/index.html.erb
ollietreend: > This will allow you to do school.name and it will return nil when the gias_school association is missing.

Just to chime in here – I think we should avoid using `allow_nil: true` here until we have a clear need for it.

A School should _always_ have an associated GiasSchool. And if it doesn't, something has gone wrong. Given that this should only happen in exceptional circumstances, I'd rather we end up finding the exception in Sentry rather than it silently failing and users are presented with schools that have no name.

So to be clear: I think it should be entirely safe to assume we can always call `school.name` without having to worry about missing GiasSchool associations.

SocFoot: The relationship between user and school didn't exist when i started the ticket. The path is NOT /schools/:school_id. It is under /. In the ticket they were no mention or indication that's what we would want. If you take a look at the prototype. When you sign in for school you only have access to one school. 
Most of this changes are necessary but probably better for a another ticket . I m going to use fake data for now and for clarity sake.

Nitemaeric: Please move the code into the `/schools/:school_id` path. The data can be entirely fake.

app/views/claims/pages/index.html.erb
Nitemaeric, line 1: Please remove this deletion from the set of changes.

app/controllers/claims/pages_controller.rb
Nitemaeric, line 1: This file also doesn't need to change

config/routes/claims.rb
```diff
@@ -1,7 +1,7 @@
 scope module: :claims, as: :claims, constraints: { host: ENV["CLAIMS_HOST"] } do
-  root to: "pages#index"
+  root to: "schools#show"
```
Nitemaeric, line 2: Please leave this as `pages#index`. We will update the `after_sign_in_path` in another PR.

spec/system/claims/pages/home_spec.rb
Nitemaeric, line 1: By moving the code in `schools#show`, this is no longer the "Home Page" spec. It should be the "School Page" spec. Or `view_school_spec` to match the `view_organisations_spec` we have.

Nitemaeric (changes requested):
The user redirection flow will be

```
Sign In -> Organisations -> Organisation (if a user only has 1 Org)
```

This will also handle the case of when a user lands on the Organisations page and they will be redirected.

For now, we'll keep the root path in, but we can remove it once we know what we want to do for sure.

config/locales/en.yml
```diff
@@ -19,8 +19,11 @@ en:
             gias_school:
               required: ""
   add_organisation: Add organisation
+<<<<<<< HEAD
   organisations: Organisations
   school: School
+=======
+>>>>>>> lint
```
Nitemaeric, line 26: Please remove conflict artefacts.

app/views/claims/schools/show.html.erb
```diff
@@ -0,0 +1,54 @@
+<%= navigation_bar %>
+<div class="govuk-grid-row">
+<br><br>
+    <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l"><%= t("claims.pages.index.heading") %></h1>
```
Nitemaeric, line 5: Why is this page pointing to the `index` controller action? 😅

spec/system/claims/view_school_spec.rb
```diff
@@ -0,0 +1,54 @@
+require "rails_helper"
+
+RSpec.feature "Home Page" do
```
Nitemaeric, line 3:
```suggestion
RSpec.feature "School Page" do
```

spec/system/home_page_spec.rb
Nitemaeric, line 1: I think you can remove the added methods from this spec.

Nitemaeric (changes requested):

config/locales/en.yml
```diff
@@ -71,6 +71,15 @@ en:
       heading: Placements news and updates
     contents: Contents
   claims:
+    schools:
+      index:
```
Nitemaeric, line 75:
```suggestion
      show:
```

spec/system/claims/view_school_spec.rb
```diff
@@ -0,0 +1,54 @@
+require "rails_helper"
+
+RSpec.feature "School Page" do
```
Nitemaeric, line 3:
```suggestion
RSpec.describe "School Page", type: :system do
```

spec/system/claims/view_school_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  after { Capybara.app_host = nil }
```
Nitemaeric, line 4: Once the above is added, this is already handled by the Rails Helper.

spec/system/claims/view_school_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    when_i_visit_a_school_page
+    i_can_see_organisation_details
+    i_can_see_contact_details
+  end
+
+  private
+
+  def given_i_am_on_the_claims_site
+    Capybara.app_host = "http://#{ENV["CLAIMS_HOST"]}"
+  end
```
Nitemaeric, line 20: This is already handled by the services test helper that is included into System specs.

Nitemaeric (approved):
Thanks for going through each of the issues!

PR #75, opened 2024-01-03
Clean up User association definitions

Nitemaeric:
## Context

The associations were difficult to visually parse.

As inlining these does not go against our RuboCop lint rules, bring the definitions inline.

## Guidance to review

Does it read nicer? 😄 

## Things to check

## Screenshots

CleanShot 2024-01-03 at 17 05 26@2x

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -17,18 +17,11 @@
 #
 class User < ApplicationRecord
   has_many :memberships
-  has_many :schools,
-           through: :memberships,
-           source: :organisation,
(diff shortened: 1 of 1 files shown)


elceebee (commented):
I find the first way easier to read personally. The extra space on line 21 between `schools,` and `through` is especially jarring for me to read. :)

In general, I don't think we should reformat code like this. I could change it tomorrow because I think it reads easier over multiple lines and we could end up in an infinite loop of reformatting.

PR #73, opened 2024-01-03
Jc/add schools for support user

JamieCleare2525:
## Context

Support Users require a way of onboarding schools.

## Changes proposed in this pull request

Replicating the prototype: https://itt-mentoring-beta-94341ca35a2c.herokuapp.com/support/organisations/new
Support users can now:
- Select a type of organisation to add (ITT Provider or School).
- Search for a School (Dropdown provides a list of GIAS schools).
- Check the Schools Details
- Add (onboard) a school.

## Guidance to review

- Sign in as Colin
- Click "Add Organisation"
- Select "ITT provider" radio button
- Click "Continue"
- Search for a school
- Select a school from the dropdown
- Click "Continue"
- Check the school details
- Click "Add organisation"
- Then the school should appear in the organisation index page.

## Link to Trello card

https://trello.com/c/oSjN8CBk/66-add-a-school

## Things to check

- [ ] If the code removes any existing feature flags, a data migration has also been added to delete the entry from the database
- [ ] This code does not rely on migrations in the same Pull Request
- [x] If this code includes a migration adding or changing columns, it also backfills existing records for consistency
- [ ] If this code adds a column to the DB, decide whether it needs to be in analytics yml file or analytics blocklist
- [ ] API release notes have been updated if necessary
- [ ] If it adds a significant user-facing change, is it documented in the [CHANGELOG](CHANGELOG.md)?
- [ ] Required environment variables have been updated or [added to the Azure KeyVault](/docs/environment-variables.md#deploy-pipeline)

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/17162488/3846f874-3b5d-4e8f-9479-f6f3d4584d9f

__Error Handling__
Screenshot 2024-01-04 at 16 25 43

Screenshot 2024-01-04 at 16 26 42

PR #80, opened 2024-01-05
Allow support user to add a school in claims

CatalinVoineag:
## Context

We need to allow a support user in claims to add a school. They will be able to search through all Gias schools and link a `Gias school` to a `school` record in the `claims` domain.

This uses the `school.js` autocomplete introduced here #73 

Refactored the SchoolSuggestions controller to be a shared controller https://github.com/DFE-Digital/itt-mentor-services/pull/80#discussion_r1442793084

New controller actions in claims schools controller
New school suggestions controller
New Views
New Routes

## Guidance to review

- Go to claims.localhost:3000/personas
- Sign in as Colin
- Click Add organisation and fill in the form
- Check if you're organisation is added to the school index page

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/80fa4462-10eb-4010-9bda-84f48152a874


CatalinVoineag:
[As a Support User, I want to be able to “Add a new School”](https://trello.com/c/tVIE1Dbu/51-as-a-support-user-i-want-to-be-able-to-add-a-new-school)

spec/system/placements/schools/support_user_adds_a_school_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-RSpec.describe "Placements / Schools / Support User adds a Provider",
+RSpec.describe "Placements / Schools / Support User adds a School",
```
JamieCleare2525, line 3: Nice catch!

app/controllers/claims/support/school_suggestions_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Claims::Support::SchoolSuggestionsController < ApplicationController
```
JamieCleare2525, line 1: Given how this is identical to the `Placements::Support::SchoolSuggestionsController`, is it worth having a shared `SchoolSuggestionsController`?

CatalinVoineag: Yeah, I think so

app/controllers/school_suggestions_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+  def index
+    render json: GiasSchool.search_name_urn_postcode(query_params)
+  end
+
+  private
+
+  def query_params
+    params.require(:query)&.downcase
+  end
+end
```
Nitemaeric, line 11: I'm wondering, does this deserve to go into an API controller instead? `Api::SchoolsController` maybe? That way we can do `GET /api/schools?query=abc`.

cc: @JamieCleare2525

app/controllers/claims/support/schools_controller.rb
```diff
@@ -6,4 +6,44 @@ def index
...
+
+  def new
+    @school = Claims::School.new
+  end
+
+  def check
+    if school.valid? && !school.claims?
+      @school = school.decorate
+    else
+      school.errors.add(:urn, :taken) if school.claims?
```
Nitemaeric, line 18: Should we use the model validations for this instead?

```ruby
class Claims::School < School
  validates :urn, uniqueness: { scope: [claims: true] }
end
```

config/locales/claims/en/layouts/application.yml
```diff
@@ -4,3 +4,21 @@ en:
...
+        check:
+          add_organisation: Add organisation
+          address: Address
+          cancel: Cancel
+          contact_details: Contact details
+          organisation_name: Organisation name
+          telephone: Telephone number
+          title: Check your answers
+          urn: Unique reference number (URN)
+          website: Website
```
Nitemaeric, line 24: I think this can do into a separate locale file, `config/locales/claims/en.yml` at minimum, `config/locales/claims/en/support/schools.yml` if we want to be explicit.

app/controllers/school_suggestions_controller.rb
CatalinVoineag: Sure, it is an api call after all

app/controllers/claims/support/schools_controller.rb
```diff
@@ -6,4 +6,44 @@ def index
...
+      render :new
+    end
+  end
+
+  def create
+    school.claims = true
```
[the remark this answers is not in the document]
CatalinVoineag: It doesn't seem to set it by default

app/controllers/school_suggestions_controller.rb
JamieCleare2525: That convention is a bit weird to me, as I have always worked from the point of view that API controllers deal with APIs we have written to be consumed by other applications. Not for the retrieval of information from other APIs. However, I'm happy if we feel this will work for us.

app/controllers/claims/support/schools_controller.rb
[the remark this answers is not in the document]
JamieCleare2525: This is because the `school` association of a `GiasSchool` will return `School`, not a namespaced model.
This is important, as using `Claims::School` will not recognise if the school has already been added by Placements, causing it to error. When in actual fact all we want to do in that case is trigger the flag.

app/controllers/claims/support/schools_controller.rb
JamieCleare2525: As mentioned below, this wouldn't necessarily work because we would be working with the `School` model. Just in case the record already exists as a `Placements::School`. 
You could use `becomes` https://apidock.com/rails/ActiveRecord/Base/becomes, but this seems long winded just to add an error.

app/controllers/school_suggestions_controller.rb
Nitemaeric: I see it as a separation of response expectations. `/api/** returns `json` responses and `!/app/**` returns `html`. It would be an internal API, for external APIs, I'd recommend versioning the API, but we don't have to go there.

app/controllers/claims/support/schools_controller.rb
Nitemaeric: Hmm, makes sense in terms of the constraint. I would avoid adding model errors into controller code though 🤔 Maybe we can extract this into a model method `#validate_urn` maybe?

app/controllers/claims/support/schools_controller.rb
```diff
@@ -6,4 +6,44 @@ def index
...
+    if school.save
```
Nitemaeric, line 25: In that case, can we compact these two lines?

```suggestion
    if school.update(claims: true)
```

app/controllers/claims/support/schools_controller.rb
```diff
@@ -6,4 +6,44 @@ def index
...
+    @school ||=
+      begin
+        gias_school = GiasSchool.find_by(urn: urn_param)
+        if gias_school.blank?
+          Claims::School.new
+        else
+          gias_school.school || gias_school.build_school
+        end
+      end
+  end
```
Nitemaeric, line 44: Out of curiosity, why do we _start_ with the `GiasSchool` model? Could we do something along the lines of

```suggestion
  def school
    @school ||= Claims::School.find_or_initialize_by(gias_school:)
  end
  
  def gias_school
    @gias_school ||= GiasSchool.find_by(urn: urn_param) 
  end
```

This would ensure we always have the namespaced `School` model, right?

app/controllers/claims/support/schools_controller.rb
```diff
@@ -6,4 +6,44 @@ def index
...
+
+  def urn_param
+    params.dig(:gias_school, :urn) || params.dig(:school, :urn)
+  end
```
Nitemaeric, line 48: This also raises the question, on the `SchoolsController`, should our request body ever be `gias_school`?

This standardises body params pattern, which _should_ help with consistency:

```suggestion
  def school_params
    params.require(:school).permit(:urn)
  end
  
  def urn_param
    school_params.require(:urn)
  end
```

Not strongly opinionated about the `school_params` part, but I don't think we need to support `params[:gias_school]`.

app/controllers/claims/support/schools_controller.rb
JamieCleare2525: We can separate out the `GiasSchool` out if you really want to. But again, the `Claim::School` won't find schools which already exist as `Placements::School`. Therefore this will error when trying to save this record, as the `:urn` will have already been taken and a school record already exists.

app/controllers/claims/support/schools_controller.rb
JamieCleare2525: The reason for the original implementation is to account for the two different forms.
- check (Finding the GiasSchool to be used to create/check against a school)
- create (Creating a new School record using the `:urn`)

app/controllers/claims/support/schools_controller.rb
Nitemaeric: Ah yes, we want to share the data. I overlooked that line of thought. Makes sense to me.

Nitemaeric: Feel free to ignore this suggestion :)

JamieCleare2525: Also, that implementation won't account for when the `GiasSchool` is `nil`

app/controllers/claims/support/schools_controller.rb
Nitemaeric: Got it. Makes sense :) I'm wondering now if `check` belongs in a `GiasSchoolsController`, but no action needed. Just a thought 🤔 .

All good! Thanks!

app/controllers/school_suggestions_controller.rb
```diff
@@ -0,0 +1,11 @@
+class SchoolSuggestionsController < ApplicationController
+  def index
+    render json: GiasSchool.search_name_urn_postcode(query_params)
+  end
+
+  private
+
+  def query_params
+    params.require(:query)&.downcase
+  end
```
Nitemaeric, line 10: With `require`, can this ever be missing? Also, since it's a single param.

```suggestion
  def query_param
    params.require(:query).downcase
  end
```

JamieCleare2525: unfortunately yes. If nothing is input, then `params.require(:query)` will just return `nil`. Which obviously breaks the `downcase`

Nitemaeric: 😢

Nitemaeric (approved):

Nitemaeric:
Nice one!

JamieCleare2525 (approved):
LGTM!

PR #85, opened 2024-01-07
As a Support User, I should be able to invite new Users to a School Organisation

gms-gs:
As a Support User, I should be able to invite new Users to a School Organisation

## Context

So we can allow a support user to invite a user to a school

Screenshot 2024-01-08 at 14 08 06

Screenshot 2024-01-08 at 14 08 16

Screenshot 2024-01-08 at 14 08 25

Screenshot 2024-01-08 at 14 08 34

## Guidance to review

- Sign in as support user
- Go to a school
- Go to users tab
- Fill in details and invite a user
- Check your emails

## Link to Trello card

https://trello.com/c/B0iJQyXa/53-as-a-support-user-i-should-be-able-to-invite-new-users-to-a-school-organisation

### DIFF
--- a/app/controllers/claims/support/users_controller.rb
+++ b/app/controllers/claims/support/users_controller.rb
@@ -5,9 +5,38 @@ class Claims::Support::UsersController < Claims::Support::ApplicationController
     @users = @school.users
   end
 
+  def new
+    @user = Claims::User.new
+  end
+
+  def check
+    if user.valid?
+      @user = user.decorate
+    else
+      render :new
+    end
+  end
+
+  def create
+    if UserInviteService.call(user, @school, sign_in_url)
+      redirect_to claims_support_school_users_path(@school)
+      flash[:success] = "User added"
+    else
+      render :new
+    end
+  end
+
   private
 
   def set_school
     @school = Claims::School.find(params[:school_id])
+  end
+
+  def user
+    @user ||= Claims::User.new(user_params)
+  end
+
+  def user_params
+    params.require(:claims_user).permit(:first_name, :last_name, :email)
   end
 end
--- a/app/mailers/notify_mailer.rb
+++ b/app/mailers/notify_mailer.rb
@@ -1,0 +1,10 @@
+class NotifyMailer < ApplicationMailer
+  # NotifyMailer.send_school_invite_email.deliver_later
+  def send_school_invite_email(user, school, sign_in_url)
+    body = t("claims.notify_message.invite_user_to_school", user_name: user.full_name, school_name: school.name, sign_in_url:)
+
+    mailer_options = { to: user.email, subject: "You have been invited to #{school.name}", body: }
+
+    notify_email(mailer_options)
+  end
+end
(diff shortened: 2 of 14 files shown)


app/controllers/claims/support/users_controller.rb
```diff
@@ -5,9 +5,48 @@ def index
...
+  def user
+    @user ||= Claims::User.new(user_params)
+  end
+
+  def attach_user_to_school
+    Membership.create!(user_id: user.id, organisation_type: "School", organisation_id: @school.id)
+  end
+
+  def send_invite_email_to_user
+    NotifyMailer.send_school_invite_email(user, @school, sign_in_url).deliver_now
```
CatalinVoineag, line 46:
```suggestion
    NotifyMailer.send_school_invite_email(user, @school, sign_in_url).deliver_later
```

This should probably be deliver_later, sorry for the bad example but the email should be in a background job to not break the request.

app/mailers/notify_mailer.rb
```diff
@@ -0,0 +1,10 @@
+class NotifyMailer < ApplicationMailer
+  # NotifyMailer.send_school_invite_email.deliver_now
+  def send_school_invite_email(user, school, sign_in_url)
+    body = "Dear #{user.full_name} \n\n You have been invited to join the claims service for #{school.name}.\n\n Sign in here #{sign_in_url}"
```
CatalinVoineag, line 4: Can we do this with translations?

app/controllers/claims/support/users_controller.rb
```diff
@@ -5,9 +5,48 @@ def index
...
+      render :new
+    end
+  end
+
+  def create
+    if user.save
+      attach_user_to_school
+      send_invite_email_to_user
+      redirect_to claims_support_school_users_path(@school)
+      flash[:success] = "User added"
```
CatalinVoineag, line 25: The logic to save a user, attach it to a school, and sending an email feels like it belongs to a service like `UserInviteService.call` I think it's making it more reusable and testable

CatalinVoineag (changes requested):
I only asked for a change request because I think the email should be a background job, the rest are more suggestions. Great work though. I have run this locally and it works as expected

app/services/user_invite_service.rb
```diff
@@ -0,0 +1,26 @@
...
+  def initialize(user, school, sign_in_url)
+    @user = user
+    @school = school
+    @sign_in_url = sign_in_url
+  end
+
+  attr_reader :user, :school, :sign_in_url
+
+  def call
+    attach_user_to_school
```
CatalinVoineag, line 13: Sorry to be so anal about this. I think we should do the creation of the user in here, with the `attach_user_to_school` within an [ActiveRecord Transaction block](https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html) so that in case any of the records fail to save we roll back. I just remembered the active transaction bit, sorry I did not include it in my first review

Something like 

```
def call
  if create user
    send_email
    true
  else
    false
  end
end

def create_user
  ActiveRecord::Base.transaction do
    user.create
    user.attach_to_school
  end
end
```

CatalinVoineag (approved):
Very nice! Thank you for the changes 🚀

PR #87, opened 2024-01-08
Redirect users to organisations index/show after sign in

CatalinVoineag:
## Context

In claims we need to redirect our users to the organisations index page. Some users do have only one organisation(school) so for those users we will redirect them to the school show page, we do this via another redirect in the schools controller just to make sure a user with one school doesn't land on the organisation(school) index page.

This change also affects placements where we will always redirect the user after sign in to the organisations index page, allowing placements to add their logic in their controllers

Also I didn't want to add too much logic in the application controller.

Application controller
Scholl Controller

## Guidance to review
Claims:
- Sign in as a user with one organisation
- You should be redirected to the org show page
- Sign in as a multiple school user like Mary
- Should be redirected to org index page

Placements:
- Sign in as any user
- You should be redirected to org index page

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/469c53af-1f29-488f-82d0-d9a0cb2b3ab9

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,10 +25,8 @@ class ApplicationController < ActionController::Base
   def after_sign_in_path
     if current_user.support_user?
       support_root_path
-    elsif current_user.memberships.many?
-      organisation_index_path
     else
-      root_path
+      organisations_path
     end
   end
 
--- a/app/controllers/claims/schools_controller.rb
+++ b/app/controllers/claims/schools_controller.rb
@@ -1,5 +1,15 @@
 class Claims::SchoolsController < ApplicationController
+  before_action :redirect_to_school_when_belongs_to_one_school, only: :index
+
   def index
     @schools = current_user.schools
+  end
+
+  private
+
+  def redirect_to_school_when_belongs_to_one_school
+    if current_user.schools.one?
+      redirect_to claims_school_path(current_user.schools.first)
+    end
   end
 end
--- a/app/helpers/routes_helper.rb
+++ b/app/helpers/routes_helper.rb
@@ -13,7 +13,7 @@ module RoutesHelper
     }.fetch current_service
   end
 
-  def organisation_index_path
+  def organisations_path
     {
       claims: claims_schools_path,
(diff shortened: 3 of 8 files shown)


CatalinVoineag:
[As a User, if I belong to a single School Organisation, I should be redirected to that School when visiting the list of Schools I belong to](https://trello.com/c/hZAyLRHH/59-as-a-user-if-i-belong-to-a-single-school-organisation-i-should-be-redirected-to-that-school-when-visiting-the-list-of-schools-i)

CatalinVoineag:
[As a User, I should be able to see a list of my School Organisations](https://trello.com/c/fd8tYy8s/60-as-a-user-i-should-be-able-to-see-a-list-of-my-school-organisations)

app/controllers/application_controller.rb
```diff
@@ -25,10 +25,8 @@ def current_user
   def after_sign_in_path
     if current_user.support_user?
       support_root_path
-    elsif current_user.memberships.many?
-      organisation_index_path
     else
-      root_path
+      organisation_index_path
```
Nitemaeric, line 29: I just saw this in the helper now too. Since you're touching it, can we rename this to `organisations_path` to match Rails' conventions please?

app/controllers/claims/schools_controller.rb
```diff
@@ -1,5 +1,15 @@
 class Claims::SchoolsController < ApplicationController
+  before_action :has_one_school?, only: :index
```
Nitemaeric, line 2: Maybe a nitpick, but can we rename this to show what it does?

```suggestion
  before_action :redirect_to_school_when_belongs_to_one_school, only: :index
```

Nitemaeric (commented):
Looking good!

app/controllers/application_controller.rb
CatalinVoineag: We can but I thought `organisations_path` is like a reserved name but I guess we have placements and claims org path so for now we should be fine

Nitemaeric (approved):
🚀

PR #91, opened 2024-01-08
Add register_components helper for registering View Component helper methods

Nitemaeric:
## Context

We want to be able to rapidly create and use View Components. By following the precedent set with the GOV UK components gem, we can easily add in view helpers to render our view components.

CleanShot 2024-01-08 at 16 23 54

- Add a `register_components` method to the `ComponentsHelper`.

## Guidance to review

- See the example at [app/components/claims/greeting_component.rb](https://github.com/DFE-Digital/itt-mentor-services/blob/c57a90ad59124eebfc070a23c13f3a8e9b904ee7/app/components/claims/greeting_component.rb).
- Try adding a new custom View Component, registering it via `register_components`, and finally rendering it on a page.

## Things to check


app/helpers/components_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+      namespaced_component_name = [namespace, component_name].compact.join("_")
+      component_class_name = "#{[namespace, component_name].compact.join("/")}_component"
+
+      define_method(namespaced_component_name) do |*args, **kwargs, &block|
+        render component_class_name.classify.constantize.new(*args, **kwargs) do |component|
+          block.call(component) if block.present?
+        end
+      end
+    end
+  end
```
ollietreend, line 13: I can see the appeal of this from an aesthetic point of view. However, I do wonder if it's trying to solve a problem that doesn't really exist?

Ideally we'd try to [avoid unnecessary metaprogramming](https://rubystyle.guide/#no-needless-metaprogramming), and I'm wondering if we could survive without this particular piece of metaprogramming.

This adds a layer of indirection that will presumably break the `Cmd + Click` _jump to definition_ feature in developer IDEs. And while it suggests a convention, it doesn't enforce it, so we end up with two different ways to render the same component.

I can certainly see the appeal of implementing a more 'Rails helper method'-like public API for the [govuk-components](https://govuk-components.netlify.app/) gem. But I'm wondering if it's really necessary within this project, where we plan on using ViewComponents internally within the app.

app/helpers/components_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  register_components :primary_navigation, :secondary_navigation
+  register_components :greeting, namespace: :claims
```
ollietreend, line 16: Developer: "I made a new component but I couldn't use it in my view. Turns out I forgot to add a line to `app/helpers/components_helper.rb`" 😢

app/helpers/components_helper.rb
elceebee: I agree with Ollie here, selfishly perhaps, mostly because it breaks the `Cmd + Click` in my IDE. So I have to try to work out what the component is called based on the helper name and then find the definition separately. It's not such an easy developer experience.

Nitemaeric:
@ollietreend @elceebee @CatalinVoineag Alrighty, based on the team feedback here, I'll drop the view helpers for the view components. 😄 😢 😄 😆

PR #99, opened 2024-01-09
Allow user back to school index from show

CatalinVoineag:
If a user has multiple organisations we need to allow him to go back from the organisations show page to the organisation index page.

This is done via a content header view component.

## Things to check


CatalinVoineag:
[As a User, I should see a link to navigate to my list of Schools](https://trello.com/c/vqc6ZPFq/61-as-a-user-i-should-see-a-link-to-navigate-to-my-list-of-schools)

config/locales/en.yml
```diff
@@ -91,7 +91,8 @@ en:
       invite_user_to_school: "Dear %{user_name} \n\n You have been invited to join the claims service for %{school_name}.\n\n Sign in here %{sign_in_url}"
     schools:
       show:
-        heading: Organisations
+        change_organisation: Change organisation
```
elceebee, line 94: We had the "change organisation" button removed from the placements version of this page. Simon said it wasn't necessary anymore because of the nav bars. I wonder if maybe he didn't get a chance to update your prototype yet? I assume it would be the same for both services.

elceebee (commented):
This looks good to me, just a question about the "Change organisation" button -- Simon said we didn't need it for the Placements service, so kind of assume it will be the same for Claims. Worth checking with Simon.

config/locales/en.yml
CatalinVoineag: So I spoke with Simon, you are partially right. We don't need it for support users but for normal users that have multiple schools we do need it. So this PR is still valid :)

elceebee (approved):

PR #101, opened 2024-01-09
Paginate provider importer

JamieCleare2525:
## Context

The Publish Teacher Training API is paginated and only returns 100 records per page (Unless specifically set, with a maximum of 500 https://api.publish-teacher-training-courses.service.gov.uk/docs/api-reference.html#recruitment_cycles-year-providers-get-responses-examples). 
Therefore we need to paginate in case the number of records exceeds 100.

Recursive call in `AccreditedProvider::Importer`

## Guidance to review

- run `bundle exec rake db:drop`
- run `bundle exec rake db:setup`
- ⌛ Wait for Setup to run.
- run `rails console`
- run `Provider.count`
- 💻  Provider count should be **184**

## Link to Trello card

https://trello.com/c/yA3tm7AA/97-add-support-for-pagination-when-calling-the-teacher-training-courses-api


elceebee (approved):
Pulled locally and this works as expected. Nice work

CatalinVoineag (approved):

PR #102, opened 2024-01-09
Remove school name from search box and add to error

elceebee:
## Context

To recreate bug: 
Login as a support user (either claims or placements) and Click Add an organisation
Search for a school that has already been added.
Click “Continue”
Click into the search box
Then you see `undefined (undefined undefined)` in the drop down

Remove the default text from the search box as it is not useable (the search isn't run by default when the page reloads), and 
Add the school name to the error so we don't lose the context about which schools has already been added. 

## ACs

AS a support user
GIVEN I am logged in and on the add a school page
WHEN I Add a school that has already onboarded
AND I click into the search box again
THEN I should see the selected school in the error message
AND the search box should be empty
AND I should not see any search results when I click there (ie, I should not see not `undefined (undefined undefined)`)

## Limitations / Scope

There is another ticket to deal with the erroneous "can't be blank" error that shows up when you click continue without selecting a school. 
There is also a ticket to refactor the logic for this, and providers, so that we use a Form object instead of managing the errors in the controller.

## Guidance to review

Follow the steps above to recreate. But instead, you should see an empty search box and a descriptive error. When you click on the box, you won't see 'undefined (undefined undefined)

## Link to Trello card

https://trello.com/c/CD7HrkDz

## Things to check

- [x] If the code removes any existing feature flags, a data migration has also been added to delete the entry from the database

### Before

https://github.com/DFE-Digital/itt-mentor-services/assets/44073106/ee0a75ab-db45-4e93-95e6-7c5613681520

### After

https://github.com/DFE-Digital/itt-mentor-services/assets/44073106/985c5b4f-72f9-4ac3-ae5a-023c997172a0

### DIFF
--- a/app/controllers/claims/support/schools_controller.rb
+++ b/app/controllers/claims/support/schools_controller.rb
@@ -15,7 +15,7 @@ class Claims::Support::SchoolsController < Claims::Support::ApplicationControlle
     if school.valid? && !school.claims?
       @school = school.decorate
     else
-      school.errors.add(:urn, :taken) if school.claims?
+      school.errors.add(:urn, :already_added, school_name: school.name) if school.claims?
       render :new
     end
   end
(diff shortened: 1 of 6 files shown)


JamieCleare2525 (approved):
LGTM!

CatalinVoineag (approved):
Thank you for fixing this, pulled it on my local works as expected

PR #105, opened 2024-01-10
Rework onboarding with form objects for schools and providers

JamieCleare2525:
## Context

Currently we are seeing a number of issues with forcing ourselves to use a CRU/D for onboarding Providers and Schools onto the different services.
Replacing the models used in the different controllers with form objects standardises the approach to onboarding external records, and reduces unwanted validating errors appearing when completing the onboarding form.

- Introduces form objects (ProviderForm, SchoolForm) for onboarding (flipping the relevant service flag to `true`)
 of Provider and School records.

## Guidance to review

[ ] Test onboarding a provider on the `Placements` service.
[ ] Test onboarding a school on the `Placements` service.
[ ] Test onboarding a school on the `Claims` service.

## Link to Trello card

https://trello.com/c/xg3Ylg0f/104-refactor-school-provider-onboarding-journeys-to-use-form-objects-pattern
https://trello.com/c/ornOs5qW/103-bug-in-school-search-error-messages-placements-and-claims

### DIFF
--- a/app/controllers/claims/support/schools_controller.rb
+++ b/app/controllers/claims/support/schools_controller.rb
@@ -8,20 +8,19 @@ class Claims::Support::SchoolsController < Claims::Support::ApplicationControlle
   end
 
   def new
-    @school = Claims::School.new
+    @school_form = SchoolOnboardingForm.new
   end
 
   def check
-    if school.valid? && !school.claims?
-      @school = school.decorate
+    if school_form.valid?
+      @school = school
     else
-      school.errors.add(:urn, :already_added, school_name: school.name) if school.claims?
       render :new
     end
   end
 
   def create
-    if school.update(claims: true)
+    if school_form.onboard
       redirect_to claims_support_schools_path
     else
       render :new
@@ -30,11 +29,15 @@ class Claims::Support::SchoolsController < Claims::Support::ApplicationControlle
 
   private
 
+  def school_form
+    @school_form ||= SchoolOnboardingForm.new(urn: urn_param, service: :claims)
+  end
+
   def school
-    @school ||= School.find_by(urn: urn_param) || Claims::School.new
+    @school ||= school_form.school.decorate
   end
 
   def urn_param
-    params.dig(:selection, :urn) || params.dig(:school, :urn)
+    params.dig(:school, :search_urn)
   end
 end
--- a/app/controllers/placements/support/providers_controller.rb
+++ b/app/controllers/placements/support/providers_controller.rb
@@ -1,31 +1,32 @@
 class Placements::Support::ProvidersController < Placements::Support::ApplicationController
   def new
-    @provider = Provider.new
+    @provider_form = ProviderOnboardingForm.new
+  end
+
+  def check
+    if provider_form.valid?
+      provider
+    else
+      render :new
+    end
   end
 
   def create
-    if provider.update(placements: true)
+    if provider_form.onboard
       redirect_to placements_support_organisations_path
     else
       render :new
     end
   end
 
-  def check
-    @provider = provider.decorate
-    return if @provider.valid?
-
-    render :new
-  end
-
   private
 
-  def provider
-    @provider ||= Provider.find_by(code: provider_code, placements: false) || unknown_provider
+  def provider_form
+    @provider_form ||= ProviderOnboardingForm.new(code: provider_code)
   end
 
-  def unknown_provider
-    Provider.new(name: "Unknown Provider", provider_type: "university", code: provider_code)
+  def provider
+    @provider ||= provider_form.provider.decorate
   end
 
   def provider_code
--- a/app/views/claims/support/schools/check.html.erb
+++ b/app/views/claims/support/schools/check.html.erb
@@ -5,7 +5,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= form_with(model: @school, url: claims_support_schools_path, method: "post", data: { turbo: false }) do |f| %>
-      <%= f.hidden_field :urn, value: @school.urn %>
+      <%= f.hidden_field :search_urn, value: @school.urn %>
 
       <label class="govuk-label govuk-label--l">
         <span class="govuk-caption-l"><%= t(".add_organisation") %></span>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -7,6 +7,19 @@ en:
       long: "%e %B %Y %H:%M"
       long_with_seconds: "%e %B %Y %H:%M:%S"
       time: "%H:%M"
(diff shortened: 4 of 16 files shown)


CatalinVoineag (approved):
Tested this on local, works as expected, very good work!

app/forms/school_form.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def school
+    @school ||= School.find_by(urn:)
+  end
+
+  private
+
+  def school_already_onboarded?
+    if school.blank?
+      errors.add(:urn, :blank)
```
ollietreend, line 24: I also wonder if `:blank` is the correct message to display here? 🤔

Although it seems to be an edge case, because it should (in theory) only happen if the user fiddles with the URL parameter themselves. So perhaps it's fine to leave this as is.

app/forms/provider_form.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  private
+
+  def provider_already_onboarded?
+    if provider.blank?
+      errors.add(:code, :blank)
+    elsif provider.placements?
+      errors.add(:code, :already_added, provider_name: provider.name)
+    end
+  end
```
ollietreend, line 27: Ditto

app/controllers/claims/support/schools_controller.rb
```diff
@@ -8,20 +8,19 @@ def show
   end
 
   def new
-    @school = Claims::School.new
+    @school_form = SchoolForm.new
```
ollietreend, line 11: I'm wondering if "school onboarding form" is a more accurate name for this. That way we leave room for other potential forms that might be needed in relation to schools – such as user input of additional fields that we can't source from GIAS.

```suggestion
    @school_onboarding_form = SchoolOnboardingForm.new
```

What do you think?

app/controllers/placements/support/providers_controller.rb
```diff
@@ -1,31 +1,32 @@
 class Placements::Support::ProvidersController < Placements::Support::ApplicationController
   def new
-    @provider = Provider.new
+    @provider_form = ProviderForm.new
```
ollietreend, line 3: Same same

```suggestion
    @provider_onboarding_form = ProviderOnboardingForm.new
```

app/views/claims/support/schools/check.html.erb
```diff
@@ -5,7 +5,7 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= form_with(model: @school, url: claims_support_schools_path, method: "post", data: { turbo: false }) do |f| %>
-      <%= f.hidden_field :urn, value: @school.urn %>
+      <%= f.hidden_field :search_urn, value: @school.urn %>
```
ollietreend, line 8: Any reason why you renamed this field/param from `urn` to `search_urn`?

I have no strong opinion either way – just thought `urn` was probably a more accurate name. By the time they get to the 'check' stage, they're not 'searching' for a school any more.

I also see you've stuck with `code` on the provider form, rather than renaming that field to `search_code`.

config/locales/en.yml
```diff
@@ -7,23 +7,26 @@ en:
...
-        organisation:
-          attributes:
-            type:
-              blank: Select an organisation type
-        provider:
+        provider_form:
           attributes:
             code:
-              taken: This provider has already been added. Try another provider
+              already_added: "%{provider_name} has already been added. Try another provider"
```
ollietreend, line 16: Nice use of variables in validation messages! ✨

app/forms/school_form.rb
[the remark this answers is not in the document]
ollietreend: In fact, I can see this isn't a journey we explicitly test anyway. So maybe this is fine – it's better than no error at all, and it's a bit of an edge-case journey.

ollietreend (commented):
Really nice work @JamieCleare2525! 🥳

I've added a few comments with some things to think about.

Also, this has got me thinking. I bet the "Add organisation" journey (i.e. the intermediary "Which type of organisation do you want to add?") form is another good candidate for a Form Object. Perhaps we create another card to implement it there too?

app/views/claims/support/schools/check.html.erb
JamieCleare2525: This is because in the `new` form page, the `:urn` is used to for the search (showing the name of the school) and the JS adding a `search_urn` field.
Originally the JS added a selection field which led to the following code in the controller:
```
def urn_param
  params.dig(:selection, :urn) || params.dig(:school, :urn)
end
```
Changing this to `search_urn` means we only need to use the following code in the controller:
```
def urn_param
  params.dig(:school, :search_urn)
end
```
Minimising code and confusion (hopefully).

app/forms/school_form.rb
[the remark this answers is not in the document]
JamieCleare2525: I'm going to leave this check in. Just incase "as you say" 'the user fiddles with the URL'.

JamieCleare2525: See above.

app/forms/school_form.rb
```diff
@@ -0,0 +1,29 @@
...
+    elsif school.try(service.to_s)
+      errors.add(:urn, :already_added, school_name: school.name)
+    end
+  end
```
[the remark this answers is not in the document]
JamieCleare2525: resolved in [8a29568](https://github.com/DFE-Digital/itt-mentor-services/pull/105/commits/8a2956808a0e940e50960406e039c4df9a898787)

app/forms/provider_form.rb
JamieCleare2525: resolved in [8a29568](https://github.com/DFE-Digital/itt-mentor-services/pull/105/commits/8a2956808a0e940e50960406e039c4df9a898787)

app/controllers/claims/support/schools_controller.rb
JamieCleare2525: resolved on [8a29568](https://github.com/DFE-Digital/itt-mentor-services/pull/105/commits/8a2956808a0e940e50960406e039c4df9a898787)

app/controllers/placements/support/providers_controller.rb
JamieCleare2525: resolved in [8a29568](https://github.com/DFE-Digital/itt-mentor-services/pull/105/commits/8a2956808a0e940e50960406e039c4df9a898787)

elceebee (commented):
I haven't left any comments because they would mostly be echoing some of Ollie's comments. But just wanted to say that I think this is great. I'm glad we are starting to establish this pattern in the code base.

ollietreend (approved):
Lovely work 😍 looks great!

PR #109, opened 2024-01-12
As a Support User, I should be able to see the details of a School

gms-gs:
## Context

Update the support schools page to show the schools details

Add read only fields to show a schools details

## Guidance to review

- login as support user 
- Go to the support/schools url
- Select a school

## Link to Trello card

https://trello.com/c/4iG5S0cP/68-as-a-support-user-i-should-be-able-to-see-the-details-of-a-school-organisation

## Things to check

## Screenshots

Screenshot 2024-01-12 at 08 57 18

### DIFF
--- a/app/controllers/claims/support/schools_controller.rb
+++ b/app/controllers/claims/support/schools_controller.rb
@@ -5,7 +5,7 @@ class Claims::Support::SchoolsController < Claims::Support::ApplicationControlle
   end
 
   def show
-    @school = Claims::School.find(params.require(:id))
+    @school = Claims::School.find(params.require(:id)).decorate
   end
 
   def new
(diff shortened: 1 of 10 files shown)


app/controllers/claims/support/schools_controller.rb
```diff
@@ -5,7 +5,7 @@ def index
   end
 
   def show
-    @school = Claims::School.find(params.require(:id))
+    @school = Claims::School.find(params.require(:id))&.decorate
```
JamieCleare2525, line 8: You shouldn't need the `&`, as `find` will error if it can't find the school.

CatalinVoineag (approved):
This looks good to me, I think my only comment would be to maybe consider just reusing the partials from here as they are similar, I think. But it's not necessary, this works perfectly fine too 

https://github.com/DFE-Digital/itt-mentor-services/tree/main/app/views/claims/schools

JamieCleare2525 (approved):
Tested locally. Looks good to me.

app/controllers/claims/support/schools_controller.rb
JamieCleare2525: If that's ok. @gms-gs

PR #110, opened 2024-01-12
Jc/remove accredited filter from provider api

JamieCleare2525:
## Context

After discussions with Grace, Simon and James, we need to remove the accredited filter from the Provider API to allow for allow Providers in the current recruitment cycle to use the service.

- Remove accredited filter `filter%5Bis_accredited_body%5D=true` from the API calls to Publish Teacher Training
- Add `accredited` boolean to `Provider`, to indicate which providers are accredited
- Introduce `Timecop` into `Rspec` for time sensitive tests
- Calculate the current recruitment cycle, when calling the Publish Teacher Training API

## Guidance to review

- run `bundle exec rails db:drop`
- run `bundle exec rails db:create`
- run `bundle exec rails db:setup`
- ⌛  Wait for setup to process ⌛ 
- run `rails console` in terminal
- run `Provider.count` (should return **1033**)
- run `Provider.accredited.count` (should return **183**)

## Link to Trello card

https://trello.com/c/EjWIbUXo/114-remove-accredited-filter-from-provider-api

## Things to check


app/services/provider/importer.rb
```diff
@@ -26,7 +26,7 @@ def invalid?(provider_attributes)
   end
 
   def fetch_providers(link = nil)
-    accredited_providers = ::AccreditedProvider::Api.call(link:)
+    accredited_providers = ::Provider::Api.call(link:)
```
elceebee, line 29: To be consistent, we should rename this to providers instead of accredited_providers. (line 30 as well)
```suggestion
    providers = ::Provider::Api.call(link:)
```

spec/models/provider_spec.rb
```diff
@@ -46,4 +47,24 @@
     end
     it { is_expected.to allow_values("scitt", "lead_school", "university").for(:provider_type) }
   end
+
+  context "scopes" do
+    describe "#placements" do
+      let!(:onboarded_provider) { create(:provider, placements: true) }
```
elceebee, line 53: Not important, but you can do `create(:placements_provider)`

elceebee (approved):
Ran locally and everything works according to plan, nice one! just a couple of small comments. Would be good to change the var name from `accredited_providers` to `providers`, so it's not confusing to look at later. But I'll go ahead and approve as I don't really need to see it again.

spec/models/provider_spec.rb
JamieCleare2525: I should have remembered that. I wrote the factory. 😛

JamieCleare2525: Oh wait. I did that because `create(:placements_provider)` returns a `Placements::Provider` but the scope returns `Provider`. So it needs to be `create(:provider, placements: true)`

app/services/provider/importer.rb
JamieCleare2525: resolved in [c8aab49](https://github.com/DFE-Digital/itt-mentor-services/pull/110/commits/c8aab4935ccdfccc6124952ee65eafbea8b40219)

PR #112, opened 2024-01-12
Reorganise Claims controllers, views, and locales

Nitemaeric:
## Context

We want to be able to navigate throughout the app without broken links.

- Reorganise controllers, views, and locale files.
- Update locale files to 1 file per controller and use same nesting structure as controllers and views.
- Add blank pages with page headings for anticipated routes.

## Guidance to review

- There shouldn't be any broken links in the Claims service.

## Link to Trello card

https://trello.com/c/lqqVd0Sq/116-add-in-some-placeholder-pages

## Things to check


Nitemaeric:
[Add in some placeholder pages](https://trello.com/c/lqqVd0Sq/116-add-in-some-placeholder-pages)

config/routes/claims.rb
```diff
@@ -1,17 +1,29 @@
 scope module: :claims, as: :claims, constraints: { host: ENV["CLAIMS_HOST"] } do
-  root to: "pages#index"
+  root to: redirect("/schools")
```
CatalinVoineag, line 2: Is there a reason why we don't use rails routes helper here?

CatalinVoineag (commented):
Pulled this locally, works as expected. I like the translations being separate, I used to hate adding translations to a file with 20k lines 😄

CatalinVoineag (approved):

config/routes/claims.rb
Nitemaeric: I don't think route helpers are available within the route files, as the route files are used to generate the route helpers.

CatalinVoineag: Duh sorry.

app/controllers/claims/support/claims_controller.rb
```diff
@@ -1,11 +1,2 @@
 class Claims::Support::ClaimsController < Claims::Support::ApplicationController
-  before_action :set_school, only: [:index]
```
JamieCleare2525, line 2: If there's nothing in this controller is it necessary?

app/controllers/claims/support/schools/users_controller.rb
```diff
@@ -0,0 +1,46 @@
+class Claims::Support::Schools::UsersController < Claims::Support::ApplicationController
+  before_action :set_school
+
+  def index
+    @users = @school.users
```
JamieCleare2525, line 5: I noticed this when thinking about replicating this for placements.
This should be:
```suggestion
    @users = @school.users.claims
```

app/controllers/claims/support/claims_controller.rb
Nitemaeric: The controller is used to map to the default Rails `index` action. It only renders the view template which has no variables, hence no setup in the controller.

CleanShot 2024-01-15 at 09 22 51

app/controllers/claims/support/schools/users_controller.rb
```diff
@@ -0,0 +1,46 @@
+class Claims::Support::Schools::UsersController < Claims::Support::ApplicationController
```
ollietreend, line 1: Does there need to be a 'support'-specific controller for this?

Non-support users within a school can also manage the user list (e.g. they can invite new users, remove users, etc. from their school). So every type of user will be able to manage 'school users'. Won't this lead to duplication if we create one specifically for 'support' users?

e.g. I'm wondering if this should actually be `Claims::Schools::UsersController`

app/controllers/claims/support/schools/providers_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Claims::Support::Schools::ProvidersController < Claims::Support::ApplicationController
```
ollietreend, line 1: Does the Claims service need a Providers controller? 🤔

I was of the understanding that Providers weren't a first-class entity in Track & Pay?

app/controllers/claims/support/schools/users_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      redirect_to claims_support_school_users_path(@school)
+      flash[:success] = "User added"
+    else
+      render :new
+    end
+  end
+
+  private
+
+  def set_school
```
JamieCleare2525, line 35: This set school method is replicated in every "School" namespaced controller.
Is it with separating this out into a concern and including it?
That way should you need to change it (for what ever reason), you don't need to change it X times.

Nitemaeric: Hmm, I thought about it then decided against due to the clarity of containing everything within the individual controllers themselves.

But with the appropriate name, maybe it's clear enough. What do you think of below?

```ruby
class Claims::Support::Schools::UsersController
  include BelongsToSchool
end
```

ollietreend (commented):
I would feel slightly uneasy about assuming which controllers and routes will be needed upfront – it feels a little risky to try and pre-empt these ahead of time.

So I'm not sure that this is a pattern we'll choose to follow on the School Placements side. My thinking is that it'll become clearer what's needed as our team works to build up a backlog of work.

However: this is all based in opinion and preference. So it's not a blocker if you do choose to take this route.

app/controllers/claims/support/schools/users_controller.rb
Nitemaeric: My experience has typically led me to separate concerns between user-facing actions and admin-facing actions.

Yes, both can add users, but the side-effects and constraints between the two may differ, especially at the controller level, for example, we may want to send out notifications to admins when a user performs an action or we might want to notify users when an admin performs an invite on their behalf for example. To DRY up the code in terms of functionality, we can always call a shared service object.

app/controllers/claims/support/schools/providers_controller.rb
Nitemaeric: I'm referencing the current prototype here: https://itt-mentoring-beta-94341ca35a2c.herokuapp.com/support/organisations/049fc44e-6876-4884-8a77-d363173ce8d2#

If needed, we can drop it when it's removed from the prototype.

elceebee (commented):
I like having the locales in different files, but I also think we should have some top level translations that will keep us from introducing errors. Eg, Organisation types. It would be very easy for someone to accidentally sneak in an Organisation Types (capitalized) and then we have inconsistent usage. 

For things like School, Provider, SCITT, University, Lead partner, Organisation, Organisation type, etc, as well as actions that are going to appear in many different views, like Back, Cancel, Add, Confirm, we should always look to the same translation. 

I'm not a huge fan of placeholder controllers / views. it's good to establish a pattern, but it could also just end up dead, unused code if the direction of travel doesn't go exactly as planned. Happy to go with it if others are. Not a big deal, but not entirely convinced of the value.

app/controllers/claims/support/schools/users_controller.rb
JamieCleare2525: I think that would make a lot of sense.

app/controllers/claims/support/schools/providers_controller.rb
elceebee: I think this is the kind of thing that concerns me -- will we remember to drop it? Or will we just create the necessary thing at the time it is needed and this will just live forever, a red herring for someone trying to debug or understand our codebase in the future?  But again, don't feel strongly about it.

Nitemaeric:
> For things like School, Provider, SCITT, University, Lead partner, Organisation, Organisation type, etc, as well as actions that are going to appear in many different views, like Back, Cancel, Add, Confirm, we should always look to the same translation.

I agree in terms of attribute names or types and etc. I think we can rely heavily on ActiveModel attribute name translations for things that need to be consistent, like `User.human_attribute_name(:first_name)`.

As for "Back", "Cancel", "Add" and others, I'd counter with a general best practice. It may not be directly relevant in this specific product, as we are likely to only support British English, but in some languages and certain parts of an app, "Back" may not always purely be a single word "Back". In some cultures, you may want to put "Back to Home" vs "Back to . If it is a consistent display, it should also be a part of a shared partial/template. This means we can still refer to a single relative localisation key. See `en.layouts.application.primary_navigation`.

> I'm not a huge fan of placeholder controllers / views. it's good to establish a pattern, but it could also just end up dead, unused code if the direction of travel doesn't go exactly as planned. Happy to go with it if others are. Not a big deal, but not entirely convinced of the value.

I'm typically not a fan of temporary content either really. The exception here being that I'd like to add these in as it sets up the pattern clearly for everyone, as we have quite a few pages planned out and we are quite early in development. (Users will never see these empty/placeholder pages).

app/controllers/claims/support/schools/users_controller.rb
Nitemaeric: Added :)

app/controllers/claims/support/schools/users_controller.rb
```diff
@@ -0,0 +1,42 @@
+class Claims::Support::Schools::UsersController < Claims::Support::ApplicationController
+  include Claims::BelongsToSchool
+
+  def index
+    @users = @school.users.claims
+  end
+
+  def show
+    @user = Claims::User.find(params.require(:id))
```
JamieCleare2525, line 9: While we're here. For bit of security, could we do:
```suggestion
    @user = @school.users.find(params.require(:id)).becomes(Placements::User)
```
Then the user will only see Users related to that organisation.

Nitemaeric: Addressed in https://github.com/DFE-Digital/itt-mentor-services/pull/112/commits/38e2682f0db3e76466b875a129b796dd55411c10

PR #113, opened 2024-01-14
Lb/placements support user organisations index page improvements

elceebee:
## Context

As a placements support user
I should be able to view all organisations (schools and providers) in a single list, ordered by name.
And I should be able to search and filter onboarded organisations

- svg files
- pg multisearch for providers and schools
- css (taken from the prototype) for the filtering section
- View for index for styling and filtering / searching
- Updated controllers and added service object to get organisations

## Guidance to review
You will need to run migrations (bin/setup), and then 
EITHER delete all Providers / Schools and reseed the db
OR rebuild the PG Search docs ( you only have to do this if you already have data in the db) 
from the rails console do: `PgSearch::Multisearch.rebuild(Provider)` and `PgSearch::Multisearch.rebuild(School)`

Then you can login as Colin the placements support user, add some schools / providers and start searching and filtering.

## Link to Trello card

https://trello.com/c/AM9ysC2B

## Things to check

- [x] This code does not rely on migrations in the same Pull Request (it does, but it doesn't matter)

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/44073106/41ebeaa3-28e5-4e53-98e6-6e8e97f670ed


app/views/shared/_search_form.html.erb
```diff
@@ -1,3 +1,5 @@
+<%# locals: (url:, search_param:, clear_search_url: "", filters: []) -%>
```
CatalinVoineag, line 1: Is this meant to be here?

app/services/placements/organisation_finder.rb
```diff
@@ -0,0 +1,24 @@
...
+      organisations.where(organisation_type: filters)
+    end
+  end
+
+  private
+
+  attr_reader :filters, :search_term
+
+  def organisations
+    @organisations ||= search_term.blank? ? PgSearch::Document.all.order(:content) : PgSearch.multisearch(search_term)
```
CatalinVoineag, line 22: I'm not very familiar with PG documents and this works for now but will this return documents from another model if we decide to use it for users for example? What I'm trying to ask here is should we get all documents by `organisation_type` for example? Or is the pg document table only for searching organisations and if we want to implement this somewhere else we create a different table?

CatalinVoineag (commented):
Tested this on local. Works as expected, very good work! Just 2 questions

app/helpers/placements/support/organisations_helper.rb
```diff
@@ -0,0 +1,17 @@
...
+      I18n.t("no_results_with_filter")
+    elsif filters.blank?
+      I18n.t("no_results", for: search)
+    else
+      I18n.t("no_results_with_filter_and_search", for: search)
+    end
+  end
+
+  def filter_options
+    %w[scitt school university lead_school].sort
```
JamieCleare2525, line 15: Just in case anything changes with the provider types is it worth changing this to:
```suggestion
    (Provider.provider_types.keys << "school").sort
```
Then you can be sure you'll always show the options for all provider types.

app/views/shared/_search_form.html.erb
elceebee: https://blog.kiprosh.com/allow-template-to-set-strict-locals/

So you can see what locals are allowed in a partial and assign defaults.

app/services/placements/organisation_finder.rb
elceebee: Good point

CatalinVoineag (approved):
Thank you!

JamieCleare2525 (approved):
Tested locally. LGTM!

PR #115, opened 2024-01-15
Support user can add a school with no javascript

CatalinVoineag:
## Context

We need to allow a support user to add a school even if he doesn't have
javascript enabled.

This commit adds two new controller methods to the schools controller and a
radio form component to allow the user still search a school by urn or
name but he will be redirected to a school options page where he will
need to choose a school from a radio button form

School options controller and school_options form component

## Guidance to review

Turn off javascript in your browser
Sign in as Colin
Add a School by searching for a school name
You should be redirected to a school options radio form, and submit the form.

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/0fed34b8-61cf-4e30-ab22-d10fcd1d37e6


CatalinVoineag:
[Add a non javascript path to adding organisations for a support user](https://trello.com/c/n9uJXy6b/118-add-a-non-javascript-path-to-adding-organisations-for-a-support-user)

config/locales/en.yml
```diff
@@ -11,6 +11,15 @@ en:
       scitt: SCITT
       university: University
       lead_school: Lead partner
+    school_options_form_component:
+      add_organisation: Add Organisation
```
elceebee, line 15: Everywhere else, we are using the less formal 'Add organisation'. 
```suggestion
      add_organisation: Add organisation
```

Nitemaeric (approved):
If I understand correctly, the split of `SchoolOptions` and `Schools` is to distinctly separate the concerns, right?

The alternative would be to create an additional action for the radio selection, right? `schools#new` -> `schools#select` -> `schools#check` -> `schools#create`.

I don't _think_ they're too different, right? The existing `SchoolsController` mainly handles the management of creation of a new School, which this new controller is also doing, except for the additional step.

I don't feel too strongly, but maybe adding the additional step to the existing `SchoolsController` is less code to maintain?

Overall, looks good to me!

elceebee (changes requested):
I'm getting an unhandled error when I don't select a school on the placements side. 

https://github.com/DFE-Digital/itt-mentor-services/assets/44073106/3879a45b-368b-45ff-b476-4f7ecf57f4fc

spec/system/claims/support/schools/add_a_school_through_school_options_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+      create(:school, name: "London"),
+    ]
+  end
+
+  before do
+    schools
+    given_i_sign_in_as_colin
+  end
+
+  after { Capybara.app_host = nil }
```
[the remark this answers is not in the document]
CatalinVoineag: I keep copying without thinking

CatalinVoineag:
> If I understand correctly, the split of `SchoolOptions` and `Schools` is to distinctly separate the concerns, right?
> 
> The alternative would be to create an additional action for the radio selection, right? `schools#new` -> `schools#select` -> `schools#check` -> `schools#create`.
> 
> I don't _think_ they're too different, right? The existing `SchoolsController` mainly handles the management of creation of a new School, which this new controller is also doing, except for the additional step.
> 
> I don't feel too strongly, but maybe adding the additional step to the existing `SchoolsController` is less code to maintain?
> 
> Overall, looks good to me!

I didn't want to put it in the normal support schools controller. We already have a check step in there and a search by name and postcode and I'm sure we'll add more things in there like filtering. Because of this, I wanted to keep this 'javascript disabled option' separate, I feel like this will be a done and never touch piece of code as I'm sure no one will actually have javascript disabled. 

I do get your point though, it's doing the same thing just in a different way. I can change it if there is a strong opinion on this.

app/components/school_options_form_component.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    @model = model
+    @url = url
+    @search_param = search_param
+    @schools = schools
+    @hint_link = hint_link
+  end
+
+  def hint_translation
+    if schools.count > School::OPTIONS_PER_PAGE
```
ollietreend, line 23: I'm wondering if it makes more sense for the `OPTIONS_PER_PAGE` constant to exist on the `SchoolOptionsFormComponent` instead of the `School` model.

The component is the only thing that refers to this value. It also seems to be needed for presentational concerns, which makes me wonder whether it's a bit confusing directly on the `School` model.

What do you think?

config/locales/en.yml
```diff
@@ -11,6 +11,15 @@ en:
       scitt: SCITT
       university: University
       lead_school: Lead partner
+    school_options_form_component:
+      add_organisation: Add organisation
+      continue: Continue
+      hint_many_schools_html: Showing the first %{school_count} results. %{link_to} if the school you’re looking for is not listed.
+      hint_html: '%{link_to} if the school you’re looking for is not listed.'
+      hint_link: Change your search
+      hint_many_schools_link: Try narrowing down your search
```
ollietreend, line 20: This is more of a question - not a strong opinion or change request.

It occurs to me that we have a couple of approaches to naming locale keys.

Some locale keys _mirror the string they represent_. For example:

https://github.com/DFE-Digital/itt-mentor-services/blob/083ba33336ddaad4de46f15daa3e18d1b6bab666/config/locales/claims/en/support/schools.yml#L13-L17

In other cases, such as here, the key _does not reflect the string value_. There's no relation between the two. It means that the ERB view is a bit more opaque, because it's not easy to guess what the localised string is.

For example, I have no idea what these lines will output – unless I go and look up the value, or have a fancy IDE configured to show the translations as hints:

https://github.com/DFE-Digital/itt-mentor-services/blob/083ba33336ddaad4de46f15daa3e18d1b6bab666/app/components/school_options_form_component.html.erb#L48-L50

To some extent I assume all apps will have a mixture of both. Not least because Active Record has a specific key format it uses for attribute names and error messages.

Is there any best practice or guidance for which naming approach is best? Or does this seem like a non-issue / I'm just being pedantic?

app/controllers/claims/support/schools_controller.rb
```diff
@@ -30,6 +32,10 @@ def create
 
   private
 
+  def redirect_to_school_options
+    redirect_to new_claims_support_school_option_path(school: { search_param: school_params[:urn] })
+  end
```
ollietreend, line 37: I wonder if this could be more cleanly implemented by considering the JavaScript version to be a "progressive enhancement".

Rhetorical question: If we were building this entire journey from scratch, without any JavaScript, is this how we'd build/structure it?

I think there could be a way for this interaction to be more simply implemented within a single controller. Something like:

```ruby
class SchoolsController
  def new
    @school_form = SchoolOnboardingForm.new
    @schools = search_results if search_query.present?
  end

  private

  def search_query
    params[:search_query]
  end

  def search_results
    School.search_name_urn_postcode(search_query).map(&:decorate)
  end
end
```

So the view `new.html.erb` receives search results _if a search has been performed_, otherwise it shows the default page.

And then we might not actually need a separate 'school options' controller with a duplicate `check` action and views, etc.

What do you think?

spec/system/claims/support/schools/add_a_school_through_school_options_spec.rb
```diff
@@ -0,0 +1,128 @@
+require "rails_helper"
+
+RSpec.describe "Support User adds a School through school options", type: :system do
```
ollietreend, line 3: This might be a more meaningful name to describe the test:

```suggestion
RSpec.describe "Support User adds a School without JavaScript", type: :system do
```

ollietreend (commented):
Great work @CatalinVoineag ⭐ this is looking good.

I've added a few comments of things that might be worth thinking about. I'd be happy to talk through them if you'd like – let me know!

config/locales/en.yml
CatalinVoineag: This is an issue, I guess some translations don't have actions behind them l `continue`. I'll rename them

app/controllers/claims/support/schools_controller.rb
CatalinVoineag: The search-options#new controller method comes after the `schools#new` method but before the `schools#check` method. So we can't reuse the `schools#new` method.

We need a new one. I see there's interest in moving this in the `schools controller` so I'll see how I can put it there

CatalinVoineag: OK, I've added the methods that are neccessary to the schools controller, we can't reuse the `new` or the `check` methods. 

- The `new` method we can't use because the radio form happens after that.
- The `check` method we can't use because we need to redirect away from it when a param that is only sent by javascript, `params.dig(:school, :search_urn)`. We need to display different error messages in the school form object, so we can't tell the school form object to display this error and also go into the `check` controller method because we are redirecting away when 'javascript is disabled'

spec/system/claims/support/schools/add_a_school_through_school_options_spec.rb
ollietreend: @CatalinVoineag sorry – one more thing. Could you also rename the file so it matches the `describe` text?

CatalinVoineag:
> > If I understand correctly, the split of `SchoolOptions` and `Schools` is to distinctly separate the concerns, right?
> > The alternative would be to create an additional action for the radio selection, right? `schools#new` -> `schools#select` -> `schools#check` -> `schools#create`.
> > I don't _think_ they're too different, right? The existing `SchoolsController` mainly handles the management of creation of a new School, which this new controller is also doing, except for the additional step.
> > I don't feel too strongly, but maybe adding the additional step to the existing `SchoolsController` is less code to maintain?
> > Overall, looks good to me!
> 
> I didn't want to put it in the normal support schools controller. We already have a check step in there and a search by name and postcode and I'm sure we'll add more things in there like filtering. Because of this, I wanted to keep this 'javascript disabled option' separate, I feel like this will be a done and never touch piece of code as I'm sure no one will actually have javascript disabled.
> 
> I do get your point though, it's doing the same thing just in a different way. I can change it if there is a strong opinion on this.

Ended up with 1 controller https://github.com/DFE-Digital/itt-mentor-services/pull/115#discussion_r1455422605

app/controllers/claims/support/schools_controller.rb
ollietreend: Catalin and I have just chatted about this. We both think that there could _potentially_ be a refactor – to simplify the JS and non-JS implementations – but we agree that it falls beyond the scope of this PR.

I'm going to mark this as 'resolved', but I may follow up with some more thinking in a future PR.

ollietreend (approved):
Nice work 👍🏻 thanks for talking it through with me, too.

I'm happy with the PR how it currently is. I'll do some more thinking about a potential refactor, and will follow up with another PR if I think any improvements can be made.

PR #130, opened 2024-01-17
Lb/provider show page for support user

elceebee:
## Context

As a support user, I should be able to click on any the name of any onboarded organisation and view the details

- Refactored contact details for organisations (used for both claims and placements, schools and providers)
- Refactored organisation details (used for both claims and placements, schools and providers)
- Refactored the support secondary navigation for placements
- Added the show page (which uses all of the above)

## Guidance to review

Login as a placement support user
Click on an organisation that is a provider (not a school)
View the provider details

Also, please click around a lot and make sure I haven't broken anything with the refactoring -- fingers crossed our test coverage is good enough to have picked up any problems, but good to check things manually as well. 

## Link to Trello card
https://trello.com/c/FHlaGUxg

## Things to check

## Screenshots

image


app/helpers/placements/support/organisations_helper.rb
```diff
@@ -11,6 +11,19 @@ def no_results(search = "", filters = [])
...
+    when "Placements::School"
+      placements_support_school_path(organisation)
+    when "School"
+      placements_support_school_path(organisation)
+    when "Placements::Provider"
+      placements_support_provider_path(organisation)
+    when "Provider"
+      placements_support_provider_path(organisation)
+    end
+  end
```
Nitemaeric, line 25: This supports inheritence:

```suggestion
  def organisation_url(organisation)
    case organisation
    when School
      placements_support_school_path(organisation)
    when Provider
      placements_support_provider_path(organisation)
    end
  end
```

elceebee:
[Provider show page for placement support user](https://trello.com/c/FHlaGUxg/126-provider-show-page-for-placement-support-user)

Nitemaeric (approved):

app/helpers/placements/support/organisations_helper.rb
```diff
@@ -11,6 +11,15 @@ def no_results(search = "", filters = [])
...
   end
 
+  def organisation_url(organisation)
+    case organisation.class.name
+    when "School"
+      placements_support_school_path(organisation)
+    when "Provider"
+      placements_support_provider_path(organisation)
+    end
+  end
```
Nitemaeric, line 21: This won't work for `Placements::School` models, as you're comparing the class names as strings, not the actual classes/instances.

app/controllers/placements/support/providers_controller.rb
```diff
@@ -3,6 +3,10 @@ def new
     @provider_form = ProviderOnboardingForm.new
   end
 
+  def show
+    @provider = Provider.placements.find(params.require(:id)).decorate
```
ollietreend, line 7: I think we can also do this. Not sure if we have a preference?

```suggestion
    @provider = Placements::Provider.find(params.require(:id)).decorate
```

app/views/placements/support/_secondary_navigation.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= render SecondaryNavigationComponent.new do |component| %>
+  <% component.with_navigation_item t(".details"), organisation_url(organisation) %>
+  <% component.with_navigation_item t(".users"), "#" %>
```
JamieCleare2525, line 3: Could you put in the users paths.
You might need a helper for this, similar to what you've done above.

app/helpers/placements/support/organisations_helper.rb
ollietreend, line 25: I _think_ this should work because you can [match an object's type in case statements](https://batsov.com/articles/2012/10/14/ruby-tip-number-3-matching-on-an-objects-class-in-a-case-expression/).

```suggestion
  def organisation_url(organisation)
    case organisation
    when School
      placements_support_school_path(organisation)
    when Provider
      placements_support_provider_path(organisation)
    end
  end
```

I'm hoping it will handle both the child class (e.g. `Placements::School`) and the parent class (e.g. `School`) due to them both being a type of `School`.

elceebee: Sweet, I'll give it a go

PR #131, opened 2024-01-17
Reorganise locale files

Nitemaeric:
## Context

- The locale files nested structure doesn't match the locale key structure. This doesn't break anything, but it's strange.
- Service Name deserves to be outside of the scope of the application layout. We will need to use it elsewhere, such as in emails, notifications, references to the service name itself in all views. These will **need** to be consistent.

- Reorganise locales to match locale key hierarchies
- Update `service_name` locale key to be view agnostic.

## Guidance to review

- Check that locale keys match their nested file structure.
  - E.g. The `en.claims.support.users.index.page_title` locale key should be hosted at `config/locales/en/claims/support/users.yml`.

## Things to check


Nitemaeric:
[Reorganise locale file hierarchies](https://trello.com/c/T2EBluH3/129-reorganise-locale-file-hierarchies)

CatalinVoineag (approved):
This looks good to me 👍🏻

Nitemaeric:
@CatalinVoineag I've also updated the CODEOWNERS to match, so reviewers are added by default when we add locale changes :)

config/locales/en/placements.yml
```diff
@@ -0,0 +1,3 @@
+en:
+  placements:
+    service_name: Manage School Placements
```
JamieCleare2525, line 3:
```suggestion
    service_name: Manage school placements
```

CatalinVoineag:
> @CatalinVoineag I've also updated the CODEOWNERS to match, so reviewers are added by default when we add locale changes :)

Good idea 👍🏻

PR #135, opened 2024-01-17
Claims: Support User CRUD

Nitemaeric:
## Context

We need to add and manage Support Users.

- List all Claims Support Users
- Show a single Claims Support User
- Add a new Claims Support User
  - Add SupportUserMailer to send invitation emails
- Add email domain validation to Support Users
- Remove a Claims Support User
- Add `with(service:)` parameters to Mailers to infer the host. No more sending through urls. **We can now use url helpers within mailers themselves.**

## Guidance to review

- Admittedly, this PR would have been more manageable if broken into individual CRUD actions. Sorry. 🙇 
- This change adds Support User management to the claims service. The `AddSupportUser` and related mailer should work cross-service.
- I've opted to using a `Support::SupportUsersController` instead of `Support::UsersController` as there is a possibility that the support console adds the functionality to view _all_ users in the future and a `Support::UsersController` would be the best place for that logic.
- Support users require an `@education.gov.uk` email address.
- After this is merged in, you'll want to run `rails db:seed:replant` to update Colin's email address.
  - You'll also need to restart the server to update the list of `PERSONA_EMAILS`.

### Conventions introduced

- `t(".success")` and `t(".failure")` keys for controller action `notice` and `alert` values.
- `t(".submit")` key for main form submission text.

## Screenshots

| URL | Screenshot |
| --- | ---------- |
| /support/support_users | CleanShot 2024-02-01 at 14 48 35 |
| /support/support_users/new | CleanShot 2024-02-01 at 14 49 35 |
| /support/support_users/check | CleanShot 2024-02-01 at 14 50 30 |
| /support/support_users/{id} | CleanShot 2024-02-01 at 14 50 53 |

| Error | Screenshot |
| ----- | ---------- |
| Invalid @education.gov.uk email address | CleanShot 2024-02-01 at 14 53 59 |


Nitemaeric:
[As a Support User, I should be able to add additional Support Users](https://trello.com/c/mr9m4lAs/94-as-a-support-user-i-should-be-able-to-add-additional-support-users)

app/controllers/claims/support/support_users_controller.rb
```diff
@@ -0,0 +1,39 @@
+class Claims::Support::SupportUsersController < Claims::Support::ApplicationController
+  def index
+    @users = Claims::User.support_users
+  end
+
+  def new
```
elceebee, line 6: If first name, last name, email params are available, you should initialise the form with them. See Jamie's work on the add user journey. Otherwise, the 'check page' just takes you back to an empty form which is a very unsatisfying journey.

app/views/claims/support/support_users/check.html.erb
```diff
@@ -0,0 +1,53 @@
...
+        <label class="govuk-label govuk-label--l">
+          <span class="govuk-caption-l"><%= t(".caption") %></span>
+          <%= t(".title") %>
+        </label>
+
+        <%= govuk_summary_list do |summary_list| %>
+          <% summary_list.with_row do |row| %>
+            <% row.with_key(text: User.human_attribute_name(:first_name)) %>
+            <% row.with_value(text: @user.first_name) %>
+            <% row.with_action(text: t(".change"), href: new_claims_support_support_user_path(@user), visually_hidden_text: @user.first_name) %>
```
elceebee, line 24: The action href should include the first name, last name, email params.

app/views/claims/support/support_users/check.html.erb
```diff
@@ -0,0 +1,53 @@
+<% content_for(:page_title) { t(".page_title") } %>
+
+<%= content_for(:before_content) do %>
+  <%= govuk_back_link(href: new_claims_support_support_user_path) %>
```
elceebee, line 4: this back link should include the form params.

app/views/claims/support/support_users/show.html.erb
```diff
@@ -0,0 +1,33 @@
...
+        <% summary_list.with_row do |row| %>
+          <% row.with_key(text: User.human_attribute_name(:last_name)) %>
+          <% row.with_value(text: @user.last_name) %>
+          <% row.with_action(text: t(".change"), href: edit_claims_support_support_user_path(@user), visually_hidden_text: @user.last_name) %>
+        <% end %>
+
+        <% summary_list.with_row do |row| %>
+          <% row.with_key(text: User.human_attribute_name(:email)) %>
+          <% row.with_value(text: @user.email) %>
+          <% row.with_action(text: t(".change"), href: edit_claims_support_support_user_path(@user), visually_hidden_text: @user.email) %>
```
elceebee, line 28: When we create a user, the first and last name are used for the invite. After that, we get the information from DfE Sign in -- it's not something that can be changed. I talked to Simon about this briefly when i was in the office on Thursday. So these 'change' links should be removed.

Again, this is the 3rd time we have created this show page, so perhaps it's time to create a shared partial.

app/services/add_support_user.rb
```diff
@@ -0,0 +1,25 @@
+class AddSupportUser
```
elceebee, line 1: I don't feel strongly about this, but we are using form objects else where, that might be useful pattern to maintain, but again, I don't really mind.

elceebee (changes requested):
Would be good to 
- change back / change links on the check page to pre-populate the form
- remove the change links from the show page as they don't go anywhere and there is no planned user journey for them

app/controllers/claims/support/support_users_controller.rb
Nitemaeric: I've updated the flows to support adding default values via query params.

app/views/claims/support/support_users/check.html.erb
Nitemaeric: Same as above. Now addressed.

app/views/claims/support/support_users/check.html.erb
Nitemaeric: Same as above. Now addressed.

app/views/claims/support/support_users/show.html.erb
Nitemaeric: Contextually, the partial differs from other user show pages. It only happens to show the same content in the current design.

app/models/user.rb
```diff
@@ -15,13 +15,16 @@
...
+  SUPPORT_EMAIL_REGEXP = /\A[a-zA-Z0-9.!\#$%&'*+\/=?^_`{|}~-]+@education.gov.uk/
+
+  has_many :memberships, dependent: :destroy
 
   validates :first_name, presence: true
   validates :last_name, presence: true
   validates :email, presence: true, format: { with: URI::MailTo::EMAIL_REGEXP }
   validates :type, presence: true
   validates :email, uniqueness: { scope: :type, case_sensitive: false }
+  validates :email, format: { with: SUPPORT_EMAIL_REGEXP, message: :invalid_support_email }, if: :support_user?
```
JamieCleare2525, line 27: Should this logic be in the `ActsAsSupportUser` concern, as this is only relevant to SupportUsers.

Nitemaeric: Sounds good :)

Nitemaeric: Addressed in https://github.com/DFE-Digital/itt-mentor-services/pull/135/commits/a688a17652a1abc5964759bc7d9b022fa5aaa59c

Nitemaeric:
[As a Support User, I should not be able to sign in without an ...@education.gov.uk email address](https://trello.com/c/CGrTfKzl/109-as-a-support-user-i-should-not-be-able-to-sign-in-without-an-educationgovuk-email-address)

app/controllers/claims/support/support_users_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    @support_user = if params[:support_user].present?
+                      Claims::SupportUser.new(support_user_params)
+                    else
+                      Claims::SupportUser.new
+                    end
+  end
+
+  def check
+    @support_user = if params[:id].present?
```
JamieCleare2525, line 17: Is this necessary? I can't see a case where `:id` would ever be passed to the `:check` method

Nitemaeric: You're right. Sorry, this was a relic from the edit flow.

JamieCleare2525:
Tested on https://track-and-pay-135.test.teacherservices.cloud/support/support_users/check?support_user%5Bfirst_name%5D=Test&support_user%5Blast_name%5D=McTesterson&support_user%5Bemail%5D=%40education.gov.gov.uk
screencapture-track-and-pay-135-test-teacherservices-cloud-support-support-users-check-2024-02-01-15_36_49

1. The secondary nav isn't highlighted on "User".
2. The wrong error message is displaying when adding a support user. It should be "Enter a Department for Education email address in the correct format, like name@education.gov.uk", perhaps this has occured since moving the validation to the concern.

app/mailers/application_mailer.rb
```diff
@@ -1,8 +1,25 @@
 class ApplicationMailer < Mail::Notify::Mailer
   GENERIC_NOTIFY_TEMPLATE = "a2c0c8f0-6d15-47db-b334-5fe198da1740".freeze
 
+  default from: "no-reply@education.gov.uk"
```
Nitemaeric, line 4: This is needed for "test" emails to be valid.

JamieCleare2525:
screencapture-track-and-pay-135-test-teacherservices-cloud-support-support-users-083a82c5-f890-4854-a401-13feb87f204b-2024-02-01-15_43_23
"Users" is also not highlighted in the Support Users show page.
I'm also not sure that "test.tester@education.gov.ukk" should be allowed as an email address.

Nitemaeric:
> "Users" is also not highlighted in the Support Users show page.

I'll address this in a separate PR.

> I'm also not sure that "test.tester@education.gov.ukk" should be allowed as an email address.

Nice catch!

app/models/concerns/acts_as_support_user.rb
```diff
@@ -1,7 +1,11 @@
 module ActsAsSupportUser
   extend ActiveSupport::Concern
 
+  SUPPORT_EMAIL_REGEXP = /\A[a-zA-Z0-9.!\#$%&'*+\/=?^_`{|}~-]+@education.gov.uk\z/
+
   included do
+    validates :email, format: { with: SUPPORT_EMAIL_REGEXP, message: :invalid_support_email }
```
JamieCleare2525, line 7:
```suggestion
    validates :email, presence: true, format: { with: SUPPORT_EMAIL_REGEXP, message: :invalid_support_email }
```

CatalinVoineag (approved):
Once the comments that are already here are addressed I think this can go out. Very nice work!

app/models/concerns/acts_as_support_user.rb
Nitemaeric: 🙇

Nitemaeric:
> Approved. Given the understanding that the missing navigation highlighting behaviour across all of these pages will be resolved in a different PR.

Thanks for your patience and the thorough review! 🙇

app/views/claims/support/support_users/new.html.erb
```diff
@@ -0,0 +1,42 @@
+<% content_for(:page_title) { t(".page_title") } %>
```
elceebee, line 1: Simon pointed out to me yesterday that if there are errors on the page, the page title should be prefixed with "Errors: " It looks like the prototypes are busted, so I can't screen shot an example, but I did it correctly for adding provider / school users as non-support users on the placements side. Something like
`<%= content_for :page_title, @support_user.errors.any? ? t(".page_title_with_error") : t(".page_title") %>`

spec/system/claims/support/support_users/add_a_support_user_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  scenario "Attempt to add a support user without an @education.gov.uk email address" do
+    when_i_sign_in_as_a_support_user(support_user)
+    and_i_visit_the_support_users_page
+    and_i_click_on_add_a_support_user
+    then_i_fill_in_the_support_user_form(email_address: "john.doe@example.com")
+    then_i_see_an_error
+  end
+
+  private
```
elceebee, line 31: I think there are two scenarios missing here. One "Attempt to add an already-existing support user". 

And the other (which can be handled implicitly in the first scenario), is navigating back to the form using the change or the back button, making a change and then continuing. That would test the functionality, but it also tests that all the expected elements (the back button, the change links) are on the page. This ensures we'll catch it if we are make changes to the UI and accidentally delete things. (been known to happen...)

elceebee (commented):
Just a few little things

elceebee (approved):
Thanks for addressing all my comments! I'm happy for this to go in now.

PR #140, opened 2024-01-19
Claim multi step form V1

CatalinVoineag:
## Context

We don't know exactly how the claim form will look or work. But we do know it will be a multi-step form. So this is the V1 of the claim form. With changes to a lot of the code here likely coming in the next few days. 

This is more of a prototype of how the claim form might work.

Might be easier to review by commit, they also have more description:

https://github.com/DFE-Digital/itt-mentor-services/pull/140/commits

## Known issues
- I needed a select search to pass the id of a record to the controller. I didn't want to refactor the ones we have for schools and providers. I found that the [Accessible Autocomplete](https://github.com/alphagov/accessible-autocomplete?tab=readme-ov-file#progressive-enhancement) works with a select as well but it only searches by name and the search is done on the client side. Not really a problem for a small list like providers but the fact that we can't search by postcode or something else is an issue, we have to come back and refactor this.
- You can't put a blank value after saving a provider or mentor, this again is because of the accessible autocomplete.

- Seed file
- Migrations
- Claim form
- Claim controller, decorator, views
- Specs

## Guidance to review

- Run migrations
- Run seed file to add the mentors
- Sign in as a normal user
- Go to the claims index page
- Add a claim
- Add a provider
- Add a mentor
- Go back to the claims index.

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/a7c3916a-a8ea-4aa8-9717-9059c46ea388


app/forms/claim_form.rb
```diff
@@ -0,0 +1,45 @@
...
+    claim.errors.blank?
+  end
+
+  def valid_mentors?
+    return true unless step == "mentors"
+
+    claim.mentor_trainings.each do |mentor_training|
+      if mentor_training.teacher_id.nil?
+        mentor_training.errors.add(:teacher_id, :blank)
+        claim.errors.add(:base, :enter_teacher_name)
```
CatalinVoineag, line 39: This might spit out too many errors if we allow the user to add multiple mentors at the same time. At this point though, we are not sure if the user can add multiple mentors or just one, or the process might be different. So the idea is to leave it as it is and come back when we have more details.

CatalinVoineag:
[As a User, I should be able to start a new claim​](https://trello.com/c/SuTciR5w/95-as-a-user-i-should-be-able-to-start-a-new-claim)

Gemfile
```diff
@@ -27,6 +27,9 @@ gem "view_component"
 # GOV.UK Notify
 gem "mail-notify"
 
+# Hotwire's modest JavaScript framework [https://stimulus.hotwired.dev]
+gem "stimulus-rails"
+
```
ollietreend, line 32: I'm new to Stimulus. I've never needed to use it before, but I can see that it brings some advantages. I like the way it defines an explicit relationship between HTML and associated JS behaviour.

However, if this is something that's worth using, then I think we should adopt it wholesale. Thankfully we're still early on so that's not too disruptive of a change to make.

I'd consider this to be an architectural decision, and therefore I think we should probably open a standalone PR which:

- Adds an ADR to explain how/when/why we use Stimulus
- Migrates existing JavaScript within the app to use Stimulus

What do you think @CatalinVoineag?

ollietreend (changes requested):

Gemfile
CatalinVoineag: Sure, I'll get that PR merged first

app/models/teacher.rb
```diff
@@ -0,0 +1,12 @@
...
+# Table name: teachers
+#
+#  id         :uuid             not null, primary key
+#  first_name :string           not null
+#  mentor     :boolean          default(FALSE)
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+class Teacher < ApplicationRecord
+end
```
ollietreend, line 12: I can see you've gone with the name "Teacher" rather than "Mentor", which means we need to specify `source: "teacher"` on the `Claim#mentors` association.

I'm wondering: what led you to choose this name?

In the [data model document](https://github.com/DFE-Digital/itt-mentor-services/blob/beaa69d2273f9db16b52e341ed04d519675b3517/docs/data-model.md), we have a concept of a "Mentor" and a "Trainee" being two independent entities. Both have a TRN so they can be looked up in DQT – and I'm wondering if this is the reason you've named the model "Teacher"?

However conceptually Mentors and Trainees two different things – they have different meanings within the service. And in fact, based on our current definition of MVP, we won't actually need Trainee yet.

So I'm wondering if we'd be better off just calling this "Mentor"?

app/controllers/claims/schools/claims_controller.rb
```diff
@@ -1,3 +1,55 @@
 class Claims::Schools::ClaimsController < ApplicationController
   include Claims::BelongsToSchool
+
+  def index
+    render locals: { claims: @school.claim_applications }
+  end
+
+  def new
+    claim = Claim.create!(school_id: params.require(:school_id), draft: true)
```
ollietreend, line 9: This might be more succinct (thanks to `Claims::BelongsToSchool`):

```suggestion
    claim = @school.create_claim!(draft: true)
```

app/controllers/claims/schools/claims_controller.rb
```diff
@@ -1,3 +1,55 @@
...
+  end
+
+  def claim_params
+    params.require(:claim).permit(
+      mentor_trainings_attributes: %w[provider_id teacher_id id],
+    )
+  end
+
+  def claim
+    @claim ||= Claim.find(params.require(:id)).decorate
```
ollietreend, line 49: Should we also scope this "find" to the current school?

```suggestion
    @claim ||= @school.claims.find(params.require(:id)).decorate
```

app/models/claims/school.rb
```diff
@@ -48,4 +48,5 @@ class Claims::School < School
   default_scope { claims }
 
   has_many :users, -> { claims }, through: :memberships
+  has_many :claim_applications, class_name: "Claim"
```
ollietreend, line 51: Would this be more succinct if modelled as "claims"?

```suggestion
  has_many :claims
```

app/forms/claim_form.rb
```diff
@@ -0,0 +1,45 @@
+class ClaimForm
+  include ActiveModel::Model
+  STEPS = %w[providers mentors].freeze
```
ollietreend, line 3: No change needed at this point. But in the future, I could imagine you wanting to have one Form Object per 'step' of the Claim form.

Otherwise I could imagine this class becoming huge and conditionally doing different things depending on the current step.

app/models/claim.rb
```diff
@@ -0,0 +1,27 @@
...
+# Indexes
+#
+#  index_claims_on_school_id  (school_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (school_id => schools.id)
+#
+class Claim < ApplicationRecord
+  belongs_to :school
```
ollietreend, line 20: If you go with a non-standard name for the association, presumably we'll also want:

```suggestion
  belongs_to :school, inverse_of: :claim_applications
```

Although IMHO I think it'd probably be cleaner just to go with the conventional `:claims` name for this association.

app/models/mentor_training.rb
```diff
@@ -0,0 +1,31 @@
...
+# Foreign Keys
+#
+#  fk_rails_...  (claim_id => claims.id)
+#  fk_rails_...  (provider_id => providers.id)
+#  fk_rails_...  (teacher_id => teachers.id)
+#
+class MentorTraining < ApplicationRecord
+  belongs_to :claim, class_name: "Claim"
+  belongs_to :teacher, optional: true
+  belongs_to :provider, optional: true
```
ollietreend, line 30: Why are these two associations optional? 🤔

I'd have thought Mentor Training required a Mentor and a Provider, and that Claim was optional.

app/views/claims/schools/claims/_mentors_form.html.erb
```diff
@@ -0,0 +1,41 @@
...
+  <h1 class="govuk-heading-l"><%= t(".heading") %></h1>
+
+  <%= f.fields_for :mentor_trainings do |mentor_form| %>
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <div class="govuk-form-group">
+
+          <%= mentor_form.govuk_collection_select(
+            :teacher_id,
+            Teacher.all,
```
ollietreend, line 24: I think Mentor should additionally belong to a School. (As per the [data model document](https://github.com/DFE-Digital/itt-mentor-services/blob/beaa69d2273f9db16b52e341ed04d519675b3517/docs/data-model.md).)

That way we'd be able to scope this list of Mentors to only those who belong to the current School.

```suggestion
            @school.mentors,
```

db/seeds.rb
```diff
@@ -42,4 +42,9 @@
     end
   end
 
+# Create dummy teachers
+%w[Sarah Bob Pomona].each do |first_name|
+  Teacher.find_or_create_by!(first_name:)
+end
```
ollietreend, line 48: I wonder if we could/should start using factories in this seeds file? 🤔

ollietreend (commented):
Really nice work on this! 🤓

I've added a few comments, mostly to do with the data model side of things. Happy to talk through any of it!

app/models/teacher.rb
CatalinVoineag: I named it teacher because from what I understand is that DQT doesn't know which people are mentors and which are teachers, so I assumed we will import all teachers from DQT and have to distinguish between teachers and mentors so a boolean field on it was doing the job. I can change it to mentors though, don't mind

app/models/claims/school.rb
CatalinVoineag: I did use has many claims but it conflicts with the `claims` boolean attribute on the school record.

app/forms/claim_form.rb
CatalinVoineag: Yeah, good point. At this point it didn't feel necessary

app/models/mentor_training.rb
CatalinVoineag: It's because when creating a mentor_training in the provider step for example, we don't have a mentor on the mentor training for example. So there is a state where a mentor training has a provider or a mentor without the other.

db/seeds.rb
CatalinVoineag: I still want to specify the name of a teacher/mentor here don't want to use faker to generate random names every time. So I don't really see a need to use factories here

Gemfile
ollietreend: Resolved by #152

app/models/claims/school.rb
ollietreend: D'oh! 🤦🏻 Makes me wish we'd named the boolean attribute `claims_enabled`, because it feels more natural for this association to be called `claims`.

db/seeds.rb
ollietreend: Cool cool. I don't have a strong opinion either way.

app/models/teacher.rb
CatalinVoineag: I guess it's a bit misleading as I'm not actually using the `mentor` boolean, I'll make some changes so that it's actually used

CatalinVoineag: Sure that makes sense. I didn't know all of this. I'll change it to mentor

app/javascript/controllers/select_autocomplete_controller.js
```diff
@@ -0,0 +1,15 @@
...
+
+// Connects to data-controller="select-autocomplete"
+export default class extends Controller {
+  static targets = [ "input" ]
+
+  connect() {
+    accessibleAutocomplete.enhanceSelectElement({
+      selectElement: document.querySelector(
+        `#${this.inputTarget.id}`
+        )
```
ollietreend, line 12: The indentation of this line is slightly off. Do we have a linter enabled for JS? 🤔

ollietreend, line 12: Shouldn't it be possible to use the target directly? Rather than querying the document for the element that we already have available to us.

```suggestion
      selectElement: this.inputTarget
```

app/views/claims/schools/claims/_mentors_form.html.erb
```diff
@@ -0,0 +1,41 @@
...
+            :id,
+            :first_name,
+            data: {
+              controller: "select-autocomplete",
+              select_autocomplete_target: "input",
+            },
+            id: "select-autocomplete-input",
```
ollietreend, line 31: Presumably this doesn't need a hardcoded ID now that we're using a Stimulus target?

ollietreend (commented):
Sorry - a couple more things. I may be misunderstanding because I've not used Stimulus before.

CatalinVoineag:
> Sorry - a couple more things. I may be misunderstanding because I've not used Stimulus before.

I'll do all the changes after you changed the claims boolean field on the school, let me know if you need a hand with that if you're too busy. I can pick it up

app/javascript/controllers/select_autocomplete_controller.js
Nitemaeric: We don't. We'd need to add in ESLint.

Also, out of curiosity, is there anything stopping us from doing...

```js
    accessibleAutocomplete.enhanceSelectElement({
      selectElement: this.inputTarget
    })
```

ollietreend:
@CatalinVoineag I've just merged in #157 which frees up `School#claims` for the association! ✨

CatalinVoineag:
> @CatalinVoineag I've just merged in #157 which frees up `School#claims` for the association! ✨

Thank you. I'll get this ready

spec/factories/mentors.rb
```diff
@@ -0,0 +1,32 @@
...
+#
+FactoryBot.define do
+  factory :mentor do
+    association :school
+
+    first_name { Faker::Name.first_name }
+    last_name { Faker::Name.last_name }
+    sequence(:trn) { _1 }
+
+    initialize_with { Mentor.find_or_create_by(first_name:, school:) }
```
ollietreend, line 30: What's the significance of this? Is it needed?

It'll lead to unusual behaviour because `build(:mentor)` will now perform actions against the database (i.e. find or create) – whereas the 'build' strategy is usually expected to initialise a record but not persist it.

ollietreend (approved):
Just one final question from me – I'm not sure if we really want the `initialize_by`, or if/why it's necessary.

But I'm approving this because you've addressed all my other points. Thanks!

Nice work! 🚀

app/controllers/claims/schools/claims_controller.rb
```diff
@@ -1,3 +1,51 @@
 class Claims::Schools::ClaimsController < ApplicationController
   include Claims::BelongsToSchool
+
+  def index
+    render locals: { claims: @school.claims }
```
Nitemaeric, line 5: Is there a strong reason behind passing this through as a local instead of an instance variable?

```suggestion
    @claims = @school.claims
```

spec/models/claim_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+RSpec.describe Claim, type: :model do
+  context "associations" do
+    it { should belong_to(:school) }
+    it { should have_many(:mentor_trainings) }
+    it { should have_many(:mentors).through(:mentor_trainings) }
+    it { should have_many(:providers).through(:mentor_trainings) }
+  end
+
+  it { should accept_nested_attributes_for(:mentor_trainings) }
```
Nitemaeric, line 29: This kind of feels like it belongs in the `associations` context. Would you agree?

spec/models/mentor_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+require "rails_helper"
+
+RSpec.describe Mentor, type: :model do
+  context "associations" do
+    it { should belong_to(:school) }
+  end
+
+  context "validations" do
+    subject { create(:mentor) }
```
Nitemaeric, line 30: Let's avoid creating in unit tests unless it's required.

```suggestion
```

I _think_ the default subject for a model test is a mentor instance. If not, let's make it a `build` call instead.

```suggestion
    subject { build(:mentor) }
```

spec/factories/mentors.rb
CatalinVoineag: It's more to accommodate the seed file, to find or initialize mentors https://github.com/DFE-Digital/itt-mentor-services/blob/452a9e2751aedf0d2340c761250f4f933d1a9111/db/seeds.rb#L51

I'll just not use the factory in the seed file

app/controllers/claims/schools/claims_controller.rb
CatalinVoineag: Personally, I prefer to use locals over instance variables. For this index method instance, variables work but when you have to render partials, like in the edit controller method and view I think you need to use locals. So because of this I prefer to always use locals. But I know there are places where we use instance variables and places where we use both 🤷🏻

Nitemaeric: As far as I can tell, the distinction of using instance variables vs locals is:

- Instance Variables for passing variables from Controllers down to Views
- Locals when passing variables from a View into a Partial

Would you object to this?

I'm a personal fan of [this gem](https://github.com/hashrocket/decent_exposure) that cleans this up a lot, but it's not heavily maintained. But this is just to share "exposure" 😆

CatalinVoineag: I don't object to the distinction, I'm just saying that we can just use locals to do both to reduce a little context switching. 

About the gem, it looks good but I feel like we are bringing a 3rd party dependency to do something we can easily do ourselves. and if we do decide to use it should we ask placements as well?

CatalinVoineag: For now I can just change the methods that don't require locals to use instance variables

CatalinVoineag: Addressed in https://github.com/DFE-Digital/itt-mentor-services/pull/140/commits/319fcb453b11564aaa3b3be6843e8138378a34e4

Nitemaeric: Sorry, I don't think I was clear in my latest comment. I meant that there's a nice gem out there, but I don't trust it enough to add into this project.

app/controllers/claims/schools/claims_controller.rb
```diff
@@ -1,3 +1,51 @@
...
+
+  private
+
+  def mentor_trainings
+    claim.mentor_trainings.build if claim.mentor_trainings.blank?
+  end
+
+  def claim_params
+    params.require(:claim).permit(
+      mentor_trainings_attributes: %w[provider_id mentor_id id],
```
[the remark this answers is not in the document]
CatalinVoineag: fixed in https://github.com/DFE-Digital/itt-mentor-services/pull/140/commits/f8da8169bb14c56d39041d3def8ad976e11c0455

app/forms/claim_form.rb
```diff
@@ -0,0 +1,45 @@
...
+  def save
+    claim.assign_attributes(claim_params)
+    return false unless valid?
+
+    claim.save!
+  end
+
+  def valid?
+    claim.valid? && valid_providers? && valid_mentors?
+  end
```
Nitemaeric, line 16: 👍 I think form objects should mirror active record methods as you've done. Nice!

Nitemaeric (approved):
LGTM.

I think I have some nitpicks in mind, but nothing big enough to block progress! 🚀

PR #142, opened 2024-01-22
Lb/bug fix membership of multiple organisations

elceebee:
## Context

As a support user, I should be able to add the same user to multiple organisations 

Added a UserInviteForm to be used by both placements and claims
Added unique index on org and user id on the membership model
Added lots of test scenarios

## Guidance to review

You'll need to run migrations
Check you can add the same user to multiple organisations
Check you see errors if you try to add the same user to the same organisation. 

## Link to Trello card

https://trello.com/c/XnT1qjyj

## Things to check

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/44073106/f3888d94-160f-4254-aa0b-cf07f647c828


elceebee:
[Bug - user should be able to be a member of multiple organisations](https://trello.com/c/XnT1qjyj/132-bug-user-should-be-able-to-be-a-member-of-multiple-organisations)

CatalinVoineag (approved):
This looks very good! Thank you for spotting and fixing it 👍🏻

spec/models/membership_spec.rb
```diff
@@ -11,8 +11,9 @@
 #
 # Indexes
 #
-#  index_memberships_on_organisation  (organisation_type,organisation_id)
```
JamieCleare2525, line 14: Could you add a validation for the unique org-to-user association?

app/forms/user_invite_form.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def membership
+    @membership ||= user.memberships.new(organisation:)
+  end
+
+  def send_invite_email
+    url = Rails.application.routes.url_helpers.sign_in_url(host:)
+    NotifyMailer.send_organisation_invite_email(user, organisation, url).deliver_later
+  end
```
Nitemaeric, line 60: I don't think we should be adding business logic to form objects, right? The way I understand form objects is they are a way to apply data modelling and validations, like models, to forms that do not map 1-1 from model to view.

What do you think?

Nitemaeric (changes requested):
I'm adding this request changes in to block merging for now, I'll add in some more comments soon.

app/forms/user_invite_form.rb
elceebee: But this is a form for inviting a user, and kicking off an email is essential to inviting. So I think it is the most sensible place for it. Otherwise we just call a mailer from the controller, definitely the wrong place for business logic.

Nitemaeric: Would that business logic belong to a service object?

elceebee: All that service object would do now is call the mailer. Before it was saving -- now that lives in the form. I think having a service object that just calls a single method is an unhelpful level of abstraction.

elceebee: Added user invite service here https://github.com/DFE-Digital/itt-mentor-services/pull/142/commits/dc405e51987c00c9aabe7d2faeb4b97ed2c2fe0c

Nitemaeric (approved):

PR #143, opened 2024-01-22
Jc/non js provider onboarding

JamieCleare2525:
## Context

Replicate "Javascript Disabled" functionality from School Onboarding onto Provider Onboarding

- Add controller methods, views, etc. to replicate the process used when onboarding school, when javascript is disabled, for Providers.

## Guidance to review

⚠️ DISABLE YOUR JAVASCRIPT ⚠️ 
- (On Placements) Sign in as Colin
- Click "Add Organisation"
- Select "ITT Provider" radio button
- Click "Continue"
- Enter a provider name/postcode/etc into the text input (i.e Bath)
- Click "Continue"
- You should now see a selection of radio buttons, display options for Providers related to your search.

## Link to Trello card

https://trello.com/c/67HdUwFH/117-javascript-disabled-version-of-provider-and-school-onboarding-pages

## Screenshots
Screenshot 2024-01-22 at 12 25 39


CatalinVoineag (approved):
Dev tested this on local, works as expected, nice work!

app/forms/provider_onboarding_form.rb
```diff
@@ -28,4 +28,16 @@ def provider_already_onboarded?
       errors.add(:code, :already_added, provider_name: provider.name)
     end
   end
+
+  def code_presence
+    errors.add(:code, code_error_message) if code.blank?
```
elceebee, line 33: I'm not positive this is where the issue is, but the href in the error summary does not take you to the input field. The error summary is looking for a field with id "provider-code-field-error"
image

app/views/placements/support/providers/provider_options.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= content_for(:before_content) do %>
+  <%= govuk_back_link(href: new_placements_support_provider_path) %>
```
elceebee, line 2: I think we want to preserve the search term in the form -- that's what happens on the prototype. But if you've heard differently from Simon, just ignore this.

elceebee (commented):
I think it would be good to sort out the error summary navigation with this PR, but if is massively complicated, could be done in a separate one. What do you think?

app/views/placements/support/providers/provider_options.html.erb
elceebee: As discussed, this is rather complicated due to handling the search params different for the js and non-js versions. Should be a separate ticket.

JamieCleare2525: Given how the JS autocomplete works this will need to be separated out into a more complex ticket.
Likely this will need event listeners added, etc.
https://trello.com/c/e5SwYKuK/138-fix-back-links-on-providers-schools-onboarding-pages-to-pre-fill-the-search-input-new-page-with-users-input

elceebee (approved):
Thanks for fixing the error summary bug. Understand we'll need to pick up the search params issue in a separate ticket (outside the scope of this one). Nice one

PR #151, opened 2024-01-23
Add users list for claims school page

gms-gs:
## Context

This change adds the Users flow to the claims school page

- Sign in as a lead mentor claims user
- Go to `http://claims.localhost:3000/schools/{id}/users` to see a list of users
- Add a user
- Go to the users index page
- Find the added user
- Click on the users name 

## Link to Trello card

https://trello.com/c/Lxs1LngQ/97-create-the-school-users-page

## Things to check

## Screenshots
Screenshot 2024-01-23 at 07 14 59
Screenshot 2024-01-23 at 07 15 20
Screenshot 2024-01-23 at 07 15 25
Screenshot 2024-01-23 at 07 15 29
Screenshot 2024-01-23 at 07 15 35

### DIFF
--- a/app/controllers/claims/schools/users_controller.rb
+++ b/app/controllers/claims/schools/users_controller.rb
@@ -1,3 +1,40 @@
 class Claims::Schools::UsersController < ApplicationController
   include Claims::BelongsToSchool
+
+  def index
+    @users = @school.users
+  end
+
+  def show
+    @user = @school.users.find(params.require(:id))
+  end
+
+  def new
+    @user_form = params[:user_invite_form].present? ? user_form : UserInviteForm.new
+  end
+
+  def check
+    render :new unless user_form.valid?
+  end
+
+  def create
+    if user_form.invite
+      redirect_to claims_school_users_path(@school)
+      flash[:success] = t(".user_added")
+    else
+      render :new
+    end
+  end
+
+  private
+
+  def user_params
+    params.require(:user_invite_form)
+          .permit(:first_name, :last_name, :email)
+          .merge({ service: current_service, organisation: @school })
+  end
+
+  def user_form
+    @user_form ||= UserInviteForm.new(user_params)
+  end
 end
--- a/app/controllers/concerns/claims/belongs_to_school.rb
+++ b/app/controllers/concerns/claims/belongs_to_school.rb
@@ -5,7 +5,15 @@ module Claims::BelongsToSchool
     before_action :set_school
 
     def set_school
-      @school = Claims::School.find(params.require(:school_id))
+      @school = scoped_schools.find(params.require(:school_id))
+    rescue ActiveRecord::RecordNotFound
+      redirect_to not_found_path
+    end
+
+    def scoped_schools
+      return Claims::School.all if current_user.support_user?
+
+      current_user.schools
     end
   end
 end
--- a/app/views/claims/schools/users/check.html.erb
+++ b/app/views/claims/schools/users/check.html.erb
@@ -1,0 +1,52 @@
+<%= content_for :page_title, sanitize(t(".page_title", school_name: @school.name)) %>
+
+<%= content_for(:before_content) do %>
+  <%= govuk_back_link(
+    href: new_claims_school_user_path(
+      @user_form.as_form_params.merge(school_id: @school.id),
+    ),
+  ) %>
+<% end %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= form_with(model: @user_form, url: claims_school_users_path, method: :post) do |f| %>
+        <%= f.hidden_field :first_name, value: @user_form.first_name %>
+        <%= f.hidden_field :last_name, value: @user_form.last_name %>
+        <%= f.hidden_field :email, value: @user_form.email %>
+
+        <label class="govuk-label govuk-label--l">
+          <span class="govuk-caption-l"><%= t(".caption", school_name: @school.name) %></span>
+          <%= t(".title") %>
+        </label>
+
+        <%= govuk_summary_list do |summary_list| %>
+          <% summary_list.with_row do |row| %>
+            <% row.with_key(text: User.human_attribute_name("first_name")) %>
+            <% row.with_value(text: @user_form.first_name) %>
+          <% end %>
+          <% summary_list.with_row do |row| %>
+            <% row.with_key(text: User.human_attribute_name("last_name")) %>
+            <% row.with_value(text: @user_form.last_name) %>
+          <% end %>
+          <% summary_list.with_row do |row| %>
+            <% row.with_key(text: User.human_attribute_name("email")) %>
+            <% row.with_value(text: @user_form.email) %>
+          <% end %>
+        <% end %>
+
+        <div class="govuk-warning-text">
+          <span class="govuk-warning-text__icon" aria-hidden="true">!</span>
+          <strong class="govuk-warning-text__text"><%= t(".warning", school_name: @school.name) %></strong>
+        </div>
+
+        <%= f.govuk_submit t(".add_user") %>
+
+        <p class="govuk-body">
+          <%= govuk_link_to(t(".cancel"), claims_school_users_path(@school)) %>
+        </p>
+      <% end %>
+    <div>
+  <div>
+</div>
--- a/app/views/claims/schools/users/index.html.erb
+++ b/app/views/claims/schools/users/index.html.erb
@@ -4,4 +4,29 @@
 
 <div class="govuk-width-container">
   <h1 class="govuk-heading-l"><%= t(".heading") %></h1>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= govuk_button_to(t(".add_user"), new_claims_school_user_path, method: :get) %>
+      <% if @users.any? %>
+        <%= govuk_table do |table| %>
+          <% table.with_head do |head| %>
+            <% head.with_row do |row| %>
+              <% row.with_cell(header: true, text: User.human_attribute_name("full_name")) %>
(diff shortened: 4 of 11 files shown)


app/controllers/claims/schools/users_controller.rb
```diff
@@ -1,3 +1,40 @@
 class Claims::Schools::UsersController < ApplicationController
   include Claims::BelongsToSchool
+
+  def index
+    @users = @school.users
```
JamieCleare2525, line 5:
```suggestion
    @users = @school.users.claims
```

app/controllers/claims/schools/users_controller.rb
```diff
@@ -1,3 +1,40 @@
...
+  end
+
+  def show
+    @user = @school.users.find(params.require(:id))
```
JamieCleare2525, line 9:
```suggestion
    @user = @school.users.claims.find(params.require(:id))
```

app/controllers/claims/schools/users_controller.rb
JamieCleare2525: Might be worth doing something like:
```
def users
  @users ||= @school.users.claims
end

File: spec/system/claims/invite_a_user_to_a_school_spec.rb [hunk 3]
```diff
@@ -41,15 +47,15 @@
 
   scenario "I add a user to different schools" do
```
JamieCleare2525, line 48: Surely, a general user should only be able to add users to their own schools.

JamieCleare2525:
> > Can we add Rspec tests for `user_views_a_list_of_users_in_their_school` and `user_views_another_users_details`?
> 
> Does this test file not test what we need already? `spec/system/claims/invite_a_user_to_a_school_spec.rb` @JamieCleare2525
> 
> It tests details appear ect

It doesn't specifically test that only users in for a specific organisation can be seen by a user in that organisation.
It also doesn't test the links to the show pages from the index page.

I would also expect some form of security to be introduced here. As I would expect general users should only see schools that they are assigned to.

app/controllers/claims/schools/users_controller.rb
Nitemaeric: To add additional context here. The `Claims::School` redefines the `has_many :users` scope to include the `.claims` scope already.

This means, any `Claims::School#users` calls will already scope the users by the claims service.

app/views/claims/schools/users/check.html.erb
```diff
@@ -0,0 +1,91 @@
...
+        <%= f.hidden_field :first_name, value: @user_form.first_name %>
+        <%= f.hidden_field :last_name, value: @user_form.last_name %>
+        <%= f.hidden_field :email, value: @user_form.email %>
+
+        <label class="govuk-label govuk-label--l">
+          <span class="govuk-caption-l"><%= t(".caption", school_name: @school.name) %></span>
+          <%= t(".title") %>
+        </label>
+
+        <dl class="govuk-summary-list">
```
elceebee, line 24: could we place use the govuk_summary_list component here? See [here](https://github.com/DFE-Digital/itt-mentor-services/blob/ec5fcd1c683c9e5373473a93718bb3204e10b851/app/views/placements/support/organisations/users/_check_form.html.erb
)
Better yet, move that component to a shared space and use it directly so we don't have so much repetition.

app/views/claims/schools/users/index.html.erb
```diff
@@ -4,4 +4,30 @@
 
 <div class="govuk-width-container">
   <h1 class="govuk-heading-l"><%= t(".heading") %></h1>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= govuk_button_to(t(".add_user"), new_claims_school_user_path, method: :get) %>
+
+      <% if @users.any? %>
+        <table class="govuk-table">
```
elceebee, line 12: Could we use the govuk_table component? See [here](https://github.com/DFE-Digital/itt-mentor-services/blob/d8b74634014de56998d6cf585466844257435610/app/views/placements/support/schools/users/index.html.erb)
Again, it's now something that is used in many places, so good time to make it a shared component. And in doing so, you'll fix the empty state problem everywhere at once.

app/views/claims/schools/users/show.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <%= govuk_back_link(href: claims_school_users_path(@school)) %>
+<% end %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-l"><%= @school.name %></span>
+      <h2 class="govuk-heading-l"><%= @user.full_name %></h2>
+
+      <dl class="govuk-summary-list">
```
elceebee, line 11: Again, let's use the govuk_summary list component. See [here](https://github.com/DFE-Digital/itt-mentor-services/blob/7758930db2dbabbf75e05d07541e45c75dd78999/app/views/placements/support/organisations/users/_details.html.erb)

And better yet, move that component to a the shared folder and just use it directly so we don't have so much repetition.

elceebee (commented):
I agree with Jamie that we need to check about the requirements -- unless it's a support user, they should only be able to add users to their own school, right?

And would be really nice to reduce some of the repetition through refactoring of reused code. If not, at least use the govuk components to make the code more concise and consistent.

app/controllers/claims/schools/users_controller.rb
CatalinVoineag, line 5: Sorry to open this again, I think we have scoping issue. If you know the ID of a school of a different user you can go to that user's school. I'm not saying we should fix this in this PR as I expect we have similar scoping issues in other places. But to me this feels like where policies come in.

I would expect an index policy here that checks the school_id in the param against the current user schools. Something like:
```
School policy 

def index
  current_user.schools.pluck(:id).inlcude?(params.require(:school_id))
end
```


https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/a0f5a973-363a-43a4-9d09-43e7f2cb9765

Don't want to block this PR, but we should probably address this either here or in a different PR

CatalinVoineag: Thinking about it more, we can also do this scoping in the controller, we just need to handle the user not being allowed to view another school gracefully. Don't mind how we solve this.

Nitemaeric: @CatalinVoineag I like the idea of adding Pundit in, but maybe only when we have resource-based permissions in place. Right now, any user belonging to a school can edit any other user, claim, etc. For now, we can likely get by by scoping down the lookup of school.

@gms-gs 
We could update the `Claims::BelongsToSchool` to scope down the top-level scope to `current_user.schools.find_by(params.require(:id))`. The challenge here is to now make sure it's using the right namespace.

```diff
class Claims::User < User
  default_scope { claims }

+ has_many :schools,
+          through: :memberships,
+          source: :organisation,
+          source_type: "School"
end
```

Adding this in should make sure `current_user.schools` uses the `Claims::School` model as long as `current_user` is of class `Claims::User`. As far as I can tell, the codebase currently does this namespacing.

ollietreend: Please just be aware that `Claims::BelongsToSchool` is also used by the support controllers.

As a rule, support users will need access to _all_ schools. But it seems wasteful & like unnecessary overhead to set up membership records linked every support user with every school.

Instead, I think you'll probably want a `Claims::SupportUser` which redefines the association to `has_many :schools, -> { claims }` so every school is available on `current_user`.

app/controllers/concerns/claims/belongs_to_school.rb
```diff
@@ -5,7 +5,16 @@ module Claims::BelongsToSchool
     before_action :set_school
 
     def set_school
-      @school = Claims::School.find(params.require(:school_id))
+      @school = scoped_schools.find(params.require(:school_id))
+    rescue ActiveRecord::RecordNotFound
+      flash[:alert] = "School not found"
+      redirect_to claims_schools_path
```
Nitemaeric, line 11: Didn't we decide to let this go to a 404 error page? 🤔

config/environments/development.rb
```diff
@@ -12,7 +12,7 @@
   config.eager_load = false
 
   # Show full error reports.
-  config.consider_all_requests_local = true
+  config.consider_all_requests_local = false
```
Nitemaeric, line 15: 🙊

config/locales/en/claims/schools/users.yml
```diff
@@ -4,3 +4,39 @@ en:
       users:
         index:
           heading: Users
+          add_user: Add user
+          no_users: There are no users for %{school_name}.
+          attributes:
+            users:
+              name:  Name
+              email: Email
```
Nitemaeric, line 12: Instead of adding these attributes localisation keys in. Please use `User.human_attribute_name(attribute_name)`. e.g. `User.human_attribute_name(:first_name)` or `User.human_attribute_name(:email)`.

This is create consistency in how we present the attributes names to users.

CatalinVoineag (approved):
Once Dan's comments are addressed I think this is ready to go 🚀

config/environments/development.rb
elceebee: Why are we changing this? Surely we want this to be true for development so we can see more detailed error logs for debugging? 
https://guides.rubyonrails.org/configuring.html#config-consider-all-requests-local

config/locales/en/activerecord.yml
```diff
@@ -18,3 +19,4 @@ en:
               blank: Enter an email address
               taken: This email address is already in use. Try another email address
               invalid: Enter an email address in the correct format, like name@example.com
+              invalid_support_email: Enter a Department for Education email address in the correct format, like name@education.gov.uk
```
Nitemaeric, line 22: i think this is out of scope.

Nitemaeric (approved):
1 thing that slipped in, all else looks good!

PR #152, opened 2024-01-23
Add stimulus framework with ADR

CatalinVoineag:
## Things to check


CatalinVoineag:
[Add stimulus js framework](https://trello.com/c/KbUEubjN/139-add-stimulus-js-framework)

adr/00004-use-stimulus-as-our-javascript-framework.md
```diff
@@ -0,0 +1,23 @@
...
+
+Date: 2024-01-23
+
+## Status
+
+Accepted
+
+## Context
+
+We want to better organise our JavaScript. Stimulus is the default js framework for rails and offers a lot more control then not having a js framework. It allows us to choose which pages or html elements do need a bit JavaScript.
```
Nitemaeric, line 11:
```suggestion
We want to better organise our JavaScript. Stimulus is the default JavaScript framework for Rails and offers a lot more control then not having a JavaScript framework. It allows us to choose which pages or html elements do need a bit JavaScript.
```

adr/00004-use-stimulus-as-our-javascript-framework.md
```diff
@@ -0,0 +1,23 @@
...
+
+Use Stimulus controllers to enhance html elements or pages. This can be done via the stimulus-rails gem.
+- https://stimulus.hotwired.dev/
+- https://github.com/hotwired/stimulus-rails
+
+## Consequences
+
+As with any JavaScript framework or any code, there is a possibility of adding a lot of code and things can get hard to maintain. Luckily our app doesn't require too much JavaScript.
+
+If we broadly follow the rails conventions on name spacing/folder structure, we should be able not have many issues with this, a Stimulus controller can be reused in any view or just one view.
```
Nitemaeric, line 23:
```suggestion
If we broadly follow the Rails conventions on name spacing/folder structure, we should be able not have many issues with this, a Stimulus controller can be reused in any view or just one view.
```

Nitemaeric: I'm not sure if it's best to put the general guidelines for Stimulus controllers here or not, but we should highlight that we want to create behaviour-driven controllers, not context-driven ones.

For example: `autocomplete-controller` not `schools-select-controller`. The former can then be configured to be used in multiple pages.

Nitemaeric (approved):
I know @ollietreend would want us to also update the existing JS ([provider.js](https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/javascript/provider.js), [school.js](https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/javascript/school.js)) into Stimulus, but personally, I'm okay with pushing that into its own PR. We should create a ticket in Trello though.

CatalinVoineag:
> I know @ollietreend would want us to also update the existing JS ([provider.js](https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/javascript/provider.js), [school.js](https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/javascript/school.js)) into Stimulus, but personally, I'm okay with pushing that into its own PR. We should create a ticket in Trello though.

https://trello.com/c/XJu7bKii/140-move-our-existing-javascript-to-stimulus
I don't want to block the claim form

elceebee (approved):
👍

adr/00004-use-stimulus-as-our-javascript-framework.md
CatalinVoineag: I've moved it in the Decision

PR #155, opened 2024-01-23
Lb/placements support navigation

elceebee:
## Context
The current implementation of the `PrimaryNavigation` component only showed a nav item as current if it matched the url. This is not how it is used in the prototype, where it is at times more like a bread crumb

- Add a optional `current` arg to the `PrimaryNavigation` component, so that we can show the nav as current even though the path does not match the `current_url?`
- Add accessible selectors for capaybara. See [here for details](https://github.com/citizensadvice/capybara_accessible_selectors)
- For placements only, I've added the appropriate primary navigation to each support page. 
- I'm *not* doing this for existing claims pages. But the tools / a pattern is there now so they can pick it up when it's a priority. 
- I'm also *not* adding the primary navigation for non-support users in placements. When we start building those pages, we will need to add navigation, page titles, etc etc.

## Guidance to review

- Login as a placements support user. 
- View an organisation. Make sure "Organisations" stays current when you click on 'Details' or 'Users' in the secondary navigation
- Add a user to the organisation. Make sure "Organisation" stays current when you click through the form. 
- Add organisations (schools or providers), make sure "Organisations" stays current as you click through the form. 

Login as a claims user -- make sure everything is how it has always been. (should be no changes for claims users)

## Link to Trello card

https://trello.com/c/v4CUXP3m

## Things to check

## Screenshots
Before
image

After 
image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -96,6 +96,7 @@ end
 group :test do
   # Use system testing [https://guides.rubyonrails.org/testing.html#system-testing]
   gem "capybara"
+  # The accessible selectors gem is maintained by Citizens Advice, not yet a registered gem.
   gem "capybara_accessible_selectors", git: "https://github.com/citizensadvice/capybara_accessible_selectors", branch: "main"
   gem "capybara-screenshot"
   gem "rails-controller-testing"
--- a/Gemfile
+++ b/Gemfile
@@ -96,6 +96,7 @@ end
 group :test do
   # Use system testing [https://guides.rubyonrails.org/testing.html#system-testing]
   gem "capybara"
+  gem "capybara_accessible_selectors", git: "https://github.com/citizensadvice/capybara_accessible_selectors", branch: "main"
   gem "capybara-screenshot"
   gem "rails-controller-testing"
   gem "selenium-webdriver"
--- a/app/components/primary_navigation_component.rb
+++ b/app/components/primary_navigation_component.rb
@@ -2,19 +2,26 @@ class PrimaryNavigationComponent < ApplicationComponent
   renders_many :navigation_items, "NavigationItemComponent"
 
   class NavigationItemComponent < ApplicationComponent
-    attr_reader :name, :url
+    attr_reader :name, :url, :current
 
-    def initialize(name, url, classes: [], html_attributes: {})
+    def initialize(name, url, current: false, classes: [], html_attributes: {})
       @name = name
       @url = url
+      @current = current
 
       super(classes:, html_attributes:)
     end
 
     def call
       content_tag(:li, class: "app-primary-navigation__item") do
-        link_to name, url, class: "app-primary-navigation__link", aria: { current: current_page?(url) && "page" }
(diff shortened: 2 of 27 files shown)


app/views/placements/support/_primary_navigation.html.erb
```diff
@@ -0,0 +1,10 @@
+<%# locals: (current_navigation: nil) -%>
+
+<%= content_for(:primary_navigation_content) do %>
+  <%= render PrimaryNavigationComponent.new do |component| %>
+    <% component.with_navigation_item t(".organisations"),
+                                      support_organisations_path,
+                                      current: current_navigation == :organisations %>
```
Nitemaeric, line 7: There's a potential way to condense this logic.

```suggestion
                                      current: controller_name == "placements/support/organisations" %>
```

```suggestion
                                      current: controller_name.include?("support/organisations") %>
```

Nitemaeric: And if you also need to support the schools and providers controllers:

```suggestion
                                      current: %w[placements/support/organisations placements/support/providers placements/support/schools].include?(controller_name) %>
```

Gemfile
```diff
@@ -96,6 +96,7 @@ end
 group :test do
   # Use system testing [https://guides.rubyonrails.org/testing.html#system-testing]
   gem "capybara"
+  gem "capybara_accessible_selectors", git: "https://github.com/citizensadvice/capybara_accessible_selectors", branch: "main"
```
Nitemaeric, line 99: Could you provide the context on why we need to source the gem from github please?

app/components/primary_navigation_component.rb
```diff
@@ -2,19 +2,26 @@ class PrimaryNavigationComponent < ApplicationComponent
...
     def call
       content_tag(:li, class: "app-primary-navigation__item") do
-        link_to name, url, class: "app-primary-navigation__link", aria: { current: current_page?(url) && "page" }
+        link_to name, url, class: "app-primary-navigation__link", aria: { current: show_as_current?(url) && "page" }
       end
     end
+
+    private
+
+    def show_as_current?(url)
```
Nitemaeric, line 23: Nitpick. I wonder if this needs to be that explicit?

```suggestion
    def current?(url)
```

app/views/placements/support/_primary_navigation.html.erb
Nitemaeric: Not a blocker.

elceebee: I feel like this loses the explicit, easy-to-read nature of what was there before.

Gemfile
elceebee: Because it's maintained by CA, not yet a registered gem. https://github.com/citizensadvice/capybara_accessible_selectors

Nitemaeric: This is likely the only comment holding me back from giving a 👍

Nitemaeric: 🤔 I see.

Nitemaeric (approved):

JamieCleare2525 (approved):

PR #157, opened 2024-01-23
Rename school service attributes

ollietreend:

This PR renames the `claims` and `placements` boolean attributes to `claims_service` and `placements_service`.

This frees up the methods `School#claims` and `School#placements` so they can be used by more meaningful things – for example, to define a School `has_many` Claim association.

## Guidance to review

This is mostly just a 'find & replace' exercise, so there's not too much to see here.

I have made one functional change, though. I've removed the `claims` and `placements` scope from the top-level `School` model. I'd be open to opinions about whether this is a good change or not. I did it mainly to discourage use of the top-level `School` model, since each service has a namespaced `School` model that is already scoped.

## Link to Trello card

https://trello.com/c/2Uw4TwWm/141-rename-claims-and-placements-attributes-on-the-school-model


CatalinVoineag (approved):
Looks good!

app/forms/school_onboarding_form.rb
```diff
@@ -25,7 +25,7 @@ def school_exists?
   end
 
   def school_already_onboarded?
-    if school&.try(service.to_s)
+    if school&.try("#{service}_service?".to_s)
```
JamieCleare2525, line 28: not a blocker but you probably don't need the `to_s`

JamieCleare2525 (approved):

elceebee (commented):
Nice one -- good to spot this now as a problem rather than later when things are too complicated to unpick. 

For consistency, should we also change the bool on providers to follow the same pattern? I know it is just placements, not claims, but it feels right to make it match 
image

Also, apologies for being pedantic, but it would be good to keep the data-model.doc up-to-date with these changes.

elceebee (approved):

app/forms/school_onboarding_form.rb
ollietreend: Done! Thanks for pointing that out.

PR #159, opened 2024-01-25
Dry up common user lists in views

gms-gs:
## Guidance to review

Dry up common user lists in views

## Link to Trello card

https://trello.com/c/BZ2gdAkZ/136-add-empty-state-for-users-list-for-placements-and-claims

## Things to check

### DIFF
--- a/app/views/claims/schools/users/index.html.erb
+++ b/app/views/claims/schools/users/index.html.erb
@@ -8,11 +8,7 @@
     <div class="govuk-grid-column-two-thirds">
       <%= govuk_button_to(t(".add_user"), new_claims_school_user_path, method: :get) %>
 
-      <%= render partial: "shared/users_list", locals: {
-        users: @users,
-        school_or_org: @school,
-        user_show_path: "claims_school_user_path",
-      } %>
+      <%= render "shared/users_list", users: @users, organisation: @school %>
     </div>
   </div>
 </div>
--- a/app/views/claims/schools/users/index.html.erb
+++ b/app/views/claims/schools/users/index.html.erb
@@ -7,26 +7,12 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <%= govuk_button_to(t(".add_user"), new_claims_school_user_path, method: :get) %>
-      <% if @users.any? %>
-        <%= govuk_table do |table| %>
-          <% table.with_head do |head| %>
-            <% head.with_row do |row| %>
-              <% row.with_cell(header: true, text: User.human_attribute_name("full_name")) %>
-              <% row.with_cell(header: true, text: User.human_attribute_name("email")) %>
-            <% end %>
-          <% end %>
-          <% table.with_body do |body| %>
-            <% @users.each do |user| %>
-              <% body.with_row do |row| %>
-                <% row.with_cell(text: govuk_link_to(user.full_name, claims_school_user_path(id: user.id))) %>
-                <% row.with_cell(text: user.email) %>
-              <% end %>
-            <% end %>
-          <% end %>
-        <% end %>
-      <% else %>
-        <p><%= t(".no_users", school_name: @school.name) %></p>
-      <% end %>
+
+      <%= render partial: "shared/users_list", locals: {
+        users: @users,
+        school_or_org: @school,
+        user_show_path: "claims_school_user_path",
+      } %>
     </div>
   </div>
 </div>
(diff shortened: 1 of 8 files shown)


app/views/claims/schools/users/index.html.erb
```diff
@@ -7,26 +7,12 @@
...
-            <% end %>
-          <% end %>
-        <% end %>
-      <% else %>
-        <p><%= t(".no_users", school_name: @school.name) %></p>
-      <% end %>
+
+      <%= render partial: "shared/users_list", locals: {
+        users: @users,
+        school_or_org: @school,
```
Nitemaeric, line 13: Does `organisation` cover both these cases? I believe an Organisation is either a School or a Provider.

app/views/claims/schools/users/index.html.erb
```diff
@@ -7,26 +7,12 @@
...
+        user_show_path: "claims_school_user_path",
```
Nitemaeric, line 14: Also, it might be more consistent to add a helper to `RouteHelpers` that handles the different paths based on different services. `organisations_path` already exists, we can add `organisation_path(organisation)`. Furthermore, we can just use that url helper in the partial itself.

app/views/claims/schools/users/index.html.erb
```diff
@@ -7,26 +7,12 @@
...
-                <% row.with_cell(text: user.email) %>
-              <% end %>
-            <% end %>
-          <% end %>
-        <% end %>
-      <% else %>
-        <p><%= t(".no_users", school_name: @school.name) %></p>
-      <% end %>
+
+      <%= render partial: "shared/users_list", locals: {
```
Nitemaeric, line 11: Nitpick.
```suggestion
      <%= render "shared/users_list", users: @users, organisation: @school %>
```

config/locales/en.yml
```diff
@@ -1,6 +1,7 @@
 en:
   pagination_info: Showing %{from} to %{to} of %{count} results
   no_results: There are no results for '%{for}'.
+  no_users: There are no users for %{school_name}.
```
Nitemaeric, line 4:
```suggestion
  no_users: There are no users for %{organisation_name}.
```

app/views/placements/support/providers/users/index.html.erb
```diff
@@ -12,22 +12,12 @@
...
-        <% table.with_body do |body| %>
-          <% @users.each do |user| %>
-            <% body.with_row do |row| %>
-              <% row.with_cell(text: govuk_link_to(user.full_name, placements_support_provider_user_path(id: user.id))) %>
-              <% row.with_cell(text: user.email) %>
-            <% end %>
-          <% end %>
-        <% end %>
-      <% end %>
+      <%= render partial: "shared/users_list", locals: {
```
elceebee, line 15: Not a blocker, but can you delete the unused translations (`.attributes.users.email`)?

elceebee (approved):
No blockers. Very nice to have this tidied up, thanks!
- I agree with Dan that we should just use `organisation` for provider or school rather than `school_or_org`
- Would be nice to delete the unused translations so we can keep those files clean.

app/views/shared/_users_list.html.erb
```diff
@@ -0,0 +1,20 @@
+<% if users.any? %>
+  <%= govuk_table do |table| %>
+    <% table.with_head do |head| %>
+      <% head.with_row do |row| %>
+        <% row.with_cell(header: true, text: User.human_attribute_name("full_name")) %>
+        <% row.with_cell(header: true, text: User.human_attribute_name("email")) %>
```
Nitemaeric, line 6: Tiny thing, but a symbol is better suited for attribute names.

```suggestion
        <% row.with_cell(header: true, text: User.human_attribute_name(:full_name)) %>
        <% row.with_cell(header: true, text: User.human_attribute_name(:email)) %>
```

Nitemaeric:
After some discussion, we're decided to drop this piece of work. I will create a ticket to use the govuk components instead of the HTML after I've submitted this comment.

For context, we've analysed the needs and conceptual differences between the different pages and we've determined they are different. We currently have 4 conceptually different user lists:

1. List of School Users from a **User**'s perspective within the **Claims** service
2. List of School Users from a **Support User**'s perspective within the **Claims** service
3. List of Organisation Users from a **User**'s perspective within the **Placements** service
4. List of Organisation Users from a **Support User**'s perspective within the **Placements** service

If you look at the files changed, you'll see that the 4 pages affected map to 3 of the 4 cases above. Only (3) being missing as the user-facing pages are in development.

There is currently no contract binding the two services together to present data in the same way. As they are different services servicing different needs, there is a likelihood that the pages need to surface different data.

Moving forward, we should be mindful about sharing Views. It's great to share View Components, View Helpers, Controller Concerns, our data layer, and ultimately our best practices and learnings, but sharing any Controllers and Views couples our service too tight together.

cc: @DFE-Digital/school-placements

PR #161, opened 2024-01-25
Basic smoke tests

Nitemaeric:
## Context

We want to be able to perform smoke tests on deployed environments. This gives us confidence that we haven't broken core functionality when deploying.

Unfortunately, all of our core functionality is behind DfE Sign In. This limits how effective a minimal smoke test can be for our services.

We should review how we want to improve our smoke tests later.

- Add a way to configure the targeted service for system and request specs
- Add a way to perform smoke tests via Capybara

## Guidance to review

- Smoke tests are currently configured to make requests on the `http://claims.localhost:3000` and `http://placements.localhost:3000` hosts.
- Ensure your development server is running and runs the smoke tests with `bundle exec rspec --tag smoke_test`. You should see access logs.
- Shut down the server and re-run the smoke tests. The tests should fail.

## Things to check


spec/system/home_page_spec.rb
```diff
@@ -1,33 +0,0 @@
-require "rails_helper"
-
-RSpec.describe "Home Page", type: :system do
-  scenario "User visits the claims homepage" do
-    given_i_am_on_the_claims_site
```
ollietreend, line 5: You've introduced `service:` metadata as a new way to specify which service the test runs against. So should we remove these helper methods now?

https://github.com/DFE-Digital/itt-mentor-services/blob/a691cfdb0ba7ff8367606f4598a855ee8deb9c2a/spec/support/services_test_helper.rb#L1-L9

Otherwise we'll have two ways to do it, which is a bit ambiguous.

.github/workflows/deploy.yml
```diff
@@ -96,7 +96,7 @@ jobs:
         run: |
           bundle exec rake db:create
           bundle exec rails db:schema:load
-          bundle exec rspec --format progress
+          bundle exec rspec --format progress --exclude-pattern "spec/smoke_tests/**/*_spec.rb"
```
ollietreend, line 99: Does this mean the smoke tests will execute when we run RSpec locally via:

```
bundle exec rspec
```

?

Since these are a different class of test - based on the assumption that the server is up & running - I wonder if that makes our DX a little bit worse.

Perhaps we could do [something like this](https://stackoverflow.com/a/53661752) so that smoke tests are excluded by default, and then a different command has to be used to specifically run them?

Nitemaeric: Nice! I like it!

ollietreend (approved):
Nice! 🤩 and thanks for making those changes.

Nitemaeric:
[Add minimal smoke test](https://trello.com/c/woI13Aj3/146-add-minimal-smoke-test)

PR #163, opened 2024-01-25
Refactor autocomplete javascript into stimulus

CatalinVoineag:
## Context

Previously we had two autocompletes, one for schools and one for
providers. These were also not using Stimulus.

This PR ports these two into one autocomplete stimulus controller
which should work with any model.

A refactor has been done to the provider form and school form that
allows the user to add a school or a provider.

These were using `urn` and `code` to search a school or a provider in DB, this has
been changed to `id`, to comply with the stimulus autocomplete.

I have also changed the form input name of the school and provider from
`urn` and `code` to just `id`

- Stimulus autocomplete
- SchoolOnbordingForm
- ProviderOnboardingForm
- Specs

## Guidance to review
- Add a school in claims and placements
- Add a provider in placements
- Do these steps without javascript as well 😄 
- Everything should work as before.

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/d4c2d8e4-cf3a-4168-8ce4-25d916468aac


app/controllers/claims/support/schools_controller.rb
```diff
@@ -24,7 +24,9 @@ def school_options
 
   def check_school_option
     if school_form(javascript_disabled: true).valid?
-      redirect_to check_claims_support_schools_path(school: { search_urn: urn_param })
+      redirect_to check_claims_support_schools_path(
+        school: { id: school_params.fetch(:id), name: search_param },
```
CatalinVoineag, line 28: We need to trick the `javascript_disabled?` method here to get to the `check` method

CatalinVoineag:
[Move our existing javascript to Stimulus](https://trello.com/c/XJu7bKii/140-move-our-existing-javascript-to-stimulus)

elceebee (changes requested):
This looks fantastic. Epic piece of work. Two bugs:
1. the stuff in brackets shouldn't have nulls
image
2. We have lost the functionality where if you click on the href in the error summary, it takes the user to the input.

CatalinVoineag:
> This looks fantastic. Epic piece of work. Two bugs:
> 
> 1. the stuff in brackets shouldn't have nulls
> 
> image 2. We have lost the functionality where if you click on the href in the error summary, it takes the user to the input.

Ah, I think I've not ported all the functionality from the old javascript. Nice catch

elceebee (approved):
Tested locally and the bugs have been fixed! Great work!

app/forms/provider_onboarding_form.rb
```diff
@@ -12,26 +12,26 @@ def onboard
   end
 
   def provider
-    @provider ||= Provider.find_by(code:)
+    @provider ||= Provider.find_by_id(id)
```
JamieCleare2525, line 15: I was under the impression `find_by_attributes` methods we deprecated back in Rails 5. Shouldn't this be `find(id)`, as you don't need to specify `id` and `id` should be present.

app/forms/school_onboarding_form.rb
```diff
@@ -13,26 +13,26 @@ def onboard
   end
 
   def school
-    @school ||= School.find_by(urn:)
+    @school ||= School.find_by_id(id)
```
JamieCleare2525, line 16: Same as above.

app/forms/provider_onboarding_form.rb
CatalinVoineag: I kept the find_by pattern because we expect a nil or a record here https://github.com/DFE-Digital/itt-mentor-services/blob/92461ea900f3cc52f89749ecbcf03e0c1530d3e9/app/forms/provider_onboarding_form.rb#L21

`find` would return us an error if we can't find it. I can just rescue a `ActiveRecord::RecordNotFound` and put the message in there if you want

CatalinVoineag: Fixed 9bc10093ea8ca35cf0f036cd52a556352a202e45

app/forms/school_onboarding_form.rb
CatalinVoineag: Fixed 9bc10093ea8ca35cf0f036cd52a556352a202e45

JamieCleare2525 (approved):
Tested locally. Works wonderfully. Great work!

PR #164, opened 2024-01-25
Ot/jc/lb/scope user memberships to service for organisations

elceebee:
## Context

In our controllers, we are currently having to do a bit of gymnastics to ensure we are limiting our organisations to those that not only belong to a particular service user, but that are organisations with the service == true. Eg:

`current_user.schools.where(placements_service: true).first.becomes(Placements::School)`

after this pr, you can just call:`current_user.schools.first` it will be a `Placements::School` if the current_user is a `Placements::User`. It will be a `Claims::School` if the current_user is a `Claims::User`

Moved the association to the child classes. 
Added back scopes to parent models since we are using them in multiple places

## Guidance to review

You shouldn't need to reseed the database. 
Mess about with it on the command line.
None of the functionality should change, so just click around and make sure everything is as it should be. 

## Link to Trello card

https://trello.com/c/sHJqWuJ9

## Things to check

## Screenshots

image

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -45,6 +45,7 @@ class Provider < ApplicationRecord
   validates :code, uniqueness: { case_sensitive: false }
 
   scope :accredited, -> { where accredited: true }
+  scope :placements_service, -> { where placements_service: true }
 
   multisearchable against: %i[name postcode],
                   if: :placements_service?,
--- a/app/models/school.rb
+++ b/app/models/school.rb
@@ -48,12 +48,14 @@ class School < ApplicationRecord
   include PgSearch::Model
 
   has_many :memberships, as: :organisation
-  has_many :users, through: :memberships
   has_many :mentors
 
   validates :urn, presence: true
   validates :urn, uniqueness: { case_sensitive: false }
   validates :name, presence: true
+
+  scope :placements_service, -> { where placements_service: true }
+  scope :claims_service, -> { where claims_service: true }
 
   multisearchable against: %i[name postcode],
                   if: :placements_service?,
--- a/spec/models/placements/user_spec.rb
+++ b/spec/models/placements/user_spec.rb
@@ -18,6 +18,16 @@
 require "rails_helper"
 
 RSpec.describe Placements::User do
+  describe "associations" do
+    it { should have_many(:schools).through(:memberships).source(:organisation) }
+
+    it "association returns Placements::Schools objects" do
+      placements_user = create(:placements_user)
+      placements_user.memberships.create!(organisation: create(:school, :placements))
+      expect(placements_user.schools.first.class.name).to eq "Placements::School"
+    end
+  end
+
   describe "default scope" do
     let(:email) { "same_email@email.co.uk" }
     let!(:user_with_placements_service) { create(:placements_user, email:) }
--- a/spec/system/claims/view_organisations_spec.rb
+++ b/spec/system/claims/view_organisations_spec.rb
@@ -12,7 +12,7 @@ RSpec.describe "View schools", type: :system do
     and_persona_has_multiple_schools(persona)
     when_i_visit_claims_personas
     when_i_click_sign_in(persona)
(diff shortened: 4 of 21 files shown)


JamieCleare2525 (approved):
Tested locally.

CatalinVoineag (approved):
Very nice work!

app/models/school.rb
```diff
@@ -48,13 +48,15 @@ class School < ApplicationRecord
...
   has_many :memberships, as: :organisation
-  has_many :users, through: :memberships
   has_many :mentors
 
   validates :urn, presence: true
   validates :urn, uniqueness: { case_sensitive: false }
   validates :name, presence: true
 
+  scope :placements_service, -> { where placements_service: true }
+  scope :claims_service, -> { where claims_service: true }
```
ollietreend, line 58: It might be a good idea to re-introduce the scope tests that I removed in #157.

https://github.com/DFE-Digital/itt-mentor-services/blob/2e0369e234ed49b8ffb7a18c233284b46bde6cac/spec/models/school_spec.rb#L62-L78

I realise that we do still [implicitly test this functionality](https://github.com/DFE-Digital/itt-mentor-services/blob/f8c00ef494d491f36c47f754816a41a80afd1686/spec/models/placements/school_spec.rb#L50-L58) on the namespaced classes. So this isn't technically un-covered. But would be nice to have an explicit test since it's part of `School`'s public interface.

app/models/provider.rb
```diff
@@ -45,6 +45,7 @@ class Provider < ApplicationRecord
   validates :code, uniqueness: { case_sensitive: false }
 
   scope :accredited, -> { where accredited: true }
+  scope :placements_service, -> { where placements_service: true }
```
ollietreend, line 48: Ditto

https://github.com/DFE-Digital/itt-mentor-services/blob/2e0369e234ed49b8ffb7a18c233284b46bde6cac/spec/models/provider_spec.rb#L53-L60

spec/models/placements/user_spec.rb
```diff
@@ -18,6 +18,16 @@
 require "rails_helper"
 
 RSpec.describe Placements::User do
+  describe "associations" do
+    it { should have_many(:schools).through(:memberships).source(:organisation) }
+
+    it "association returns Claims::Schools objects" do
```
ollietreend, line 24:
```suggestion
    it "association returns Placements::Schools objects" do
```

spec/system/claims/view_organisations_spec.rb
```diff
@@ -57,10 +57,11 @@ def i_am_redirected_to_the_root_path
   end
 
   def i_am_redirected_to_organisation_index(persona)
-    expected_content = persona.schools.many? ? "Organisations" : "Organisation details"
+    user = persona.becomes(User).becomes(Claims::User)
```
ollietreend, line 60: This `.becomes().becomes()` feels a bit weird. But I'm expecting this will probably go away or change once we've [set up STI on the User model](https://trello.com/c/IB7IpN2x/148-refactor-user-models-to-use-single-table-inheritance).

ollietreend (changes requested):
Really nice work! This is going to make the data model much simpler to work with. ✨

Just a couple of small things which I think we might want to change.

app/models/school.rb
elceebee: Doh, of course!

spec/system/claims/view_organisations_spec.rb
elceebee: Yeah, it it isn't pretty. On the placements side, I removed the amiguity and just checked for specific words on the page, which I think is a better way to test the front end more generally. I just got lazy here...

spec/models/placements/user_spec.rb
elceebee: Doh

ollietreend (approved):
LGTM! 🚀

PR #166, opened 2024-01-26
Align sign-in flows for DfE Sign In and Persona

Nitemaeric:
## Context

Right now, we skip a step from our intended production flow by redirecting to the `/personas` page. If we are to treat persona selection as an OmniAuth method, we should mirror the sign-in flows.

- Stop redirecting to `/personas` automatically.
- Skip authentication when rendering error pages.

## Guidance to review

- You should be able to access the `/sign-in` path.

## Screenshots

| DfE Sign In | Persona |
| ----------- | -------- |
| CleanShot 2024-01-26 at 13 38 37 | CleanShot 2024-01-26 at 13 39 04 |

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ErrorsController < ApplicationController
-  skip_before_action :verify_authenticity_token
+  skip_before_action :verify_authenticity_token, :authenticate_user!
 
   def not_found
     render "not_found", status: :not_found
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,6 +1,5 @@
 class SessionsController < ApplicationController
   skip_before_action :authenticate_user!, only: %i[new callback]
-  before_action :redirect_to_personas, only: [:new]
 
   # setup inspired by register-trainee-teacher
   # TODO: Replace with commented code when DfE Sign In implemented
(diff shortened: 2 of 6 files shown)


CatalinVoineag (approved):
Works on local for personas

JamieCleare2525 (commented):
I can't see a test that checks this behaviour for Placements. As this change will affect both services, could you add/amend a test to check the behaviour.

Nitemaeric:
> I can't see a test that checks this behaviour for Placements. As this change will affect both services, could you add/amend a test to check the behaviour.

Most of the tests visit the `personas_path` directly. The tests don't attempt to land on the homepage and navigate from there.

I had 1 system spec that I wrote that walked through the flow from landing on the root url.

JamieCleare2525:
> > I can't see a test that checks this behaviour for Placements. As this change will affect both services, could you add/amend a test to check the behaviour.
> 
> Most of the tests visit the `personas_path` directly. The tests don't attempt to land on the homepage and navigate from there.
> 
> I had 1 system spec that I wrote that walked through the flow from landing on the root url.

Sorry Dan, what I'm asking is could you add a spec to test this functionality for the Placements service as well. 
Just a small one to say
- given_i_start_on_the_root_page
- when_i_click_sign_in_via_personas
- then_i_can_see_a_list_of_persona_options

Nitemaeric:
@JamieCleare2525 I've added an example of walking through the flow to one of your system specs. I didn't explicitly test any of the personas page behaviour.

My upcoming goal is to pivot away from using the personas page to perform system specs. Hopefully with the work @CatalinVoineag is working on to get DfE Sign In ready, we should be able to mock DfE Sign In calls.

PR #171, opened 2024-01-27
As a Support User, I should be able to see all Claims made by a School

gms-gs:
## Context

This change allows a support user to view a list of all claims made by a school 

**Note: We don't have an example of this page in the prototype. Once updated I can update this PR..**

Added content to the claims page

## Guidance to review

- Sign in as a support user
- Go to a school
- Click on the secondary navigation Claims link

## Link to Trello card

https://trello.com/c/ftOHgLbl/100-as-a-support-user-i-should-be-able-to-see-all-claims-made-by-a-school

## Things to check

## Screenshots

Screenshot 2024-01-27 at 17 05 14

### DIFF
--- a/app/components/claim/status_tag_component.rb
+++ b/app/components/claim/status_tag_component.rb
@@ -1,0 +1,16 @@
+class Claim::StatusTagComponent < ApplicationComponent
+  attr_reader :claim
+
+  def initialize(claim:, classes: [], html_attributes: {})
+    super(classes:, html_attributes:)
+
+    @claim = claim
+  end
+
+  def call
+    css_class = claim.draft ? "govuk-tag--grey" : "govuk-tag--blue"
+    text = claim.draft ? t(".draft") : t(".submitted")
+
+    content_tag(:p, text, class: "govuk-tag #{css_class}")
+  end
+end
--- a/spec/system/claims/support/schools/claims/view_a_schools_claims_spec.rb
+++ b/spec/system/claims/support/schools/claims/view_a_schools_claims_spec.rb
@@ -1,0 +1,46 @@
+require "rails_helper"
+
+RSpec.describe "View a schools claims", type: :system, service: :claims do
+  let!(:school) { create(:school, :claims).becomes(Claims::School) }
+  let!(:another_school) { create(:school, :claims).becomes(Claims::School) }
+
+  let!(:colin) { create(:persona, :colin, service: "claims") }
+
+  let!(:submitted_claim) { create(:claim, school_id: school.id, draft: false) }
+  let!(:draft_claim) { create(:claim, school_id: school.id, draft: true) }
+  let!(:claim_from_another_school) { create(:claim, school_id: another_school.id, draft: true) }
+
+  scenario "View a school's claims as a support user" do
+    given_i_sign_in_as_colin
+    when_i_visit_the_claims_support_school_claims_page
+    i_see_a_list_of_the_schools_claims
+    i_dont_see_claims_from_other_schools
+  end
+
+  private
+
+  def given_i_sign_in_as_colin
+    and_i_visit_the_personas_page
+    and_i_click_sign_in_as("Colin")
+  end
+
+  def and_i_visit_the_personas_page
+    visit personas_path
+  end
+
+  def and_i_click_sign_in_as(persona_name)
+    click_on "Sign In as #{persona_name}"
+  end
+
+  def when_i_visit_the_claims_support_school_claims_page
+    visit claims_support_school_claims_path(school)
+  end
+
+  def i_see_a_list_of_the_schools_claims
+    expect(page).to have_content("#{draft_claim.id}\nDraft\n#{submitted_claim.id}\nSubmitted")
+  end
+
+  def i_dont_see_claims_from_other_schools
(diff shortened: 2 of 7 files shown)


app/helpers/claims/support/claims_helper.rb
```diff
@@ -0,0 +1,8 @@
+module Claims::Support::ClaimsHelper
+  def claim_draft_state(claim)
+    css_class = claim.draft ? "govuk-tag--grey" : "govuk-tag--blue"
+    text = claim.draft ? "Draft" : "Submitted"
+
+    "<strong class='govuk-tag #{css_class}'>#{text}</strong>".html_safe
```
CatalinVoineag, line 6: Do you think we can use `contet_tag` for this html? You shouldn't need to call html_safe with a content tag I think https://apidock.com/rails/ActionView/Helpers/TagHelper/content_tag

app/helpers/claims/support/claims_helper.rb
```diff
@@ -0,0 +1,8 @@
+module Claims::Support::ClaimsHelper
+  def claim_draft_state(claim)
+    css_class = claim.draft ? "govuk-tag--grey" : "govuk-tag--blue"
+    text = claim.draft ? "Draft" : "Submitted"
```
CatalinVoineag, line 4: Can we put the `Draft` and `Submitted` into translations

CatalinVoineag (commented):
This is almost ready to go, just minor things

app/helpers/claims/support/claims_helper.rb
```diff
@@ -0,0 +1,8 @@
+module Claims::Support::ClaimsHelper
```
JamieCleare2525, line 1: I'm a big believer in that model specific logic, like this, should go into a decorator. Or alternatively a ViewComponent. 
I would also consider renaming this method to `state_tag`, as this doesn't just return something if the claim is in a draft start.

spec/system/claims/support/schools/claims/view_a_schools_claims_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  def and_i_click_sign_in_as(persona_name)
+    click_on "Sign In as #{persona_name}"
+  end
+
+  def when_i_visit_the_claims_support_school_claims_page
+    visit claims_support_school_claims_path(school)
+  end
+
+  def i_see_a_list_of_the_schools_claims
+    expect(page).to have_content("Claims\nID #{claim.id} Submitted")
```
JamieCleare2525, line 34: Could you also add a test to show that a "Draft" claim shows as "Draft" and possibly add a random claim (_not associated with that school)_ to demonstrate that only school related claims are returned.

CatalinVoineag (approved):
Beautiful, thank you!

app/components/state_tag_component.rb
```diff
@@ -0,0 +1,16 @@
+class StateTagComponent < ApplicationComponent
+  attr_reader :claim
+
+  def initialize(claim:, classes: [], html_attributes: {})
+    super(classes:, html_attributes:)
+
+    @claim = claim
+  end
+end
```
Nitemaeric, line 9: Following the [guidance from GitHub themselves](https://viewcomponent.org/viewcomponents-at-github.html#app-specific-viewcomponents), we should namespace this component if it is dependent on the model.

```ruby
Claim::StateTagComponent
```

Also, let's name it `StatusTagComponent` instead of `State` as other BAT apps tend to use `status` rather than `state`.

app/components/state_tag_component.rb
```diff
@@ -0,0 +1,16 @@
...
+
+def state_tag
```
Nitemaeric, line 11: Also, if this doesn't really require a template, you can name this method `call` and it will render `call` instead of the template.

Nitemaeric: In that case, would it make more sense to bring this render logic to the template file itself?

app/components/claim/status_tag_component.rb
```diff
@@ -0,0 +1,16 @@
...
+  def initialize(claim:, classes: [], html_attributes: {})
+    super(classes:, html_attributes:)
+
+    @claim = claim
+  end
+end
+
+def state_tag
+  css_class = claim.draft ? "govuk-tag--grey" : "govuk-tag--blue"
+  text = claim.draft ? I18n.t("components.state_tag_component.draft") : I18n.t("components.state_tag_component.submitted")
```
Nitemaeric, line 13: Relative locale keys should work here :)

```suggestion
  text = claim.draft ? I18n.t(".draft") : I18n.t(".submitted")
```

JamieCleare2525 (approved):

app/components/claim/status_tag_component.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def initialize(claim:, classes: [], html_attributes: {})
+    super(classes:, html_attributes:)
+
+    @claim = claim
+  end
+
+  def call
+    css_class = claim.draft ? "govuk-tag--grey" : "govuk-tag--blue"
+    text = claim.draft ? I18n.t(".draft") : I18n.t(".submitted")
```
Nitemaeric, line 12: Ah, I know why....

```suggestion
    text = claim.draft ? t(".draft") : t(".submitted")
```

PR #173, opened 2024-01-29
Add SupportUser class and use implicit association class casting

Nitemaeric:
## Context

As [STI has been adopted for the User class](https://github.com/DFE-Digital/itt-mentor-services/pull/168), we can now fully lean into the practice of using only namespaced models and with Rails' _magic_ ✨, implicit namespaces for associations.

- Drop `.claims` and `.placements` scopes from User. A User can be one or the other, never both, therefore we can rely on calling `Claims::User` and `Placements::User` for `User.claims` and `User.placements`.

## Guidance to review

  ```ruby
  class Claims::User < User
    has_many :schools # This will use the current namespace and look for Claims::School by default.
  end
  ```

- Try going through flows. Nothing should break.
- Try the following in your console:

  ```ruby
  # To silence SQL query logs
  ApplicationRecord.logger.silence do
    [
      Claims::User.all.flat_map(&:schools).all? { _1.class == Claims::School },
      Claims::School.all.flat_map(&:users).all? { _1.class == Claims::User },
      Placements::User.all.flat_map(&:schools).all? { _1.class == Placements::School },
      Placements::User.all.flat_map(&:providers).all? { _1.class == Placements::Provider },
      Placements::School.all.flat_map(&:users).all? { _1.class == Placements::User },
      Placements::Provider.all.flat_map(&:users).all? { _1.class == Placements::User },
    ]
  end
  ```

-----

Whilst trying to fix the `config/initializers/preload_stis.rb` code running without a database, I stumbled on optimising our GitHub workflow. Lint is around 50% faster without setting up Node and more.

| Before | After |
| ------ | ----- |
| CleanShot 2024-01-29 at 15 30 05 | CleanShot 2024-01-29 at 15 31 36 |

### DIFF
--- a/app/models/claims/school.rb
+++ b/app/models/claims/school.rb
@@ -47,6 +47,6 @@
 class Claims::School < School
   default_scope { claims_service }
 
-  has_many :users, -> { claims }, through: :memberships
+  has_many :users, through: :memberships
   has_many :claims
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -33,4 +33,12 @@ class User < ApplicationRecord
   def support_user?
     false
   end
+
+  # Extracts the namespace of the class as a symbol to determine the service
+  # e.g.
+  # - Placements::User => :placements
+  # - Claims::User => :claims
+  def service
+    self.class.name.deconstantize.downcase.to_sym
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -23,9 +23,6 @@ class User < ApplicationRecord
   validates :type, presence: true
   validates :email, uniqueness: { scope: :type, case_sensitive: false }
 
-  scope :claims, -> { where(type: "Claims::User") }
-  scope :placements, -> { where(type: "Placements::User") }
-
   def full_name
     "#{first_name} #{last_name}".strip
   end
--- a/spec/models/claims/school_spec.rb
+++ b/spec/models/claims/school_spec.rb
@@ -49,6 +49,21 @@ require "rails_helper"
 RSpec.describe Claims::School do
   context "associations" do
     it { should have_many(:claims) }
+
+    context "#users" do
+      it { should have_many(:users).through(:memberships) }
+
+      it "returns only Claims::User records" do
+        claims_school = create(:claims_school)
+        claims_user = create(:claims_user)
+
+        claims_school.users << claims_user
+        claims_school.users << create(:placements_user)
+
+        expect(claims_school.users).to contain_exactly(claims_user)
+        expect(claims_school.users).to all(be_a(Claims::User))
+      end
+    end
   end
 
   describe "default scope" do
--- a/.env.test
+++ b/.env.test
@@ -1,3 +1,5 @@
 PLACEMENTS_HOST=placements.localhost
 CLAIMS_HOST=claims.localhost
+SIGN_IN_METHOD=persona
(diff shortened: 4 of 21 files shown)


config/initializers/preload_stis.rb
```diff
@@ -0,0 +1,10 @@
+# https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#option-4-preload-types-from-the-database
+
+if Rails.env.development? && Rails.application.config.eager_load
+  Rails.application.config.to_prepare do
+    ApplicationRecord.logger.silence do
+      types = User.unscoped.select(:type).distinct.pluck(:type)
+      types.compact.each(&:constantize)
+    end
+  end
+end
```
ollietreend, line 10: I'm not sure why this is needed. I've read the linked docs, but I'm still not sure what the need is for preloading the different user classes.

When I check out the branch locally and run this in the Rails console:

```ruby
User.first.class.name
  User Load (0.3ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT $1  [["LIMIT", 1]]
=> "Claims::SupportUser"
```

I get the expected result, and no weird log output, regardless of whether I have this initializer or not. It seems to work the same either way? 🤔 

Can you ELI5? 🙏🏻

app/models/user.rb
```diff
@@ -33,4 +33,8 @@ def full_name
   def support_user?
     false
   end
+
+  def service
+    self.class.name.split("::").first.downcase.to_sym
+  end
```
ollietreend, line 36: A comment could be nice here, because this line seems quite difficult to parse:

```suggestion
  def service
    # Infer service from class namespace – e.g. Placements::User becomes :placements
    self.class.name.split("::").first.downcase.to_sym
  end
```

app/models/claims/school.rb
```diff
@@ -47,6 +47,6 @@
 class Claims::School < School
   default_scope { claims_service }
 
-  has_many :users, -> { claims }, through: :memberships
+  has_many :users, through: :memberships
```
ollietreend, line 50: Nice!

spec/models/claims/school_spec.rb
```diff
@@ -49,6 +49,21 @@
...
+        claims_school = create(:claims_school)
+        claims_user = create(:claims_user)
+
+        claims_school.users << claims_user
+        claims_school.users << create(:placements_user)
+
+        expect(claims_school.users).to contain_exactly(claims_user)
+        expect(claims_school.users).to all(be_a(Claims::User))
+      end
+    end
```
ollietreend, line 66: Nice test coverage, too.

ollietreend (commented):
Really nice stuff 🎉

I think we've really hit upon something good with #164, #168 and now this PR.

Just one question from me, and one comment.

config/initializers/preload_stis.rb
Nitemaeric: I messed up inverting the latter condition... It should be `if Rails.env.development? && !Rails.application.config.eager_load`

Anyway...

Try running `SupportUser.all` without this code. It will default to `where(type: "SupportUser")` in development as it isn't aware of the additional subclasses that inherit from `SupportUser`.

app/models/user.rb
Nitemaeric: Updated. Added examples and method documentation using standard RDoc/YARD documentation format.

ollietreend (approved):
Nice work ✨ 🚀

PR #176, opened 2024-01-29
Update tests to use updated way of creating users

gms-gs:
## Context

Update test to use correct way of creating users

## Guidance to review

Tests pass

## Things to check


elceebee (approved):
Cool, thanks.

For future reference, you can use `create(:claims_school)` rather than using `becomes(Claims::School)`. Another one of the advantages of the STI changes.

PR #184, opened 2024-01-30
Add DFE sign in integration for login 

CatalinVoineag:
## Context

We need to allow our users to login into our service using DFE sign in.

This PR adds the configuration to accommodate this, we find the user by the `dfe_sign_in_uid` or `email`. This should work for placements and claims through env variables. And for different env, production, staging, local etc.

We need to create users with dfe sign in emails so that they can login in our app using dfe sign in. This can be done by a support persona now.

There is no system test for the dfe sign in path, this is because we need to refactor most of our tests as they use persona sign in, they cannot work together because the omniauth config loads only once

Dfe sign-in works with localhost even if their documentation says it doesn't 😄. There is a bit of service configuration that needs to happen in here https://dev-manage.signin.education.gov.uk, mainly the url of the app, the redirect path and the log out path

- Added [awesome_print](https://github.com/awesome-print/awesome_print) Helps me a lot when debugging
- Migrations to user table, `dfe sign in uid and last_signed_at`
- omniauth.rb
- Sessions controller
- DFE sign in user file
- Update user after sign in service
- Updated the Notice banner to allow other messages beside success ones

## Guidance to review

- Have a dfe sign in account
- Login as a support user using persona sign in, `SIGN_IN_METHOD=persona`
- Create a new user in an organisation with your dfe sign in email, `@education.gov.uk`
- Restart the app with `SIGN_IN_METHOD=dfe-sign-in`
- Login with the new user created in our app
- You should be successfully logged in

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/1f1751f6-0c7d-4192-84ad-0161e28fc7a9

### DIFF
--- a/app/models/dfe_sign_in_user.rb
+++ b/app/models/dfe_sign_in_user.rb
@@ -1,30 +1,26 @@
 class DfESignInUser
-  attr_reader :email, :dfe_sign_in_uid
+  attr_reader :email, :dfe_sign_in_uid, :id_token, :provider
   attr_accessor :first_name, :last_name, :service
 
-  # setup inspired by register-trainee-teacher
-  # TODO: Replace with commented code when DfE Sign In implemented
-
-  # def initialize(email:, dfe_sign_in_uid:, first_name:, last_name:, id_token: nil, provider: "dfe", service:)
-  def initialize(email:, first_name:, last_name:, service:)
+  def initialize(email:, dfe_sign_in_uid:, first_name:, last_name:, service:, id_token: nil, provider: nil)
     @email = email&.downcase
-    # @dfe_sign_in_uid = dfe_sign_in_uid
+    @dfe_sign_in_uid = dfe_sign_in_uid
     @first_name = first_name
     @last_name = last_name
-    # @id_token = id_token
-    # @provider = provider&.to_s
+    @id_token = id_token
+    @provider = provider&.to_s
     @service = service
   end
 
   def self.begin_session!(session, omniauth_payload)
     session["dfe_sign_in_user"] = {
       "email" => omniauth_payload["info"]["email"],
-      # "dfe_sign_in_uid" => omniauth_payload["uid"],
+      "dfe_sign_in_uid" => omniauth_payload["uid"],
       "first_name" => omniauth_payload["info"]["first_name"],
       "last_name" => omniauth_payload["info"]["last_name"],
-      # "last_active_at" => Time.zone.now,
-      # "id_token" => omniauth_payload["credentials"]["id_token"],
-      # "provider" => omniauth_payload["provider"],
+      "last_active_at" => Time.current,
+      "id_token" => omniauth_payload["credentials"]["id_token"],
+      "provider" => omniauth_payload["provider"],
     }
   end
 
@@ -32,20 +28,27 @@ class DfESignInUser
     dfe_sign_in_session = session["dfe_sign_in_user"]
     return unless dfe_sign_in_session
 
+    # Users who signed in before session expiry was implemented will not have
+    # `last_active_at` set. In that case, force them to sign in again.
+    return unless dfe_sign_in_session["last_active_at"]
+
+    return if dfe_sign_in_session.fetch("last_active_at") < 2.hours.ago
+
+    dfe_sign_in_session["last_active_at"] = Time.current
+
     new(
       email: dfe_sign_in_session["email"],
-      # dfe_sign_in_uid: dfe_sign_in_session["dfe_sign_in_uid"],
+      dfe_sign_in_uid: dfe_sign_in_session["dfe_sign_in_uid"],
       first_name: dfe_sign_in_session["first_name"],
       last_name: dfe_sign_in_session["last_name"],
-      # id_token: dfe_sign_in_session["id_token"],
-      # provider: dfe_sign_in_session["provider"],
+      id_token: dfe_sign_in_session["id_token"],
+      provider: dfe_sign_in_session["provider"],
       service: session["service"],
     )
   end
 
   def user
-    # TODO: When dfe sign-in is fully implemented, we will be able to find the user by the id == id_token.
-    @user ||= user_for_service
+    @user ||= user_by_uid || user_by_email
   end
 
   def self.end_session!(session)
@@ -54,12 +57,35 @@ class DfESignInUser
 
   private
 
-  def user_for_service
+  def user_by_uid
+    return if dfe_sign_in_uid.blank?
+
+    support_user_klass.find_by(dfe_sign_in_uid:) ||
+      user_klass.find_by(dfe_sign_in_uid:)
+  end
+
+  def user_by_email
+    return if email.blank?
+
+    support_user_klass.find_by(email:) ||
+      user_klass.find_by(email:)
+  end
+
+  def support_user_klass
     case service
-    when :claims
-      Claims::SupportUser.find_by(email:) || Claims::User.find_by(email:)
     when :placements
-      Placements::SupportUser.find_by(email:) || Placements::User.find_by(email:)
+      Placements::SupportUser
+    when :claims
+      Claims::SupportUser
+    end
+  end
+
+  def user_klass
+    case service
+    when :placements
+      Placements::User
+    when :claims
+      Claims::User
     end
   end
 end
--- a/app/services/dfe_sign_in/user_update.rb
+++ b/app/services/dfe_sign_in/user_update.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+module DfESignIn
+  class UserUpdate
+    include ServicePattern
+
+    attr_reader :current_user, :successful
+    alias_method :successful?, :successful
+
+    def initialize(current_user:, sign_in_user:)
+      @current_user = current_user
+
+      attributes = {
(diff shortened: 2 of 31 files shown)


CatalinVoineag:
[DfE sign in](https://trello.com/c/3vNaoJhG/106-dfe-sign-in)

CatalinVoineag:
[Investigate how other DfE Rails apps integrate with DfE Sign-in](https://trello.com/c/AUp2psWt/16-investigate-how-other-dfe-rails-apps-integrate-with-dfe-sign-in)

config/locales/en.yml
```diff
@@ -4,6 +4,7 @@ en:
   no_results_with_filter_and_search: There are no results for '%{for}' and the selected filter.
   no_results_with_filter: There are no results for the selected filter.
   you_cannot_perform_this_action: You cannot perform this action
+  you_dont_have_access_to_this_service: You don't have access to this service
```
JamieCleare2525, line 7: Personal preference: (feel free to ignore)
Could this be changed to:
```suggestion
  you_do_not_have_access_to_this_service: You do not have access to this service
```
I'm not a fan of contraction in things like error messages.

db/seeds.rb
```diff
@@ -69,4 +69,12 @@
   end
 end
 
+# Claims::User for DFE SIGN IN, we don't have an invite process yet
+Claims::SupportUser.find_or_create_by!(
+  first_name: "Catalin",
```
JamieCleare2525, line 74: Are you sure you want to use your own name here?

JamieCleare2525: Perhaps these should be environment variables?

CatalinVoineag: This has been resolved here https://github.com/DFE-Digital/itt-mentor-services/pull/184/commits/aa850c5ca2aa452233102b5b30dcd7ac478329fd

We had a chat and I think we want to allow any user with a valid dfe email or a dfe uid to login, so this means that  a support user can create a user in our app and if it has the same email address as the the dfe sign in email address we allow them to log in.

I'll change the description of the PR as well

config/locales/en.yml
CatalinVoineag: This has been resolved here https://github.com/DFE-Digital/itt-mentor-services/pull/184/commits/aa850c5ca2aa452233102b5b30dcd7ac478329fd

app/services/dfe_sign_in/user_update.rb
```diff
@@ -0,0 +1,31 @@
...
+    include ServicePattern
+
+    attr_reader :current_user, :successful
+    alias_method :successful?, :successful
+
+    def initialize(current_user:, sign_in_user:)
+      @current_user = current_user
+
+      attributes = {
+        last_signed_in_at: Time.zone.now,
```
Nitemaeric, line 14:
```suggestion
        last_signed_in_at: Time.current,
```

lib/hosting_environment.rb
```diff
@@ -15,4 +15,29 @@ def self.phase(current_service)
                  env
                end
   end
+
+  def self.dfe_sign_in_client_id(current_service)
+    ENV["#{current_service.upcase}_DFE_SIGN_IN_CLIENT_ID"]
```
Nitemaeric, line 20: I'm worried this will be hard to pin down if something goes wrong here. Can we use a hash lookup like above instead of string manipulation?

app/views/sessions/new.html.erb
```diff
@@ -8,7 +8,7 @@
       <% if ENV["SIGN_IN_METHOD"] == "persona" %>
         <%= govuk_button_to t(".sign_in_with_persona"), personas_path, method: :get %>
       <% else %>
-        <%= govuk_button_to t(".sign_in_with_dsi"), "#" %>
+        <%= govuk_button_to t(".sign_in_with_dsi"), "/auth/dfe" %>
```
Nitemaeric, line 11: We should use a path helper here.

app/models/dfe_sign_in_user.rb
```diff
@@ -57,9 +60,17 @@ def self.end_session!(session)
...
+      Claims::SupportUser.find_by(find_by_attribute) || Claims::User.find_by(find_by_attribute)
     when :placements
-      Placements::SupportUser.find_by(email:) || Placements::User.find_by(email:)
+      Placements::SupportUser.find_by(find_by_attribute) || Placements::User.find_by(find_by_attribute)
+    end
+  end
+
+  def find_by_attribute
+    if provider == "dfe"
+      "dfe_sign_in_uid='#{dfe_sign_in_uid}' OR email='#{email}'"
```
Nitemaeric, line 71: Will this always find by uid first?

If we have a match for both `uid` and `email`, is it dependent on the user `id` - default sort order.

My trail of thinking is:

1. SQL will filter the results down to 2 records and return them sorted by default table sort order.
2. Rails will select the first record due to `find` instead of `where`.

Nitemaeric (commented):
Looks good!

app/views/sessions/new.html.erb
CatalinVoineag: This path is not in our routes path, it's a path opened by the omniauth config, going to the DFE sign in form

app/models/dfe_sign_in_user.rb
CatalinVoineag: So these are the two sql queries that can get executed when we use dfe sign in

It will always find by uid first. So if you have two records here the first one will be the one with the UID that matches. In theory we should not have 2 or more users with same uids or emails here
```
  Claims::SupportUser Load (0.5ms)  SELECT "users".* FROM "users" WHERE "users"."type" = $1 AND (dfe_sign_in_uid='0F1D1624-4BF2-4B3C-8BE6-57B7AB22F2CF' OR email='catalin.voineag@educa
tion.gov.uk') LIMIT $2  [["type", "Claims::SupportUser"], ["LIMIT", 1]]
  ↳ (byebug):1:in `user_for_service'
nil
  Claims::User Load (0.7ms)  SELECT "users".* FROM "users" WHERE "users"."type" = $1 AND (dfe_sign_in_uid='0F1D1624-4BF2-4B3C-8BE6-57B7AB22F2CF' OR email='catalin.voineag@education.go
v.uk') LIMIT $2  [["type", "Claims::User"], ["LIMIT", 1]]

```

lib/hosting_environment.rb
CatalinVoineag: Addressed https://github.com/DFE-Digital/itt-mentor-services/pull/184/commits/40e0a910922358b89dc9c8585fbfb06bb82a1ec5

app/models/dfe_sign_in_user.rb
Nitemaeric: I mean, let's say we have the following:

```ruby
user_1 = create(:user, uid: 123)
user_2 = create(:user, email: "123@example.com")
```

Will an oauth payload of `{ uid: 123, email: "123@example.com" }` always find `user_1` or is it dependent on the default database order? For example, do these return the same record?

```ruby
Claims::User.order(id: :asc).find_by(find_by_attribute)
Claims::User.order(id: :desc).find_by(find_by_attribute)
```

app/views/sessions/new.html.erb
Nitemaeric: I see it. 🤔 Just for consistency, could we add to `routes_helper.rb`?

Maybe?

```ruby
def omniauth_sign_in_path(provider)
  "/auth/#{provider}"
end
```

app/models/dfe_sign_in_user.rb
CatalinVoineag: Oh, I see what you mean. No the current query will not always return user_1. Should be fixed in this https://github.com/DFE-Digital/itt-mentor-services/pull/184/commits/812d3b46bba0af04c83bfeca3aed1d04f66bfa75

app/views/sessions/new.html.erb
CatalinVoineag: https://github.com/DFE-Digital/itt-mentor-services/pull/184/commits/812d3b46bba0af04c83bfeca3aed1d04f66bfa75

Nitemaeric (approved):
Awesome work! Excited to get this up and running! 🚀

JamieCleare2525 (commented):
Code looks good. I'm reserving approval until after tomorrows demo.
Placement also needs to setup their service config. Perhaps we could do that as part of the demo?

CatalinVoineag:
> Code looks good. I'm reserving approval until after tomorrows demo. Placement also needs to setup their service config. Perhaps we could do that as part of the demo?

We can do it as part of the demo, what can go wrong right? :D But someone from Placements needs to pull this branch down as I don't have access to the placement side of the dfe console

JamieCleare2525 (approved):
Setup for both Claims and Placements Tested Locally. LGTM!

CatalinVoineag:
[As a User, I should be able to sign into the service using DfE sign-in](https://trello.com/c/jmVPtXdR/130-as-a-user-i-should-be-able-to-sign-into-the-service-using-dfe-sign-in)

CatalinVoineag:
[As a Support User, I should be able to sign into the service using DfE sign-in](https://trello.com/c/Etd5b2ri/131-as-a-support-user-i-should-be-able-to-sign-into-the-service-using-dfe-sign-in)

PR #186, opened 2024-01-31
Lb/allow support users to remove users from organisations

elceebee:
## Context

Support users in the placements service should be able to remove users from organisations.

- Moved the user invite email to a UserMailer class (and created a removal email)
- Styling from the prototype for the Remove link
- Views, routes, controllers and mailers for removing the user from an organisation

## Guidance to review

- Sign in to the placements service as a support user
- Go to an organisation with users
- Remove a user
- Check Notify that an email has been sent

## Link to Trello card

https://trello.com/c/RmON2bbZ

## Things to check

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/44073106/96561d38-1e97-43ff-9db3-ad13f40628de


elceebee:
[Allow support user to remove users from organisations](https://trello.com/c/RmON2bbZ/157-allow-support-user-to-remove-users-from-organisations)

app/controllers/placements/support/organisations/users_controller.rb
```diff
@@ -26,8 +25,26 @@ def create
     end
   end
 
+  def remove; end
+
+  def destroy
+    membership = @user.memberships.find_by(organisation: @organisation)
+    if membership.present? && membership.destroy
+      RemoveUserService.call(@user, @organisation)
```
ollietreend, line 33: This is not a change request – just an opinion and some thinking out loud.

It feels weird (slightly pointless?) to have an entire service object simply as a wrapper around sending the notification email. The service only does one thing so it just adds a layer of abstraction which you might consider unnecessary.

But I appreciate that this is a continuation of an earlier conversation with @Nitemaeric about where to locate this sort of business logic: https://github.com/DFE-Digital/itt-mentor-services/pull/142#discussion_r1461766399

I'm wondering whether it'd be appropriate to move the database changes into the service object too? I could imagine the service object accepting a `Membership` or a `User` + `Organisation` and performing the destroy and sending the email notification as one action. That seems like an appropriate coupling to me because it describes an atomic interaction/transaction in terms of business logic.

app/controllers/placements/support/organisations/users_controller.rb
```diff
@@ -26,8 +25,26 @@ def create
...
+      redirect_to_index
+      flash[:success] = t(".user_removed")
+    else
+      # TODO: There is no design for handling failure? Eg, if the membership has already been deleted?
+      redirect_to_index
+    end
```
ollietreend, line 39: This is a bit of a weird edge case, isn't it. It's entirely possible that this could occur, but without knowing the reasons (e.g. a race condition, or a database connection issue, or a faulty flux capacitor..) it's difficult to know how to gracefully handle it without making assumptions.

I wonder if we would be better using `membership.destroy!` – in other words, keeping it simple. And then if/whenever this specific failure occurs, we'll have error tracking so we can inspect and debug it.

My hunch is that we'll never, or very rarely, see it happen in the real world. So it's probably not worthy of our attention until it actually affects users.

app/mailers/remove_user_mailer.rb
```diff
@@ -0,0 +1,17 @@
...
+        organisation_name: organisation.name,
+      ),
+      body: t("#{user.service}.mailers.remove_user.body",
+              user_name: user.full_name,
+              organisation_name: organisation.name),
+    }
+
+    notify_email(mailer_options)
+  end
+end
```
ollietreend, line 17: I'm wondering if this mailer, and the [existing "user invite" mailer](https://github.com/DFE-Digital/itt-mentor-services/blob/bf8bde00122b6b4d52552d97f8e71423e667e51b/app/mailers/notify_mailer.rb#L3-L17), would be better off grouped into a single mailer.

For example:

```ruby
class UserMailer < ApplicationMailer
  def invitation_email(user, organisation)
    # do something
  end

  def removal_email(user, organisation)
    # do something
  end
end
```

This seems like a reasonable grouping to make, and follows the Rails guidance which says that "[mailers are similar to controllers](https://guides.rubyonrails.org/action_mailer_basics.html#mailers-are-similar-to-controllers)" in that they can have multiple "actions".

What do you think?

app/views/claims/support/schools/users/check.html.erb
```diff
@@ -86,6 +86,6 @@
           <%= govuk_link_to(t(".cancel"), claims_support_school_users_path(@school)) %>
         </p>
       <% end %>
-    <div>
-  <div>
+    </div>
+  </div>
```
ollietreend, line 90: 🫶

config/locales/en/placements/support/schools/users.yml
```diff
@@ -18,4 +18,14 @@ en:
...
+          destroy:
+            user_removed: User removed
+          show:
+            remove_user: Remove user
+          remove:
+            page_title: Are you sure you want to remove this user? - %{user_name} - %{organisation_name}
+            are_you_sure: Are you sure you want to remove this user?
+            remove_user: Remove user
+            cancel: Cancel
+            warning: The user will be sent an email to tell them you removed them from %{organisation_name}
```
ollietreend, line 31: Not a strong opinion, but it could make the view more readable:

```suggestion
            user_will_be_sent_an_email: The user will be sent an email to tell them you removed them from %{organisation_name}
```

app/controllers/placements/support/organisations/users_controller.rb
```diff
@@ -26,8 +25,26 @@ def create
...
+  end
+
   private
 
+  def set_user
+    @user = users.find(params.require(:id))
+  end
```
ollietreend, line 46: Nice DRY-up, and "yay" for removal of `.becomes(Placements::User)` 🙌🏻

ollietreend (commented):
Excellent work! ⭐ This is very nice indeed.

I have no change requests, but I've added a few opinions which I'd be keen to talk through to see what you think.

app/mailers/remove_user_mailer.rb
elceebee: This seems sensible to me! Will make the change / refactor the other mailer

JamieCleare2525 (approved):
LGTM

app/services/remove_user_service.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  attr_reader :user, :organisation
+
+  def call
+    membership = user.memberships.find_by(organisation:)
+    ActiveRecord::Base.transaction do
+      membership.destroy!
+      UserMailer.removal_email(user, organisation).deliver_later
+    end
```
ollietreend, line 16: I suspect the database transaction has no effect here – since `membership.destroy!` is the only thing which hits the database. And once you've sent an email, you can't roll it back even as part of a transaction.

So I think it's safe to remove. But it causes no harm being here either.

ollietreend (approved):
Excellent! 😍 

I'm approving this because I'd be happy for you to merge it in regardless of whether or not you address this final comment.

PR #199, opened 2024-02-01
Claims: Fix Primary Navigation

Nitemaeric:
## Context

- We want to ensure that the Primary Navigation is always full screen width.
- When the Primary Navigation immediately follows the Phase Banner, the Phase Banner's bottom border should be hidden.
- When browsing nested paths, the top-level Primary Navigation Item should be active.

- I've had to update the school scoping, as we've updated `SupportUser` models to no longer have the `#schools` association - which was admittedly wrong anyway. A support user does not belong to a school, they manage _all_ schools.
  - This allows support users to view all user-facing school pages.
  - Uses the https://github.com/varvet/pundit gem.

## Guidance to review

- Visit as many pages as you can as a single-org User
- Visit as many pages as you can as a multi-org User (phase banner border should be present)
- Visit as many pages as you can as a Support User

## Screenshots

| Type            | Screenshot |
| --------------- | ---------- |
| Single-org User | CleanShot 2024-02-01 at 17 01 25 |
| Multi-org User  | CleanShot 2024-02-01 at 17 01 35 |
| Support User    | CleanShot 2024-02-02 at 15 24 07 |

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -76,6 +76,9 @@ gem "pg_search"
 # Hotwire's modest JavaScript framework [https://stimulus.hotwired.dev]
 gem "stimulus-rails"
 
+# User-Resource Permissions and Scoping
+gem "pundit"
+
 group :development do
   gem "annotate", require: false
   gem "prettier_print", require: false
--- a/app/policies/claims/school_policy.rb
+++ b/app/policies/claims/school_policy.rb
@@ -1,0 +1,11 @@
+class Claims::SchoolPolicy < ApplicationPolicy
+  class Scope < ApplicationPolicy::Scope
+    def resolve
+      if user.support_user?
+        scope.all
+      else
+        scope.joins(:users).where(users: { id: user })
+      end
+    end
+  end
+end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -330,6 +330,8 @@ GEM
     public_suffix (5.0.4)
     puma (6.4.2)
       nio4r (~> 2.0)
+    pundit (2.3.1)
+      activesupport (>= 3.0.0)
     racc (1.7.3)
     rack (3.0.8)
     rack-oauth2 (2.2.0)
@@ -592,6 +594,7 @@ DEPENDENCIES
   pry-byebug
   pry-rails
   puma (~> 6.4)
+  pundit
   rails (~> 7.1.3)
   rails-controller-testing
   rails-erd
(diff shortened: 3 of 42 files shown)


Nitemaeric:
[Fix fiddly banners](https://trello.com/c/J1JMc70H/115-fix-fiddly-banners)

Gemfile
```diff
@@ -74,6 +74,9 @@ gem "pg_search"
 # Hotwire's modest JavaScript framework [https://stimulus.hotwired.dev]
 gem "stimulus-rails"
 
+# User-Resource Permissions and Scoping
+gem "pundit"
+
```
ollietreend, line 79: This seems like a useful pattern to introduce, but it's the sort of change that I think warrants being documented in an ADR.

If we consider this to be our chosen pattern for handling user authorisation, then it should be adopted wholesale throughout the app.

app/policies/claims/school_policy.rb
ollietreend, line 1: Should this file be covered with a unit test?

Pundit has [RSpec helpers for Policy specs](https://github.com/varvet/pundit#rspec) – but I appreciate we're not using a policy (yet).

But it seems like a good idea to cover the Scope with a test too?

app/policies/claims/school_policy.rb
```diff
@@ -0,0 +1,11 @@
+class Claims::SchoolPolicy < ApplicationPolicy
+  class Scope < ApplicationPolicy::Scope
+    def resolve
+      if user.support_user?
+        scope.all
+      else
+        scope.joins(:users).where(users: { id: user })
```
ollietreend, line 7: It's a shame `user.schools` can't be used here 😢 that would have made for a nicer read.

ollietreend (commented):
My comments are specifically on the bits that touch both sides of the codebase 🙂

app/policies/claims/school_policy.rb
Nitemaeric: 😢

CatalinVoineag (approved):
This looks very good, I agree with the ADR point though I don't mind if we do it in a separate PR. This is ready to go after the existing comments have been addressed

Nitemaeric:
Started a PR for the ADR at https://github.com/DFE-Digital/itt-mentor-services/pull/212

ollietreend (approved):
Nice – this looks good!

Thanks for opening #212 for the Pundit ADR. I'm happy for this PR to be merged ahead of that ADR. 👍🏻

PR #203, opened 2024-02-02
Refactor system specs to use mock dfe sign in

CatalinVoineag:
## Context

There are a lot of changes but most of them just implement a helper method

Now that we have dfe sign in DfE sign in, we should have integration
tests for it. The problem is that the omniauth.rb file that configures
the persona or the DFE sign in path gets loaded only once so we cannot
switch between the two when testing.

We can technically reload the file but rails will complain about trying
to modify frozen strings.

Persona login path is only used for developers, and QA. In production we
will always use DFE sign in so our specs should reflect this, when we test
the login path, we should only test DFE sign in not persona.

This commit adds a dfe_sign_in_user_helper to allow us to easily create
the omniauth payload that dfe sign in needs by passing the user we want
to sign in

- Changed specs to sign in users via dfe sign in
- SIGN_IN_METHOD is dfe-sign-in for test env
- dfe_sign_in_user_helper to make it easy to login the user you want
- Spec that cheks the Persona sign in button is there.

## Guidance to review

- Review the changes?
- If the specs pass here it should be fine 😅


CatalinVoineag:
[Refactor integrations specs to use DFE sign in, not persona login](https://trello.com/c/2yGEoapm/163-refactor-integrations-specs-to-use-dfe-sign-in-not-persona-login)

elceebee (approved):

JamieCleare2525 (approved):
Reviewed and tested locally. Really nice work.

PR #205, opened 2024-02-02
lb/Allow non-support users to remove other users

elceebee:
## Context

Users should be allowed to remove other users from their organisations
They should *not* be allowed to remove themselves.

Controllers, views, specs for users 
Also moved the "Change organisation" component to the primary nav for non-support users. So the views are just a little cleaner and less repetitive.

## Guidance to review

Login as a placements user (non support)
View the user you are logged in as 
You should *not* see the Remove user link
Go back

View a user that is not you.
You should see the Remove user link
And you should be able to remove the user.

Do this all for both providers and schools.

## Link to Trello card

https://trello.com/c/294x8A8L

## Things to check

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/44073106/c8419fb7-59b5-48af-9bdf-b2e204e0c2ea

### DIFF
--- a/app/policies/placements/user_policy.rb
+++ b/app/policies/placements/user_policy.rb
@@ -1,0 +1,5 @@
+class Placements::UserPolicy < ApplicationPolicy
+  def destroy?
+    user != record
+  end
+end
--- a/app/views/placements/schools/users/index.html.erb
+++ b/app/views/placements/schools/users/index.html.erb
@@ -1,13 +1,5 @@
-<% content_for(:page_title) { t(".users") } %>
-<% if current_user.organisation_count > 1 %>
-  <%= content_for(:header_content) do %>
-    <%= render(ContentHeaderComponent.new(
-      title: @organisation.name,
-      actions: [govuk_link_to(t(".change_organisation"), placements_organisations_path, no_visited_state: true)],
-    )) %>
-  <% end %>
-<% end %>
 <%= render "placements/primary_navigation", organisation: @organisation, current_navigation: :users %>
+<% content_for(:page_title) { t(".users") } %>
 
 <div class="govuk-width-container">
   <h1 class="govuk-heading-l"><%= t(".users") %></h1>
--- a/spec/policies/placements/user_policy_spec.rb
+++ b/spec/policies/placements/user_policy_spec.rb
@@ -1,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe Placements::UserPolicy do
+  subject { described_class.new(current_user, user).destroy? }
+
+  context "where current user and user are the the same user" do
+    let(:current_user) { create(:placements_user) }
+    let(:user) { current_user }
+
+    it "returns false, user cannot destroy themselves" do
+      expect(subject).to eq false
+    end
+  end
+
+  context "where current user and user are different users" do
+    let(:current_user) { create(:placements_user) }
+    let(:user) { create(:placements_user) }
+
+    it "returns true, user can destroy other users" do
+      expect(subject).to eq true
+    end
+  end
+end
--- a/spec/system/placements/organisations/users/user_removes_other_users_from_organisation_spec.rb
+++ b/spec/system/placements/organisations/users/user_removes_other_users_from_organisation_spec.rb
@@ -28,7 +28,7 @@ RSpec.describe "Placements support user removes a user from an organisation", ty
 
     scenario "I try to remove myself from a school" do
       given_i_am_signed_in_as_mary
-      when_i_try_to_visit_the_remove_path_form_myself(mary, school)
+      when_i_try_to_visit_the_remove_path_for(mary, school)
       then_i_am_redirected_to_the_index_page(school)
     end
   end
@@ -60,7 +60,7 @@ RSpec.describe "Placements support user removes a user from an organisation", ty
 
     scenario "I try to remove myself from a school" do
       given_i_am_signed_in_as_mary
-      when_i_try_to_visit_the_remove_path_form_myself(mary, provider)
+      when_i_try_to_visit_the_remove_path_for(mary, provider)
       then_i_am_redirected_to_the_index_page(provider)
     end
   end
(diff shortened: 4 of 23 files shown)


spec/system/placements/organisations/users/user_removes_other_users_from_organisation_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+      then_i_am_asked_to_confirm(anne, school)
+      when_i_click_on("Remove user")
+      then_the_the_user_is_removed_from_the_organisation(anne, school)
+      and_message_is_sent_to_user(anne, school)
+    end
+
+    scenario "I try to remove myself from a school" do
+      given_i_am_signed_in_as_mary
+      when_i_try_to_visit_the_remove_path_form_myself(mary, school)
+      then_i_am_redirected_to_the_index_page(school)
```
ollietreend, line 32: Would be sensible to additionally assert that the "Remove user" link doesn't show?

```suggestion
      given_i_am_signed_in_as_mary
      and_i_visit_the_user_page(mary)
      then_i_cannot_see("Remove user")
      when_i_try_to_visit_the_remove_path_form_myself(mary, school)
      then_i_am_redirected_to_the_index_page(school)
```

**Note:** `then_i_cannot_see` is made up!

spec/system/placements/organisations/users/user_removes_other_users_from_organisation_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+  def when_i_click_on(text)
+    click_on text
+  end
+
+  def and_i_visit_the_user_page(user)
+    click_on "Users"
+    click_on user.full_name
+  end
+
+  def when_i_try_to_visit_the_remove_path_form_myself(user, organisation)
```
ollietreend, line 92: Nitpick: This this a typo?

```suggestion
  def when_i_try_to_visit_the_remove_path_for_myself(user, organisation)
```

And actually, now I think about it, is `_myself` actually needed? Looks like it could be used for any user, not only the current user.

```suggestion
  def when_i_try_to_visit_the_remove_path_for(user, organisation)
```

spec/system/placements/organisations/users/user_views_other_users_spec.rb
```diff
@@ -67,6 +81,14 @@ def then_i_see_user_details(user:)
     expect(page).to have_content user.email
   end
 
+  def and_i_see_the_remove_user_link
+    expect(page).to have_link "Remove user"
+  end
+
+  def and_i_do_not_see_the_remove_user_link
+    expect(page).to_not have_link "Remove user"
+  end
```
ollietreend, line 90: Oh! My mistake. I think you can probably ignore the comment above, since you've covered it in this system spec instead.

ollietreend (commented):
This is looking excellent! 😍  Nice work.

Just a few comments from me (including some comments which cancel themselves out..)

No changes required – just pointing things out.

LeaWelch (approved):

spec/system/placements/organisations/users/user_removes_other_users_from_organisation_spec.rb
elceebee: tested in the other spec! :)

app/views/placements/schools/users/index.html.erb
```diff
@@ -1,13 +1,5 @@
...
-<% if current_user.organisation_count > 1 %>
-  <%= content_for(:header_content) do %>
-    <%= render(ContentHeaderComponent.new(
-      title: @organisation.name,
-      actions: [govuk_link_to(t(".change_organisation"), placements_organisations_path, no_visited_state: true)],
-    )) %>
-  <% end %>
-<% end %>
 <%= render "placements/primary_navigation", organisation: @organisation, current_navigation: :users %>
+<% content_for(:page_title) { t(".users") } %>
```
CatalinVoineag, line 2:
```suggestion
<%= content_for(:page_title) { t(".users") } %>
```
Does this need to be like this?

CatalinVoineag (approved):
Looks good! Just 1 small thing but not a blocker

app/views/placements/schools/users/index.html.erb
ollietreend: I don't think `=` is required here because it's not actively outputting something to the page. The return value of `content_for` is presumably `nil` because it performs a side-effect (i.e. setting a value for the layout) rather than actively causing output itself.

ollietreend: In other words: I think they'll be functionally the same with or without the `=`.

app/policies/placements/user_policy.rb
```diff
@@ -0,0 +1,5 @@
+class Placements::UserPolicy < ApplicationPolicy
+  def destroy?
+    user != record
+  end
+end
```
ollietreend, line 5: Nice! 😍

spec/policies/placements/user_policy_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    it "returns false, user cannot destroy themselves" do
+      expect(subject).to eq false
+    end
+  end
+
+  context "where current user and user are different users" do
+    let(:current_user) { create(:placements_user) }
+    let(:user) { create(:placements_user) }
+
+    it "retursn true, user can destroy other users" do
```
ollietreend, line 19:
```suggestion
    it "returns true, user can destroy other users" do
```

ollietreend (approved):
This looks great 😍

I've spotted a few small typos, and now that I've seen them I can't unsee them. But I'm approving this PR on the assumption that a) you'll fix them before merging, or b) I'm a pedant and it doesn't matter in the grand scheme of things. 🙃

In addition to the one comment, I also spotted typos in these commit messages:

- Use policy to determine **~is~ if** a user can be destroyed
- **~CR~ PR** comment

spec/policies/placements/user_policy_spec.rb
elceebee: So embarrassing when I'm usually the person complaining about other people's typos.

PR #207, opened 2024-02-02
Add CRU for Placement Support Users

JamieCleare2525:
## Context

Support Users will be able to onboard other support users (placements).

- Add the controllers/views/routes to allow Placement Support Users to add other Support Users.

## Guidance to review

(On Placements Service)
- Login in as Support User Colin
- Click "Users" in the navigation
- This will display a list of Support Users (including Colin)
- Click "Add user"
- Fill in the form, using an email address ending in "@education.gov.uk"
- (Feel free to test form validation)
- Click "Continue"
- This will direct you to the "Check your answers" page
- Click "Add user"
- User should be added successfully
- Click on the name of the new Support User you have added
- This will direct you the the Support User show page.
- You should see the User's details here.

## Link to Trello card

https://trello.com/c/qmhJnd3t/143-support-user-crud

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/17162488/d78ec1b8-1161-49ac-b090-cacf7217ff41

### DIFF
--- a/app/controllers/placements/support/support_users_controller.rb
+++ b/app/controllers/placements/support/support_users_controller.rb
@@ -1,0 +1,51 @@
+class Placements::Support::SupportUsersController < Placements::Support::ApplicationController
+  before_action :set_support_user, only: %i[show]
+
+  def index
+    @support_users = Placements::SupportUser.order(created_at: :desc)
+  end
+
+  def new
+    @support_user = if params[:support_user].present?
+                      Placements::SupportUser.new(support_user_params)
+                    else
+                      Placements::SupportUser.new
+                    end
+  end
+
+  def check
+    @support_user = Placements::SupportUser.new(support_user_params)
+
+    render :new unless @support_user.valid?
+  end
+
+  def create
+    @support_user = Placements::SupportUser.new(support_user_params)
+
+    if SupportUser::Invite.call(support_user: @support_user)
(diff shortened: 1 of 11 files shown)


app/controllers/placements/support/support_users_controller.rb
```diff
@@ -0,0 +1,51 @@
+class Placements::Support::SupportUsersController < Placements::Support::ApplicationController
+  before_action :set_support_user, only: %i[show]
+
+  def index
+    @support_users = Placements::SupportUser.order(created_at: :desc)
```
CatalinVoineag, line 5: I assume we won't have too many support users to require pagination?

CatalinVoineag (approved):
Looks good, one comment, not a blocker

app/controllers/placements/support/support_users_controller.rb
JamieCleare2525: That would be my assumption. This is extracted some the same implementation on Claim, so I'm guessing that this is ok. There also doesn't appear to be pagination on the prototype

CatalinVoineag: There is pagination on the support#org#index page, but you need to create over 25 organisations to get it to show. It's OK, if we find a need we can add it later 🚀

PR #209, opened 2024-02-05
Add Rspec smoke tests

johnake:
## Things to check


johnake:
[ITTMS: Implement simple smoke tests](https://trello.com/c/oMVpoXmi/1592-ittms-implement-simple-smoke-tests)

.github/workflows/actions/smoke-tests/action.yml
```diff
@@ -0,0 +1,34 @@
...
+        run: |
+            gem install bundler
+            echo 'gem "rspec"' >> Gemfile
+            bundle
+
+      - name: Run ${{ inputs.environment }} tests
+        run: bin/bundle exec rspec --tag smoke_test -b
+        env:
+            SMOKE_TEST: "true"
+            CLAIMS_EXTERNAL_HOST: "track-and-pay-${{ inputs.environment }}.test.teacherservices.cloud"
```
saliceti, line 33: This is not valid for the production environment
Get the values from the yml file

terraform/application/config/qa.tfvars.json
```diff
@@ -1,5 +1,7 @@
 {
     "cluster": "test",
     "namespace": "bat-qa",
-    "enable_postgres_ssl": true
-}
+    "environment": "qa",
+    "deploy_azure_backing_services": false,
```
saliceti, line 5: Is this for testing?

.github/workflows/actions/smoke-tests/action.yml
saliceti: You can read it from the module output https://github.com/DFE-Digital/terraform-modules/blob/main/aks/application/outputs.tf#L6

saliceti: Or add logic in the workflow

terraform/application/config/qa.tfvars.json
johnake: The looks like this on the branch:

    "cluster": "test",
    "namespace": "bat-qa",
    "environment": "qa",
    "deploy_azure_backing_services": false,
    "enable_postgres_ssl": false
  }

Should deploy_azure_backing_services and/or enable_postgres_ssl be set to true ?

.github/workflows/actions/smoke-tests/action.yml
johnake: Fixed

terraform/application/outputs.tf
```diff
@@ -0,0 +1,17 @@
...
+# }
+
+# output "claims_web_external_hostname" {
+#   value = module.web_application.web_external_hostnames[0]
+# }
+
+# # Access the second hostname
+# output "placements_external_hostname" {
+#   value = module.web_application.hostname[1]
+# }
```
RMcVelia, line 13: can we remove the above comments if not using these?

terraform/application/config/qa.tfvars.json
```diff
@@ -1,5 +1,6 @@
 {
     "cluster": "test",
     "namespace": "bat-qa",
+    "deploy_azure_backing_services": true,
```
RMcVelia, line 4: this should be the default, so we don't normally specify it, same for enable_postgres_ssl below

.github/actions/deploy-environment/action.yml
```diff
@@ -35,3 +35,36 @@ runs:
...
+             CLAIMS_EXTERNAL_HOST=$(echo $HOSTNAMES | jq -r '.[0]')
+             PLACEMENTS_EXTERNAL_HOST=$(echo $HOSTNAMES | jq -r '.[1]')
+
+             export CLAIMS_EXTERNAL_HOST=$CLAIMS_EXTERNAL_HOST
+             export PLACEMENTS_EXTERNAL_HOST=$PLACEMENTS_EXTERNAL_HOST
+
+             bin/bundle exec rspec --tag smoke_test -b
+            env:
+                SMOKE_TEST: "true"
+                HOSTNAMES: ${{ env.HOSTNAMES }}
```
RMcVelia, line 70: Might want to check if the devs want slack notifications for smoke test failures for review apps, as I think that will happen looking at the build-and-deploy workflow (you could test this out to be sure).

johnake: We haven't enabled slack notification for the review App

terraform/application/outputs.tf
johnake: Fixed

terraform/application/config/qa.tfvars.json
johnake: Removed

.github/actions/deploy-environment/action.yml
```diff
@@ -35,3 +35,36 @@ runs:
...
+            uses: ruby/setup-ruby@v1
+            with:
+              ruby-version: 3.2.2
+
+          - name: Bundle smoke test gems
+            shell: bash
+            run: |
+                gem install bundler
+                echo 'gem "rspec"' >> Gemfile
+                bundle
```
Nitemaeric, line 56: I think you can merge these steps by following this https://github.com/ruby/setup-ruby?tab=readme-ov-file#single-job

```suggestion
          - name: Set up Ruby 3.2.2
            uses: ruby/setup-ruby@v1
            with:
              ruby-version: 3.2.2
              bundler-cache: true
```

spec/rails_helper.rb
```diff
@@ -28,12 +28,16 @@
...
 # Rails.root.glob('spec/support/**/*.rb').sort.each { |f| require f }
 
-# Checks for pending migrations and applies them before tests are run.
-# If you are not using ActiveRecord, you can remove these lines.
-begin
-  ActiveRecord::Migration.maintain_test_schema!
-rescue ActiveRecord::PendingMigrationError => e
-  abort e.to_s.strip
+if ENV["SMOKE_TEST"] == "true"
+  ActiveRecord::Base.establish_connection adapter: :nulldb
```
CatalinVoineag, line 32: Can you add a bit more detail in the commit message for why we do this?

Something like:
```
We're using the nulldb gem because our setup requires us to mock a db connection running locally even though we point our capybara instance to a remote server to run these smoke tests.
```

CatalinVoineag (commented):
Looks good to me, can we add more detail on the PR and commit descriptions?

PR #211, opened 2024-02-05
Resolve small issues with Persona login tests

JamieCleare2525:
## Context/Changes proposed in this pull request

- Correct order Personas on the Persona login page
- Correct upper case "Sign In" to "Sign in"
- Named and fixed spec file which:
  - Didn't run, because it doesn't end in `_spec.rb`
  -  Relied on Persona login, which (@CatalinVoineag confirmed) can't be tested easily with our setup. So converted to use DfE Sign In

## Screenshots
_Before_
screencapture-placements-localhost-3000-personas-2024-02-05-13_20_59

_After_
screencapture-placements-localhost-3000-personas-2024-02-05-13_20_16


spec/system/claims/view_a_users_details.rb
```diff
@@ -1,46 +0,0 @@
...
-  def create_user
-    @user = create(:user, :claims, first_name: "Barry", last_name: "Garlow", email: "barry.garlow@gmail.com")
-  end
-
-  def attach_user_to_school
-    create(:membership, user: @user, organisation: @school)
-  end
-
-  def sign_in_as_support_user
-    create(:claims_support_user, :colin)
```
CatalinVoineag, line 31: So to make this work with DFE sign in we need to do something like 

```
def sign_in_as_support_user
    user = create(:claims_support_user, :colin)
    user_exists_in_dfe_sign_in(user:)
    visit sign_in_path
    click_on "Sign in using DfE Sign In"

end
```

spec/system/claims/view_a_users_details.rb
```diff
@@ -1,46 +0,0 @@
...
-  end
-
-  private
-
-  def setup_school
-    @school = create(:school, :claims, urn: "123456", name: "Test School")
-  end
-
-  def create_user
-    @user = create(:user, :claims, first_name: "Barry", last_name: "Garlow", email: "barry.garlow@gmail.com")
```
CatalinVoineag, line 23: And this needs to be something like `create(:cliams_user)`

CatalinVoineag (commented):
Sorry Jamie, I misunderstood your problem. we cannot test the persona login but we can make the test use dfe sign in and the user you sign in is a support user

spec/system/claims/view_a_users_details.rb
JamieCleare2525: resolved in [3c7e68f](https://github.com/DFE-Digital/itt-mentor-services/pull/211/commits/3c7e68fdbea308e765f63187dba7fb6dbb46d2c7)

CatalinVoineag (approved):
Yes! Thank you for catching and fixing our spec 👍🏻

PR #212, opened 2024-02-05
Add User Authorisation ADR

Nitemaeric:
## Context

We want a scalable pattern of working with user permissions.

Add the ADR to adopt Pundit for User permissions.

## Guidance to review

View the rendered ADR: https://github.com/DFE-Digital/itt-mentor-services/blob/adr/user-authorisation/adr/00005-user-authorisation-using-pundit.md

### DIFF
--- a/adr/00005-user-authorisation-using-pundit.md
+++ b/adr/00005-user-authorisation-using-pundit.md
@@ -1,0 +1,84 @@
+# 5. User Authorisation using Pundit
+
+Date: 2024-02-05
+
+## Status
+
+Accepted
+
+## Context
+
+We want to have scalable user authorisation and permissions. We started with conditional scoping and per-case permissions, such as the following code snippets:
+
+- Permissions
+
+  ```erb
+  <% if user != current_user %>
+    <%= link_to "Remove", remove_user_path(user) %>
+  <% end %>
+  ```
+
+- Scopes
+
+  ```ruby
+  class SchoolsController
+    def index
+      @schools = scoped_schools
+    end
+
+    private
+
+    def scoped_schools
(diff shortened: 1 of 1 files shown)


adr/00005-user-authorisation-using-pundit.md
```diff
@@ -0,0 +1,84 @@
...
+## Consequences
+
+Moving forward, user permissions should be defined within policy classes.
+
+The above can now be re-written as:
+
+```ruby
+class UserPolicy < ApplicationPolicy
+  def destroy?
+    current.support_user? && user != current_user
```
ollietreend, line 55: Typo?

```suggestion
    current_user.support_user? && user != current_user
```

adr/00005-user-authorisation-using-pundit.md
```diff
@@ -0,0 +1,84 @@
...
+      else
+        scope.joins(:users).where(users: { id: user })
+      end
+    end
+  end
+end
+
+class SchoolsController
+  def index
+    @schools = policy_scope(School.all)
```
ollietreend, line 81: I realise that `School.all.all` is a perfectly valid thing to do, but it feels odd. The [Pundit docs](https://github.com/varvet/pundit/blob/9dd051102bbce8899a2f5b059cd9692052c3bf8b/README.md?plain=1#L302) seem to prefer passing in just the top-level model in these cases.

```suggestion
    @schools = policy_scope(School)
```

ollietreend (approved):
This looks good to me, so I'm approving it.

However, for the sake of having examples that are technically correct, I've given a couple of suggestions of things that I think might be typos or could change.

PR #219, opened 2024-02-05
Detect code coverage regressions in CI

ollietreend:
## Context

We've spotted a few cases of missing test coverage. While I'm not a "100% coverage" purist, I do think it's healthy for us to have access to coverage reports so we have an overall awareness of our test coverage.

I've added two gems:

1. [SimpleCov](https://github.com/simplecov-ruby/simplecov) – the standard code coverage tool used in Ruby apps
2. [Undercover](https://github.com/grodowski/undercover) – a tool that will analyse the coverage report generated by SimpleCov, and produce friendly & readable warnings when we've forgotten to add test coverage in our branch

I've configured the CI workflow to upload the SimpleCov HTML code coverage report as a build artefact. This makes it really easy for us to see a detailed coverage report for our branches.

I've also added Undercover to the workflow so that PRs are blocked if they introduce untested code to the codebase. Admittedly this sounds like it could be annoying, because nobody wants _yet more_ things that can potentially block the CI pipeline. However in my experience of using this gem on other projects, I've rarely seen it produce false-positives or get in the way. It will merely hold us to account so we don't introduce new code without test coverage.

Error messages reported by Undercover aim to be friendly and helpful. For example, see [this output](https://github.com/DFE-Digital/itt-mentor-services/actions/runs/7789500838/job/21241267282#step:8:8) triggered by [this demo commit](https://github.com/DFE-Digital/itt-mentor-services/commit/d418fbd3089bc42470c0bf3b898cbe4582277981).

## Guidance to review

No specific guidance. If you're feeling really keen, go ahead and push up a commit with some uncovered code to see how it works in practice.

## Link to Trello card

https://trello.com/c/LjtUzbiP/168-detect-code-coverage-regressions-in-ci

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -49,6 +49,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
       - name: Setup Ruby
         uses: ruby/setup-ruby@v1
         with:
@@ -63,6 +65,13 @@ jobs:
         run: |
           bundle exec rails test:prepare db:test:prepare
           bundle exec rspec --format progress
+      - name: Upload code coverage report
+        uses: actions/upload-artifact@v4
+        with:
+          name: code-coverage
+          path: coverage/
+      - name: Detect missing code coverage
+        run: bundle exec undercover --compare origin/main
       - name: Upload failure screenshots
         if: failure()
         uses: actions/upload-artifact@v4
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,8 @@ tmp/pids/*
 !tmp/pids/
 !tmp/pids/.keep
 
+coverage/
+
 .yarn/*
 # Ignore terraform files
 bin/terrafile
--- a/Gemfile
+++ b/Gemfile
@@ -111,9 +111,11 @@ group :test do
   gem "selenium-webdriver"
   gem "shoulda-matchers"
   gem "simplecov", require: false
+  gem "simplecov-lcov", require: false
   # launch browser when inspecting capybara specs
   gem "launchy"
   gem "timecop"
+  gem "undercover"
   gem "webmock"
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -110,6 +110,7 @@ group :test do
   gem "rails-controller-testing"
   gem "selenium-webdriver"
   gem "shoulda-matchers"
+  gem "simplecov", require: false
   # launch browser when inspecting capybara specs
   gem "launchy"
   gem "timecop"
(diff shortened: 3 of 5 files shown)


Nitemaeric:
I wonder if using something like https://github.com/marketplace/actions/simplecov-action might help surface the percentages and ensure a threshold more effectively than `undercover`? The artefacts are a bit of a hassle to get to.

ollietreend:
@Nitemaeric this is a bit of a rambly response, but I'm hoping it explains some of my thinking more clearly while also answering your points. I'm happy to chat any of this through!

> I wonder if using something like https://github.com/marketplace/actions/simplecov-action might help surface the percentages and ensure a threshold more effectively than `undercover`?

I'd be keen to avoid focussing too much on percentages and thresholds, to be honest. The problem is that thresholds apply to the codebase as a whole, rather than individual PRs. I think this can lead to some rigid and annoying situations. For example, if we have a threshold set at 90% coverage, and the codebase currently sits at 90.1%, then all is good. But what if somebody opens a PR which removes a few hundred lines of unused code? The likelihood is that the covered percentage will drop below the threshold, and the PR will become blocked due to missing coverage in other areas of the codebase. I've [seen this sort of thing happen a lot](https://github.com/ministryofjustice/offender-management-allocation-manager/commit/50128f9055da71b4ff0ddf0d4d0e2c531c2b0264) on previous projects, and it's always annoying.

This is why I'm suggesting we use the Undercover gem rather than enforcing a particular coverage percentage. It avoids that entire class of annoyance by only picking up on missed coverage on changes made within the specific PR. The intention is for it to be a tool that 'nudges' us towards covering our code by pointing out missed test cases. The ["Why" section on the project readme](https://github.com/grodowski/undercover?tab=readme-ov-file#why) explains it far better than me. 🙂 

> The artefacts are a bit of a hassle to get to.

The SimpleCov coverage report artefact is a nice side effect, but it's not the focus of this PR. Coverage reports alone will not affect the outcome of CI checks. But I figured since they're generated as part of the test run, we might as well surface them so they're available should we want to refer to them.

The nice thing about SimpleCov's HTML reports is that they're quite detailed. They give you much more than a top-level percentage. But again, they're not the core aim of this PR. It's just that Undercover requires SimpleCov reports to do its thing.

> Anything that _posts_ a comment instead might be easier to notice.

I agree that automated comments on PRs are nice to have. But let's not make 'perfect' the enemy of 'good'. Undercover output is captured within the GitHub Actions workflow, in exactly the same place & way that Rubocop and erb-lint output is captured.

If we think linting errors like this are too hard to find in console output, then we could always consider implementing something like [Danger](https://danger.systems/ruby/) which would put them into PR comments. That supports Rubocop, Undercover, and even our JS tooling like ESLint. (Although apparently not erb-lint.)

Nitemaeric (approved):
I agree with your response. Let's ship this! 🚢

elceebee (approved):
Nice one! This is going to be a really useful tool.

PR #222, opened 2024-02-06
Jc/remove support users

JamieCleare2525:
## Context

Support Users for Claims and Placements services can remove support users from their services.

- Added views to allow a Support User to "Remove" a support user
- Discard gem introduced for soft deletion.

## Guidance to review

_Please test this functionality on both placements and claims services_
- Login as Support User Colin
- Click "Users" on the top navigation
- Click on the name of a Support User (other than Colin), if another support user does not appear in the list create one.
- This will redirect you to the Support User's show page.
- Click "Remove user", at the bottom of the Support User's show page.
- This will redirect you to the Remove page for the Support User.
- Click "Remove user", at the bottom of the Support User's remove page.
- This will redirect you to the Support Users index page.
- The removed support user will no longer appear in the list.

## Link to Trello card

https://trello.com/c/PY0pK2A0/108-as-a-support-user-i-should-be-able-to-remove-a-support-user
https://trello.com/c/gn4fXw2n/165-support-user-can-remove-other-support-users

## Things to check

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/17162488/cc7d3e6e-bcee-4f05-9179-6b17a406e39c


JamieCleare2525:
> How do we want to handle if a user is removed (soft deleted), and then re-added?
> 
> Right now, we get an error, `Email is already in use`. If this is outside of the scope of this ticket, we should probably create new tickets if that's what we want to happen. Worth asking @zoeflan or @simonwhatley or @LeaWelch what the expected behaviour should be.
> 
>  re-activating-a-user.mov

@Nitemaeric you wouldn't happen to know how we are handling this?

app/controllers/placements/support/support_users_controller.rb
```diff
@@ -34,6 +34,8 @@ def show; end
   def remove; end
 
   def destroy
+    authorize @support_user
```
elceebee, line 37: 👍

Nitemaeric:
> > How do we want to handle if a user is removed (soft deleted), and then re-added?
> > Right now, we get an error, `Email is already in use`. If this is outside of the scope of this ticket, we should probably create new tickets if that's what we want to happen. Worth asking @zoeflan or @simonwhatley or @LeaWelch what the expected behaviour should be.
> > re-activating-a-user.mov
> 
> @Nitemaeric you wouldn't happen to know how we are handling this?

We could go about it 2 ways:

1. We follow the example set at [Apply](https://qa.apply-for-teacher-training.service.gov.uk/support/users/support?removed=true) and follow up with a page to restore deleted support users.
2. We update the unique validation with `validates :email, uniqueness: { scope: [discarded_at: nil] }`. The drawback of this is that a deleted support user coming back would be disconnected from previous history.

Does anyone have an initial opinion?

JamieCleare2525:
> > > How do we want to handle if a user is removed (soft deleted), and then re-added?
> > > Right now, we get an error, `Email is already in use`. If this is outside of the scope of this ticket, we should probably create new tickets if that's what we want to happen. Worth asking @zoeflan or @simonwhatley or @LeaWelch what the expected behaviour should be.
> > > re-activating-a-user.mov
> > 
> > 
> > @Nitemaeric you wouldn't happen to know how we are handling this?
> 
> We could go about it 2 ways:
> 
> 1. We follow the example set at [Apply](https://qa.apply-for-teacher-training.service.gov.uk/support/users/support?removed=true) and follow up with a page to restore deleted support users.
> 2. We update the unique validation with `validates :email, uniqueness: { scope: [discarded_at: nil] }`. The drawback of this is that a deleted support user coming back would be disconnected from previous history.
> 
> Does anyone have an initial opinion?

I'd prefer to go the enable/restore route. Means we aren't filling up our database with essentially duplicate users.

elceebee (approved):
Tested locally and it works great, nice job!

I think it's worth having someone from UCD have a look before it is merged in (would be good to start really embedding that step in our process for user-facing changes). 

Also, someone needs to write tickets, possibly wireframes for the reinstating a support user journey.

JamieCleare2525:
> Tested locally and it works great, nice job!
> 
> I think it's worth having someone from UCD have a look before it is merged in (would be good to start really embedding that step in our process for user-facing changes).
> 
> Also, someone needs to write tickets, possibly wireframes for the reinstating a support user journey.

@elceebee 100% agreed on the UCD testing.
I'll deploy the PR to a review environment and request their reviews. 
I'll also probe the question of how to handle reenabling soft deleted users

JamieCleare2525:
[As a Support User, I should be able to remove a Support User](https://trello.com/c/PY0pK2A0/108-as-a-support-user-i-should-be-able-to-remove-a-support-user)

JamieCleare2525:
[Support User can remove other support users](https://trello.com/c/gn4fXw2n/165-support-user-can-remove-other-support-users)

app/controllers/claims/support/support_users_controller.rb
```diff
@@ -31,6 +31,19 @@ def create
...
+  def remove; end
+
+  def destroy
+    authorize @support_user
+
+    if SupportUser::Remove.call(support_user: @support_user)
+      redirect_to claims_support_support_users_path, flash: { success: t(".success") }
+    else
+      flash[:alert] = t(".failure")
+      render :remove
```
Nitemaeric, line 43: Tiny nitpick to inline the flash message.

```suggestion
      render :remove, alert: t(".failure")
```

app/mailers/support_user_mailer.rb
```diff
@@ -10,4 +10,15 @@ def support_user_invitation(support_user)
 
     notify_email to: support_user.email, subject:, body:
   end
+
+  def support_user_removal_notification(support_user)
+    subject = t(".subject", service_name: t("#{support_user.service}.service_name"))
```
Nitemaeric, line 15: Is the [service_name](https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/helpers/application_helper.rb) helper available here?

config/locales/en/placements/support/support_users.yml
```diff
@@ -27,4 +27,14 @@ en:
...
+          remove_user: Remove user
+        remove:
+          page_title: Are you sure you want to remove this user? - %{user_name}
+          title: Are you sure you want to remove this user?
+          warning: The user will be sent an email to tell them you removed them from Manage school placements.
+          submit: Remove user
+          cancel: Cancel
+        destroy:
+          success: User removed
+          failure: There was a problem removing the user
```
Nitemaeric, line 40: EOF newline :)

db/migrate/20240205155801_add_discarded_at_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddDiscardedAtToUsers < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :discarded_at, :datetime
+    add_index :users, :discarded_at
```
Nitemaeric, line 4: I wonder if we need a composite index for

```ruby
add_index :users, [:type, :discarded_at]
```

Nitemaeric: Furthermore, it _might_ even be beneficial to do:

```ruby
add_index :users, [:type, :discarded_at, :email]
```

We'd need to check our queries though.

Nitemaeric (approved):
🚀 Thanks for getting this across!

app/controllers/claims/support/support_users_controller.rb
JamieCleare2525: render doesn't work with flash, you have to declare the separately.

app/mailers/support_user_mailer.rb
JamieCleare2525: We've used `t("#{support_user.service}.service_name")` and `t("#{user.service}.service_name")` in all other mailers. I feel like it's a more guaranteed way of ensuring that the email relates to the service associated with the user, rather than the service it is being sent by.

db/migrate/20240205155801_add_discarded_at_to_users.rb
JamieCleare2525: resolved in[b1e480f](https://github.com/DFE-Digital/itt-mentor-services/pull/222/commits/b1e480fe5463bb30ad62e8d24c407cc6fb3cb962)

PR #223, opened 2024-02-06
Lb/make tests less flaky

elceebee:
## Context
Our tests where JS is enabled fail regularly. They pass locally though, so difficult to replicate

- Remove `js: true` from two tests that were meant to test without JS enabled.
- Added rspec-retry gem (https://github.com/DFE-Digital/rspec-retry)
- Added a retry parameter to our tests that have js enabled

## Guidance to review

Difficult to review this locally as we aren't able to make them fail regularly locally. 

## Link to Trello card

https://trello.com/c/8iPfLU6U

## Things to check

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -108,6 +108,7 @@ group :test do
   gem "capybara-screenshot"
   gem "climate_control"
   gem "rails-controller-testing"
+  gem "rspec-retry"
   gem "selenium-webdriver"
   gem "shoulda-matchers"
   gem "simplecov", require: false
--- a/Gemfile
+++ b/Gemfile
@@ -108,7 +108,7 @@ group :test do
   gem "capybara-screenshot"
   gem "climate_control"
   gem "rails-controller-testing"
-  gem "rspec-retry"
+  gem "rspec-retry", git: "https://github.com/DFE-Digital/rspec-retry.git", branch: "main"
   gem "selenium-webdriver"
   gem "shoulda-matchers"
   gem "simplecov", require: false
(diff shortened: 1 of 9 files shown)


CatalinVoineag (approved):
🎉 ❤️ 🚢 🇮🇹

Gemfile
```diff
@@ -108,6 +108,7 @@ group :test do
   gem "capybara-screenshot"
   gem "climate_control"
   gem "rails-controller-testing"
+  gem "rspec-retry"
```
ollietreend, line 111: Just to point out that this will pull the [rspec-retry](https://rubygems.org/gems/rspec-retry) gem from RubyGems.

However, my understanding is that @JR-G's team forked and extended it as [DFE-Digital/rspec-retry](https://github.com/DFE-Digital/rspec-retry).

The other BAT apps are using the fork:
https://github.com/DFE-Digital/apply-for-teacher-training/blob/32a56b8cb18bc052966c862f816781cefdee0df1/Gemfile#L166

I don't have a strong opinion as to whether we should be using the original gem or the DfE fork. The original gem seems to be abandonware, whereas the DfE fork claims to have improved Ruby 3 compatibility. So maybe we should stick with the fork?

ollietreend (commented):
Yay! 🥳 Thanks for doing this.

Gemfile
elceebee: Ah, thanks for pointing this out. I just did what the docs on the [DFE-Digital/rspec-retry](https://github.com/DFE-Digital/rspec-retry) repo said to do. I'll use the fork instead.

ollietreend (approved):
Looks good to me! 👍🏻

PR #227, opened 2024-02-07
Configure RuboCop to lint RSpec files

ollietreend:
## Context

Our RuboCop config inherits from the `rubocop-govuk` gem, but it didn't enable the RSpec cops. Therefore we don't currently have any opinionated linter rules to standardise the way we write specs.

- Include `config/rspec.yml` from `rubocop-govuk`
- Autocorrect lint errors
- Fix a couple more types of error manually
- Disable cops which violate the rules so we can manually fix them up in subsequent PRs

### Why not use a `.rubocop_todo.yml` file?

RuboCop can automatically generate a 'todo' config to ignore failures only in specific files. It'll produce [something like this](https://github.com/DFE-Digital/register-trainee-teachers/blob/main/.rubocop_todo.yml).

I've deliberately decided not to use this, instead opting to disable the affected cops across the entire codebase. I've done this to avoid any confusing situations where RuboCop inconsistently applies rules and it's not clear why. For example, we wouldn't want to find ourselves falling foul of RuboCop when creating new spec files that copy & paste from existing specs.

This seems like a reasonable decision to make while we have the capacity and momentum to fix-up errors and re-enable those cops.

## Guidance to review

I'd advise reviewing commit by commit, and skipping the "autocorrect" commit because that was autogenerated by RuboCop and is pretty hefty 😅

## Link to Trello card

https://trello.com/c/IzTz6Nvs/170-add-linter-rules-for-rspec

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -35,3 +35,21 @@ RSpec/SpecFilePathFormat:
 
 RSpec/SpecFilePathSuffix:
   Enabled: true
+
+# The following cops are temporarily disabled so we can incrementally
+# introduce RSpec rules while incrementally fixing up the codebase.
+# They should eventually all be re-enabled.
+RSpec/VerifiedDoubles:
+  Enabled: false
+RSpec/ExpectInHook:
+  Enabled: false
+RSpec/StubbedMock:
+  Enabled: false
+RSpec/ContextWording:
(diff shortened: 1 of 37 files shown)


elceebee (approved):
Nice to have some standards enforced in these files!

PR #238, opened 2024-02-09
Enable RSpec/ContextWording rubocop rule

CatalinVoineag:
## Context

We want to enable all the rspec rules in rubocop that are not enabled https://github.com/DFE-Digital/itt-mentor-services/blob/main/.rubocop.yml#L42-L53

This PR enables `RSpec/ContextWording` which expects the context description to be like `Context description should match /^when\b/, /^with\b/, /^without\b/,
/^and\b/, or /^but\b/`

Enable the rule
Change all specs that break this rule

## Guidance to review

Review
Do the tests pass?
Maybe have a brew after?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -45,8 +45,6 @@ RSpec/ExpectInHook:
   Enabled: false
 RSpec/StubbedMock:
   Enabled: false
-RSpec/ContextWording:
-  Enabled: false
 RSpec/InstanceVariable:
   Enabled: false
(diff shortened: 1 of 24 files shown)


CatalinVoineag:
[Fix RSpec/ContextWording linter](https://trello.com/c/BztF7m6j/178-fix-rspec-contextwording-linter)

Nitemaeric (approved):
"when" and "with". Cool 👍

PR #249, opened 2024-02-12
Test Sign out and dfe sign in bad response

CatalinVoineag:
## Context

We want to increase our testing code covrage. This PR adds specs to sign out path and when the omniauth payload from dfe sign in is empty.

## Changes

Specs

## Guidance

Do tests pass?

### DIFF
--- a/spec/support/dfe_sign_in_user_helper.rb
+++ b/spec/support/dfe_sign_in_user_helper.rb
@@ -10,6 +10,13 @@ module DfESignInUserHelper
     )
   end
 
+  def invalid_dfe_sign_in_response
+    {
+      "provider" => "dfe",
+      "uid" => nil,
+    }
+  end
+
   private
 
   def fake_dfe_sign_in_auth_hash(email:, dfe_sign_in_uid:, first_name:, last_name:)
--- a/spec/system/claims/sign_in_as_a_claims_user_spec.rb
+++ b/spec/system/claims/sign_in_as_a_claims_user_spec.rb
@@ -37,6 +37,15 @@ RSpec.describe "Sign In as a Claims User", type: :system, service: :claims do
     and_i_visit_my_account_page
     then_i_see_user_details_for_colin
   end
+
+  context "when response from dfe sign in is invalid" do
+    scenario "I sign in as user colin" do
+      invalid_dfe_sign_in_response
+      when_i_visit_the_sign_in_path
+      when_i_click_sign_in
+      i_do_not_have_access_to_the_service
+    end
+  end
 end
 
 private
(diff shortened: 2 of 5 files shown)


app/models/dfe_sign_in_user.rb
```diff
@@ -14,13 +14,13 @@ def initialize(email:, dfe_sign_in_uid:, first_name:, last_name:, service:, id_t
...
-      "last_name" => omniauth_payload["info"]["last_name"],
+      "email" => omniauth_payload["info"]["email"].presence,
+      "dfe_sign_in_uid" => omniauth_payload["uid"].presence,
+      "first_name" => omniauth_payload["info"]["first_name"].presence,
+      "last_name" => omniauth_payload["info"]["last_name"].presence,
       "last_active_at" => Time.current,
-      "id_token" => omniauth_payload["credentials"]["id_token"],
-      "provider" => omniauth_payload["provider"],
+      "id_token" => omniauth_payload["credentials"]["id_token"].presence,
+      "provider" => omniauth_payload["provider"].presence,
```
CatalinVoineag, line 23: Having these attributes as nil will redirect the user to the sign in page with a message like You cannot access this service. In centry, there should be an error raised by dfe sign in

CatalinVoineag:
[[Cov] As a User, I should be presented with an authentication error if the DfE Sign In response is invalid](https://trello.com/c/3mfvqtH3/186-cov-as-a-user-i-should-be-presented-with-an-authentication-error-if-the-dfe-sign-in-response-is-invalid)

CatalinVoineag:
[[Cov] As a User, I should be able to sign out](https://trello.com/c/LTkv29qf/185-cov-as-a-user-i-should-be-able-to-sign-out)

elceebee (approved):

app/models/dfe_sign_in_user.rb
Nitemaeric: I'm not sure I follow here. Why do we need to add the `.presence` call in again?

CatalinVoineag: It's for setting the attributes to nil if they are not returned by dfe sign in and this https://github.com/DFE-Digital/itt-mentor-services/blob/faf4e6f9a3a1944c0bfbf603bbaacc99715356bb/app/models/dfe_sign_in_user.rb#L61
Or this or this https://github.com/DFE-Digital/itt-mentor-services/blob/faf4e6f9a3a1944c0bfbf603bbaacc99715356bb/app/models/dfe_sign_in_user.rb#L68

Will return the user to the sign in page. Before, the code would just break here if the omniauth hash was not present

Nitemaeric: Hmm, I've just removed the `.presence` calls and all the tests _still_ pass. Are we sure this is necessary?

Nitemaeric: If it's necessary, can you add a test case that fails without the addition here please?

CatalinVoineag: Yeah, you're right, we don't need the `presence` here

Nitemaeric (approved):
👍

PR #254, opened 2024-02-12
Lb/raise exceptions on create

elceebee:
## Context

Improving test coverage throughout

When we have a "check" method, we check validations have been confirmed before we reach the `create` method. So if the `create` method fails for some reason, we want an exception should be raised. Otherwise, potentially important errors get swallowed up by the application and we won't be aware of them. 

This is related to test coverage because all of our create methods had untested branches. We would have had to manufacture artificial / unrealistic test scenarios to test the branches. 

- Change all "create" methods to remove branching, adding the bang (!) to the save methods (in relevant services or forms)
- Update test cases

## Guidance to review

Tests pass. 

## Link to Trello card

https://trello.com/c/rmWNKYnC

### DIFF
--- a/app/controllers/claims/schools/users_controller.rb
+++ b/app/controllers/claims/schools/users_controller.rb
@@ -18,12 +18,9 @@ class Claims::Schools::UsersController < ApplicationController
   end
 
   def create
-    if user_form.invite
-      redirect_to claims_school_users_path(@school)
-      flash[:success] = t(".user_added")
-    else
-      render :new
-    end
+    user_form.invite!
+    redirect_to claims_school_users_path(@school)
+    flash[:success] = t(".user_added")
   end
 
   private
--- a/app/controllers/claims/schools/users_controller.rb
+++ b/app/controllers/claims/schools/users_controller.rb
@@ -19,8 +19,7 @@ class Claims::Schools::UsersController < ApplicationController
 
   def create
     user_form.invite!
-    redirect_to claims_school_users_path(@school)
-    flash[:success] = t(".user_added")
+    redirect_to claims_school_users_path(@school), flash: { success: t(".user_added") }
   end
 
   private
--- a/app/forms/user_invite_form.rb
+++ b/app/forms/user_invite_form.rb
@@ -8,10 +8,12 @@ class UserInviteForm
   validate :validate_membership
   validates :service, :organisation, presence: true
 
-  def invite
-    return false unless valid?
-
-    UserInviteService.call(user, organisation) if save_user
+  def invite!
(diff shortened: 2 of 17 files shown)


elceebee:
[[Cov] Update all "Add a ..." flows with a "check" page to raise an exception at `create` instead of rendering `new` safely](https://trello.com/c/rmWNKYnC/191-cov-update-all-add-a-flows-with-a-check-page-to-raise-an-exception-at-create-instead-of-rendering-new-safely)

app/forms/user_invite_form.rb
```diff
@@ -8,10 +8,12 @@ class UserInviteForm
...
-  def invite
-    return false unless valid?
-
-    UserInviteService.call(user, organisation) if save_user
+  def invite!
+    ActiveRecord::Base.transaction do
+      user.save!
+      membership.save!
+    end
+    UserInviteService.call(user, organisation)
```
JamieCleare2525, line 16: This should be in the transaction, otherwise the UserInviteService will trigger even if the the user hasn't been saved.

elceebee: It shouldn't.... If you put a debugger in it and call this in a way that raises an error, it won't hit the debugger. It will exit the method where the exception is raised.

JamieCleare2525 (approved):
LGTM! Tested Locally

app/forms/user_invite_form.rb
Nitemaeric: I believe `ActiveRecord::Base.transaction` captures exceptions, rolls back the transaction, and the re-raises the initial exception.

app/controllers/claims/schools/users_controller.rb
```diff
@@ -18,12 +18,9 @@ def check
...
   def create
-    if user_form.invite
-      redirect_to claims_school_users_path(@school)
-      flash[:success] = t(".user_added")
-    else
-      render :new
-    end
+    user_form.invite!
+    redirect_to claims_school_users_path(@school)
+    flash[:success] = t(".user_added")
```
Nitemaeric, line 23: Nitpick and personal preference,

Does inlining these read better for you?

```suggestion
    redirect_to claims_school_users_path(@school), flash: { success: t(".user_added") }
```

This helps me see the flash as data added to the redirection.

Nitemaeric (approved):
👍

app/controllers/claims/schools/users_controller.rb
elceebee: I couldn't care if I tried. Happy to change it if it makes more sense like this to other people.

PR #262, opened 2024-02-13
Lb/dqt integration

elceebee:
## Context

We need to be able to retrieve a teacher form DQT using the teacher's TRN. 

- Dqt Client base class 
- Service class for retrieving a teacher using a trn from the Dqt client
- Env variables added to the appropriate terraform environments, example env file.
- Secrets have been added to MS Azure qa and review app environments

## Guidance to review

Add the required env vars to your local env file (let me know, I'll give you the TEACHING_RECORD_API_KEY for preprod)
Using rails console, retrieve a teacher with the trn "1234567" by calling: `TeachingRecord::GetTeacher.call(trn: "1234567")`. You should get a successful response with a teacher object
Then try calling the service with an invalid trn ("2222222" will work). You should receive an error with status code 404. 

You should be able to replicate this in the review environment as well. 

## Link to Trello card

https://trello.com/c/pohWgS9G

## Things to check

- [x] Required environment variables have been updated or [added to the Azure KeyVault](/docs/environment-variables.md#deploy-pipeline)

## Screenshots
image

### DIFF
--- a/app/services/dqt/client.rb
+++ b/app/services/dqt/client.rb
@@ -1,0 +1,26 @@
+module Dqt
+  class Client
+    class Request
+      include HTTParty
+      base_uri ENV.fetch("DQT_BASE_URL", "")
+      headers "Accept" => "application/json",
+              "Content-Type" => "application/json;odata.metadata=minimal",
+              "Authorization" => "Bearer #{ENV.fetch("DQT_API_KEY", "")}",
+              "X-Api-Version" => ENV.fetch("DQT_API_MINOR_VERSION", "20240101")
+    end
+
+    class HttpError < StandardError; end
+
+    GET_SUCCESS = 200
+
+    def self.get(path)
+      response = Request.get(path)
+
+      if response.code == GET_SUCCESS
+        JSON.parse(response.body || "{}")
+      else
+        raise(HttpError, "status: #{response.code}, body: #{response.body}, headers: #{response.headers}")
+      end
+    end
+  end
+end
--- a/spec/services/dqt/get_teacher_spec.rb
+++ b/spec/services/dqt/get_teacher_spec.rb
@@ -1,0 +1,68 @@
+require "rails_helper"
+
+RSpec.describe Dqt::GetTeacher do
+  context "with invalid trn" do
+    subject(:get_teacher) { described_class.call(trn: "2222222") }
+
+    before do
+      stub_request(:get, "https://preprod.teacher-qualifications-api.education.gov.uk/v3/teachers/2222222")
+        .with(
+          headers: {
+            "Accept" => "application/json",
+            "Accept-Encoding" => "gzip;q=1.0,deflate;q=0.6,identity;q=0.3",
+            "Authorization" => "Bearer secret",
+            "Content-Type" => "application/json;odata.metadata=minimal",
+            "User-Agent" => "Ruby",
+            "X-Api-Version" => "20240101",
+          },
+        )
+        .to_return(
+          status: 404,
+          body: "{\"type\":\"https://tools.ietf.org/html/rfc9110#section-15.5.5\",\"title\":\"Not Found\",\"status\":404,\"traceId\":\"00-dff9d2243466591e882b480c8bdbfc27-f60a1ced105d1602-00\"}",
+          headers: {},
+        )
+    end
+
+    it "raises error" do
+      expect { get_teacher }.to raise_error(Dqt::Client::HttpError)
+    end
+  end
+
+  context "with valid trn" do
+    subject(:get_teacher) { described_class.call(trn: "1234567") }
+
+    before do
+      stub_request(:get, "https://preprod.teacher-qualifications-api.education.gov.uk/v3/teachers/1234567")
+        .with(
+          headers: {
+            "Accept" => "application/json",
+            "Accept-Encoding" => "gzip;q=1.0,deflate;q=0.6,identity;q=0.3",
+            "Authorization" => "Bearer secret",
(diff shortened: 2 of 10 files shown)


elceebee:
[DQT integration](https://trello.com/c/pohWgS9G/197-dqt-integration)

app/services/dqt/client.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    def self.get(path)
+      response = Request.get(path)
+
+      if response.code == GET_SUCCESS
+        JSON.parse(response.body || "{}")
+      else
+        raise(HttpError, "status: #{response.code}, body: #{response.body}, headers: #{response.headers}")
+      end
+    end
```
Nitemaeric, line 24: HTTParty implements some dynamic helpers for checking status codes. This might clean it up a tad:

```suggestion
    def self.get(path)
      response = Request.get(path)

      if response.ok? # Status code == 200 OK - You could also do response.success? which checks for the 2xx range.
        JSON.parse(response.body || "{}")
      else
        raise(HttpError, "status: #{response.code}, body: #{response.body}, headers: #{response.headers}")
      end
    end
```

But furthermore, I'm _slightly_ conscious about the decision to only return the body.

In this specific case, we may have no use of the response headers right now, I'm not sure what headers are returned either. The Ruby net/http _does_ take the approach of `get` for body and `get_response` for the entire response, but something to be aware of. https://docs.ruby-lang.org/en/3.2/Net/HTTP.html#class-Net::HTTP-label-Single-Request+Session

spec/services/dqt/get_teacher_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+            "User-Agent" => "Ruby",
+            "X-Api-Version" => "20240101",
+          },
+        )
+        .to_return(
+          status: 200,
+          body: "{\"trn\":\"1234567\",\"firstName\":\"Judith\",\"middleName\":\"\",\"lastName\":\"Chicken\",\"dateOfBirth\":\"1991-01-22\",\"nationalInsuranceNumber\":\"B15J60R13\",\"email\":\"anonymous@anonymousdomain.org.net.co.uk\",\"qts\":null,\"eyts\":null}",
+          headers: {},
+        )
+    end
```
Nitemaeric, line 51: How do you feel about using https://github.com/vcr/vcr here instead?

app/services/dqt/client.rb
```diff
@@ -0,0 +1,26 @@
+module Dqt
+  class Client
+    class Request
+      include HTTParty
+      base_uri ENV.fetch("DQT_BASE_URL", "")
+      headers "Accept" => "application/json",
+              "Content-Type" => "application/json;odata.metadata=minimal",
+              "Authorization" => "Bearer #{ENV.fetch("DQT_API_KEY", "")}",
+              "X-Api-Version" => ENV.fetch("DQT_API_MINOR_VERSION", "20240101")
```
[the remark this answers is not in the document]
elceebee: Yes, the docs say it's required.

spec/services/dqt/get_teacher_spec.rb
elceebee: I love VCR and prefer it. But I raised it before and there was some objection and it is not the pattern others have adopted.

app/services/dqt/client.rb
elceebee: I was not aware of the `ok?` method, sounds good, I'll implement that.

I think if we have a use case for wanting more than the body, we can implement it, but until then, I don't see an argument for doing that here.

app/services/dqt/client.rb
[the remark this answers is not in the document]
Nitemaeric: Sorry, I mean in terms of the ENV var name. Is `DQT_API_VERSION` as sufficient ENV var name?

I don't know if I would _personally_ call `20240101` a _minor_ version :/ It looks to be just a "version" based on the date. Major-Minor-Patch of semver comes to mind with the terminology of "minor" version. But I understand that is how it is defined in the Teaching Record docs.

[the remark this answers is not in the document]
elceebee: I think we should refer to it as it is referred to in the docs, otherwise we have to explain somewhere that what we are calling DATED_VERSION (or similar) is what is referred to as Minor version in the docs.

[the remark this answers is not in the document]
elceebee: Also the version is actually v3. So this date is not the version exactly -- it's a subversion. Or, as the docs refer to it 'minor' version.

app/services/dqt/client.rb
```diff
@@ -0,0 +1,26 @@
+module Dqt
```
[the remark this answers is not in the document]
elceebee: Resolved in https://github.com/DFE-Digital/itt-mentor-services/pull/262/commits/738df6523b08d69f96b80c3657b30c3a0214f85d

app/services/dqt/client.rb
elceebee: resolved in https://github.com/DFE-Digital/itt-mentor-services/pull/262/commits/738df6523b08d69f96b80c3657b30c3a0214f85d

JamieCleare2525 (approved):
LGTM! Tested locally!

spec/services/dqt/get_teacher_spec.rb
elceebee: I've moved the stubs so that the tests are more readable in this commit https://github.com/DFE-Digital/itt-mentor-services/pull/262/commits/7330320168d39b142cbc561dcad3859df8649ba4

Nitemaeric (approved):
🚀

PR #231, opened 2024-02-08
Claim belongs to provider

CatalinVoineag:
## Context

We decided that a claim can only have one provider. This PR changes the form and the db relation to be a belongs_to, optional because of the way the form works now, we create a claim when landing on on providers form.

## Changes proposed in this pull request

Claim has_one mentor_training
Claim has_one provider
Change claim form
Change suppport claim show  

## Guidance to review

Go onto the review app and create a claim
Then view that claim as a support user

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/5a14e3ff-5119-4c6b-963c-558c59c01347

PR #268, opened 2024-02-14
Cv/add mentor memberships

CatalinVoineag:
## Context

This is an interation on https://github.com/DFE-Digital/itt-mentor-services/pull/251

Add mentor memberships eeaa8a4c73311de9e6be1d6a3e0010d35f9d913b
We want to allow our mentors to belongs to different schools at the same
time. This commit adds a mentor_membership table that will link a mentor
to a school

Remove school_id from mentor 77359cdd0aa6c89718e53b550d44201e7118d008
We use the mentor_membership table to link a mentor to a school, so the
school_id is not necessary on the mentor table

The mentor membership table should help us do things like
```Ruby
claims_school     = Claims::School.first
placements_school = Placements::School.first

mentor = Mentor.first # Seeds already have some _fake_ mentor data

claims_mentor_membership     = Claims::MentorMembership.create(mentor:, school: claims_school)
placements_mentor_membership = Placements::MentorMembership.create(mentor:, school: placements_school)

Mentor.all             # => [mentor]
Claims::Mentor.all     # => [mentor.becomes(Claims::Mentor)]
Placements::Mentor.all # => [mentor.becomes(Placements::Mentor)]

Mentor.first.mentor_memberships             # => [claims_mentor_membership, placements_mentor_membership]
Claims::Mentor.first.mentor_memberships     # => [claims_mentor_membership]
Placements::Mentor.first.mentor_memberships # => [placements_mentor_membership]

Mentor.first.schools             # => [claims_school, placements_school]
Claims::Mentor.first.schools     # => [claims_school]
Placements::Mentor.first.schools # => [placements_school]

School.first.mentors             # => [mentor]
Claims::School.first.mentors     # => [mentor.becomes(Claims::Mentor)]
Placements::School.first.mentors # => [mentor.becomes(Placements::Mentor)]
```

- Create mentor_membership table
- Fix claim form specs
- Remove school_id from mentor table(no longer needed)

## Guidance to review

- It might be easier to review the commits individually
- If you pull this on your local, run the seed file again
- Go and create a claim through the claim form, local or on review app

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/73b0e5cc-f896-4011-a17a-520610726210

### DIFF
--- a/app/models/mentor_membership.rb
+++ b/app/models/mentor_membership.rb
@@ -1,0 +1,29 @@
+# == Schema Information
+#
+# Table name: mentor_memberships
+#
+#  id         :uuid             not null, primary key
+#  type       :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#  mentor_id  :uuid             not null
+#  school_id  :uuid             not null
+#
+# Indexes
+#
+#  index_mentor_memberships_on_mentor_id                         (mentor_id)
+#  index_mentor_memberships_on_school_id                         (school_id)
+#  index_mentor_memberships_on_type_and_mentor_id                (type,mentor_id)
+#  index_mentor_memberships_on_type_and_school_id_and_mentor_id  (type,school_id,mentor_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (mentor_id => mentors.id)
+#  fk_rails_...  (school_id => schools.id)
+#
+class MentorMembership < ApplicationRecord
+  belongs_to :mentor
+  belongs_to :school
+
+  validates :mentor_id, uniqueness: { scope: %i[type school_id] }
+end
--- a/app/models/placements/mentor.rb
+++ b/app/models/placements/mentor.rb
@@ -8,24 +8,10 @@
 #  trn        :string           not null
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
-#  school_id  :uuid             not null
 #
-# Indexes
-#
-#  index_mentors_on_school_id  (school_id)
-#
-# Foreign Keys
-#
-#  fk_rails_...  (school_id => schools.id)
-#
-class Mentor < ApplicationRecord
-  belongs_to :school
+class Placements::Mentor < Mentor
   has_many :mentor_memberships
   has_many :schools, through: :mentor_memberships
 
-  validates :first_name, :last_name, :trn, presence: true
-
-  def full_name
-    "#{first_name} #{last_name}".strip
-  end
+  default_scope { joins(:mentor_memberships).distinct }
 end
--- a/db/migrate/20240213151647_create_mentor_memberships.rb
+++ b/db/migrate/20240213151647_create_mentor_memberships.rb
@@ -1,0 +1,13 @@
+class CreateMentorMemberships < ActiveRecord::Migration[7.1]
+  def change
+    create_table :mentor_memberships, id: :uuid do |t|
+      t.string :type, null: false
+      t.references :mentor, null: false, foreign_key: true, type: :uuid
+      t.references :school, null: false, foreign_key: true, type: :uuid
+      t.index %i[type school_id mentor_id], unique: true
+      t.index %i[type mentor_id]
+
+      t.timestamps
+    end
+  end
+end
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -73,17 +73,12 @@ schools.each do |school|
 end
 
 # Create dummy mentors
-schools = School.where(claims_service: true)
-  .or(School.where(placements_service: true))
-schools.each do |school|
-  %w[Sarah John Pomona].each_with_index do |first_name, index|
-    Mentor.find_or_create_by(
-      first_name:,
-      last_name: "Doe",
-      school:,
-      trn: index,
-    )
-  end
+mentors = %w[Sarah John Pomona].each_with_index.map do |first_name, index|
+  Mentor.find_or_create_by!(first_name:, last_name: "Doe", trn: index)
+end
+
+(Claims::School.all + Placements::School.all).each do |school|
+  school.mentors = mentors
 end
 
 Rails.logger.debug "Organisations assigned to users"
--- a/spec/factories/mentor_memberships.rb
+++ b/spec/factories/mentor_memberships.rb
@@ -1,0 +1,45 @@
+# == Schema Information
+#
+# Table name: mentor_memberships
+#
+#  id         :uuid             not null, primary key
+#  type       :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#  mentor_id  :uuid             not null
+#  school_id  :uuid             not null
+#
+# Indexes
+#
+#  index_mentor_memberships_on_mentor_id                         (mentor_id)
+#  index_mentor_memberships_on_school_id                         (school_id)
+#  index_mentor_memberships_on_type_and_mentor_id                (type,mentor_id)
+#  index_mentor_memberships_on_type_and_school_id_and_mentor_id  (type,school_id,mentor_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (mentor_id => mentors.id)
+#  fk_rails_...  (school_id => schools.id)
+#
+FactoryBot.define do
+  factory :mentor_membership do
+    association :school
+    association :mentor
+
+    trait :claims do
+      type { Claims::MentorMembership }
+    end
+
+    trait :placements do
+      type { Placements::MentorMembership }
+    end
+
+    factory :claims_mentor_membership,
+            class: "Claims::MentorMembership",
+            parent: :mentor_membership
+
+    factory :placements_mentor_membership,
+            class: "Placements::MentorMembership",
+            parent: :mentor_membership
+  end
+end
--- a/spec/factories/mentors.rb
+++ b/spec/factories/mentors.rb
@@ -8,24 +8,14 @@
 #  trn        :string           not null
 #  created_at :datetime         not null
 #  updated_at :datetime         not null
-#  school_id  :uuid             not null
-#
-# Indexes
-#
-#  index_mentors_on_school_id  (school_id)
-#
-# Foreign Keys
-#
-#  fk_rails_...  (school_id => schools.id)
 #
 FactoryBot.define do
   factory :mentor do
-    association :school
-
     first_name { Faker::Name.first_name }
     last_name { Faker::Name.last_name }
     sequence(:trn) { _1 }
   end
+
   factory :claims_mentor,
           class: "Claims::Mentor",
           parent: :mentor
--- a/spec/factories/mentors.rb
+++ b/spec/factories/mentors.rb
@@ -26,4 +26,11 @@ FactoryBot.define do
     last_name { Faker::Name.last_name }
     sequence(:trn) { _1 }
   end
+  factory :claims_mentor,
+          class: "Claims::Mentor",
+          parent: :mentor
+
+  factory :placements_mentor,
+          class: "Placements::Mentor",
+          parent: :mentor
 end
--- a/spec/models/claims/mentor_membership_spec.rb
+++ b/spec/models/claims/mentor_membership_spec.rb
@@ -1,0 +1,53 @@
+# == Schema Information
+#
+# Table name: mentor_memberships
+#
+#  id         :uuid             not null, primary key
+#  type       :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#  mentor_id  :uuid             not null
+#  school_id  :uuid             not null
+#
+# Indexes
+#
+#  index_mentor_memberships_on_mentor_id                         (mentor_id)
+#  index_mentor_memberships_on_school_id                         (school_id)
+#  index_mentor_memberships_on_type_and_mentor_id                (type,mentor_id)
+#  index_mentor_memberships_on_type_and_school_id_and_mentor_id  (type,school_id,mentor_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (mentor_id => mentors.id)
+#  fk_rails_...  (school_id => schools.id)
+#
+require "rails_helper"
+
+RSpec.describe Claims::MentorMembership, type: :model do
+  context "with associations" do
+    it { is_expected.to belong_to(:mentor) }
+    it { is_expected.to belong_to(:school) }
+
+    it "returns Claims::Mentor when calling #mentor" do
+      mentor = create(:claims_mentor)
+      mentor_membership = create(
+        :mentor_membership,
+        :claims,
+        mentor:,
+      )
+
+      expect(mentor_membership.mentor).to be_a(Claims::Mentor)
+    end
+
+    it "returns Claims::School when calling #school" do
+      school = create(:claims_school)
+      mentor_membership = create(
+        :mentor_membership,
+        :claims,
+        school:,
+      )
+
+      expect(mentor_membership.school).to be_a(Claims::School)
+    end
+  end
+end
--- a/spec/models/claims/mentor_spec.rb
+++ b/spec/models/claims/mentor_spec.rb
@@ -1,0 +1,39 @@
+# == Schema Information
+#
+# Table name: mentors
+#
+#  id         :uuid             not null, primary key
+#  first_name :string           not null
+#  last_name  :string           not null
+#  trn        :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+require "rails_helper"
+
+RSpec.describe Claims::Mentor, type: :model do
+  context "with associations" do
+    it { is_expected.to have_many(:mentor_memberships) }
+    it { is_expected.to have_many(:schools).through(:mentor_memberships) }
+  end
+
+  describe "default scope" do
+    it "returns only mentors with mentor memberships on the claims service" do
+      mentor = create(:claims_mentor)
+      create(
+        :mentor_membership,
+        :claims,
+        mentor:,
+      )
+      placements_mentor = create(:placements_mentor)
+      create(
+        :mentor_membership,
+        :placements,
+        mentor: placements_mentor,
+      )
+
+      expect(described_class.all).to contain_exactly(mentor)
+      expect(described_class.all).not_to include(placements_mentor)
+    end
+  end
+end
--- a/spec/models/placements/mentor_membership_spec.rb
+++ b/spec/models/placements/mentor_membership_spec.rb
@@ -1,0 +1,53 @@
+# == Schema Information
+#
+# Table name: mentor_memberships
+#
+#  id         :uuid             not null, primary key
+#  type       :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#  mentor_id  :uuid             not null
+#  school_id  :uuid             not null
+#
+# Indexes
+#
+#  index_mentor_memberships_on_mentor_id                         (mentor_id)
+#  index_mentor_memberships_on_school_id                         (school_id)
+#  index_mentor_memberships_on_type_and_mentor_id                (type,mentor_id)
+#  index_mentor_memberships_on_type_and_school_id_and_mentor_id  (type,school_id,mentor_id) UNIQUE
+#
+# Foreign Keys
+#
+#  fk_rails_...  (mentor_id => mentors.id)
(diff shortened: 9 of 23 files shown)


Nitemaeric:
Is there a reason for not using the base code available in https://github.com/DFE-Digital/itt-mentor-services/pull/251?

CatalinVoineag:
> Is there a reason for not using the base code available in #251?

I think it was just easier for me to not copy the code as it was. This caused some issues for example https://github.com/DFE-Digital/itt-mentor-services/blob/b2e83ea9371b73c9c28e8d291a5753aede67d4d6/app/models/claims/mentor_membership.rb#L23-L24
It returns all the mentors of any school not just of a specific school. So I used it more as an inspiration.

Nitemaeric:
@CatalinVoineag Can you validate that these all return the same?

```ruby
claims_school     = Claims::School.first
placements_school = Placements::School.first

mentor = Mentor.first # Seeds already have some _fake_ mentor data

claims_mentor_membership     = Claims::MentorMembership.create(mentor:, school: claims_school)
placements_mentor_membership = Placements::MentorMembership.create(mentor:, school: placements_school)

Mentor.all             # => [mentor]
Claims::Mentor.all     # => [mentor.becomes(Claims::Mentor)]
Placements::Mentor.all # => [mentor.becomes(Placements::Mentor)]

Mentor.first.mentor_memberships             # => [claims_mentor_membership, placements_mentor_membership]
Claims::Mentor.first.mentor_memberships     # => [claims_mentor_membership]
Placements::Mentor.first.mentor_memberships # => [placements_mentor_membership]

Mentor.first.schools             # => [claims_school, placements_school]
Claims::Mentor.first.schools     # => [claims_school]
Placements::Mentor.first.schools # => [placements_school]

School.first.mentors             # => [mentor]
Claims::School.first.mentors     # => [mentor.becomes(Claims::Mentor)]
Placements::School.first.mentors # => [mentor.becomes(Placements::Mentor)]
```

Nitemaeric:
> It returns all the mentors of any school not just of a specific school. So I used it more as an inspiration.

Also, what code did you run to get that invalid result?

CatalinVoineag:
> > It returns all the mentors of any school not just of a specific school. So I used it more as an inspiration.
> 
> Also, what code did you run to get that invalid result?

Sorry, I got things mixed up. So [these two](https://github.com/DFE-Digital/itt-mentor-services/blob/b2e83ea9371b73c9c28e8d291a5753aede67d4d6/app/models/school.rb#L61-L62) lines return duplicate instances of the mentors when calling `Claims::School.last.mentors.count
`

This is based on the seed file, which should create 3 mentors and 9 mentor Memberships, 3 for each school. But having the mentor_membership repeated in the `Claims::School` generates a very strange query.

Nitemaeric:
@CatalinVoineag I know the issue now. We can add a `distinct` call to the default scope. This is because it is performing a join and finding multiple joined records.

In the mentor_memberships table, we have the following:

| mentor_id | school_id |
| --------: | --------: |
|         1 |         1 |
|         1 |         2 |
|         1 |         3 |
|         2 |         1 |
|         2 |         2 |
|         2 |         3 |
|         3 |         1 |
|         3 |         2 |
|         3 |         3 |

The query that is run with `Claims::Mentor.all` will look _something_ like this:

```sql
SELECT *
FROM mentors
INNER JOIN mentor_memberships ON mentor_memberships.mentor_id = mentors.id
```

returns the following results (removed some columns for simplicity)

| mentors.id | mentor_memberships.mentor_id | mentor_memberships.school_id |
| ---------: | ---------------------------: | ---------------------------: |
|          1 |                            1 |                            1 |
|          1 |                            1 |                            2 |
|          1 |                            1 |                            3 |
|          2 |                            2 |                            1 |
|          2 |                            2 |                            2 |
|          2 |                            2 |                            3 |
|          3 |                            3 |                            1 |
|          3 |                            3 |                            2 |
|          3 |                            3 |                            3 |

Rails then only cares about mapping the rows to instances, so it instantiates the following:

```ruby
[
  #<Mentor id: 1>,
  #<Mentor id: 1>,
  #<Mentor id: 1>,
  #<Mentor id: 2>,
  #<Mentor id: 2>,
  #<Mentor id: 2>,
  #<Mentor id: 3>,
  #<Mentor id: 3>,
  #<Mentor id: 3>,
]
```

9 results.

To de-duplicate the results, we can add a "distinct" call to the SQL query.

```ruby
class Claims::Mentor < Mentor
  default_scope { joins(:mentor_memberships).distinct }
end
```

Which will perform:

```sql
SELECT DISTINCT *
FROM ...
``

SQL will then de-dup any non-unique records, resulting in:

| mentors.id | mentor_memberships.mentor_id | mentor_memberships.school_id |
| ---------: | ---------------------------: | ---------------------------: |
|          1 |                            1 |                            1 |
|          2 |                            2 |                            1 |
|          3 |                            3 |                            1 |

and consequently:

```ruby
[
  #<Mentor id: 1>,
  #<Mentor id: 2>,
  #<Mentor id: 3>,
]
```

Give it a go! 😄

CatalinVoineag:
[Add Mentor Memberships](https://trello.com/c/mIcgfYjg/192-add-mentor-memberships)

app/models/mentor_membership.rb
```diff
@@ -0,0 +1,25 @@
...
+#
+#  index_mentor_memberships_on_mentor_id  (mentor_id)
+#  index_mentor_memberships_on_school_id  (school_id)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (mentor_id => mentors.id)
+#  fk_rails_...  (school_id => schools.id)
+#
+class MentorMembership < ApplicationRecord
```
ollietreend, line 22: Should we be including some sort of database constraint and/or model validation to restrict multiple identical memberships being created?

Right now I can do this:

```ruby
mentor = Mentor.first
school = Placements::School.first
5.times { school.mentors << mentor }
mentor.mentor_memberships.count # => 5
mentor.schools.count { |s| s.id == school.id } # => 5
```

A mentor can be added to the same school multiple times.

db/seeds.rb
```diff
@@ -73,17 +73,12 @@
...
-      trn: index,
-    )
-  end
+mentors = %w[Sarah John Pomona].each_with_index.map do |first_name, index|
+  Mentor.find_or_create_by!(first_name:, last_name: "Doe", trn: index)
+end
+
+Claims::School.find_each do |school|
+  school.mentors = mentors
 end
```
ollietreend, line 82: This would be handy for School Placements, too. 😇🙏🏻 

```suggestion
(Claims::School.all + Placements::School.all).each do |school|
  school.mentors = mentors
end
```

spec/factories/mentor_memberships.rb
```diff
@@ -0,0 +1,35 @@
...
+    association :school
+    association :mentor
+
+    trait :claims do
+      type { "Claims::MentorMembership" }
+    end
+
+    trait :placements do
+      type { "Placements::MentorMembership" }
+    end
```
ollietreend, line 33: What about referencing the class itself? Rails accepts this for STI types – it doesn't have to be given a String.

```suggestion
    trait :claims do
      type { Claims::MentorMembership }
    end

    trait :placements do
      type { Placements::MentorMembership }
    end
```

Presumably this would give us some added intellisense in our IDEs. At the very least we'll be able to `Cmd/Ctrl + click` on the class name to jump to the definition.

No biggie, though. Feel free to skip this one if you don't think it's a valuable change.

spec/factories/mentors.rb
```diff
@@ -8,22 +8,19 @@
...
     sequence(:trn) { _1 }
   end
+
+  factory :claims_mentor,
+          class: "Claims::Mentor",
+          parent: :mentor
+
+  factory :placements_mentor,
+          class: "Placements::Mentor",
+          parent: :mentor
```
ollietreend, line 25: Likewise, [this will work](https://thoughtbot.github.io/factory_bot/ref/factory.html#options):

```suggestion
  factory :claims_mentor,
          class: Claims::Mentor,
          parent: :mentor

  factory :placements_mentor,
          class: Placements::Mentor,
          parent: :mentor
```

spec/models/claims/mentor_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rails_helper"
+
+RSpec.describe Claims::Mentor, type: :model do
+  context "with associations" do
+    it { is_expected.to have_many(:mentor_memberships) }
+    it { is_expected.to have_many(:schools).through(:mentor_memberships) }
+  end
+
+  describe "default scope" do
+    it "is scoped to providers using the placement service" do
```
ollietreend, line 10: Typo:

```suggestion
    it "is scoped to providers using the claims service" do
```

spec/models/claims/mentor_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      mentor = create(:claims_mentor)
+      create(
+        :mentor_membership,
+        :claims,
+        mentor:,
+      )
+
+      expect(described_class.all).to contain_exactly(mentor)
+    end
```
ollietreend, line 19: It'd be good to include a mentor which _should not_ appear in the results, too. Otherwise we're not demonstrating that non-claims mentors are filtered out by the default scope.

Here's an example of how a scope has been tested elsewhere:

https://github.com/DFE-Digital/itt-mentor-services/blob/9f7555d0065622ac4b3e97ade992b2251b6beb08/spec/models/provider_spec.rb#L55-L63

spec/models/claims/mentor_membership_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "rails_helper"
+
+RSpec.describe Claims::MentorMembership, type: :model do
+  context "with associations" do
+    it { is_expected.to belong_to(:mentor) }
+    it { is_expected.to belong_to(:school) }
```
ollietreend, line 6: Would it be valuable to test/show that the returned associations are all within the `Claims::` namespace?

I think that's the reason why we re-declare the association in each sub-class, so that they pick up the namespace. Right? A test here would cover the re-defining of those associations.

e.g. so `Claims::MentorMembership#school` will always return a `Claims::School`, and `Claims::MentorMembership#mentor` will always return a `Claims::Mentor`.

spec/models/placements/mentor_membership_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "rails_helper"
+
+RSpec.describe Placements::MentorMembership, type: :model do
+  context "with associations" do
+    it { is_expected.to belong_to(:mentor) }
+    it { is_expected.to belong_to(:school) }
```
ollietreend, line 6: Ditto with this, re: module namespacing.

spec/models/placements/mentor_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    it "is scoped to placements mentors" do
+      mentor = create(:placements_mentor)
+      create(
+        :mentor_membership,
+        :placements,
+        mentor:,
+      )
+
+      expect(described_class.all).to contain_exactly(mentor)
+    end
```
ollietreend, line 19: Ditto, re: testing the scope excludes non-placements mentors

db/migrate/20240213151647_create_mentor_memberships.rb
```diff
@@ -0,0 +1,11 @@
+class CreateMentorMemberships < ActiveRecord::Migration[7.1]
+  def change
+    create_table :mentor_memberships, id: :uuid do |t|
+      t.string :type
+      t.references :mentor, null: false, foreign_key: true, type: :uuid
+      t.references :school, null: false, foreign_key: true, type: :uuid
```
ollietreend, line 6: I think we should include some database indexes in here, too.

What about a composite index like this:

```ruby
t.index %i[type school_id mentor_id], unique: true
```

This should support the read patterns:

- Find all mentors associated with a school (using a partial index)
- Enforce unique membership constraint (using the full 'unique' index)

And then another index for mentors scoped to a specific service:

```ruby
t.index %i[type mentor_id]
```

Which would support the read pattern:

- Find all schools associated with a mentor

ollietreend (changes requested):
Really nice work @CatalinVoineag! ⭐ This is looking good.

I've added comments with some hopefully small-ish things.

And then there are a couple of comments which I think need attention. For that reason, I'm marking this review as "request changes". These are the comments about unique membership validation and/or database indexes and constraints.

spec/factories/mentors.rb
CatalinVoineag: Rubocop will complain about this, but I can skip this rule in this file I think

app/models/mentor_membership.rb
CatalinVoineag: This will be fixed by https://github.com/DFE-Digital/itt-mentor-services/pull/268#discussion_r1489905897

db/migrate/20240213151647_create_mentor_memberships.rb
```diff
@@ -0,0 +1,11 @@
+class CreateMentorMemberships < ActiveRecord::Migration[7.1]
+  def change
+    create_table :mentor_memberships, id: :uuid do |t|
+      t.string :type
```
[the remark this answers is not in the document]
CatalinVoineag: Fixed in c06a141d8659b694ad202fd79569d29eaeceaf2c

app/models/mentor_membership.rb
CatalinVoineag: Fixed in c06a141d8659b694ad202fd79569d29eaeceaf2c

db/seeds.rb
CatalinVoineag: c06a141d8659b694ad202fd79569d29eaeceaf2c

spec/factories/mentor_memberships.rb
CatalinVoineag: c06a141d8659b694ad202fd79569d29eaeceaf2c

spec/models/claims/mentor_spec.rb
CatalinVoineag: c06a141d8659b694ad202fd79569d29eaeceaf2c

spec/models/claims/mentor_spec.rb
CatalinVoineag: c06a141d8659b694ad202fd79569d29eaeceaf2c

spec/models/placements/mentor_spec.rb
CatalinVoineag: c06a141d8659b694ad202fd79569d29eaeceaf2c

db/migrate/20240213151647_create_mentor_memberships.rb
CatalinVoineag: c06a141d8659b694ad202fd79569d29eaeceaf2c

spec/models/placements/mentor_membership_spec.rb
CatalinVoineag: c06a141d8659b694ad202fd79569d29eaeceaf2c

app/models/placements/mentor.rb
```diff
@@ -0,0 +1,16 @@
...
+#  last_name  :string           not null
+#  trn        :string           not null
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+class Placements::Mentor < Mentor
+  default_scope { joins(:mentor_memberships).distinct }
+  has_many :mentor_memberships
+  has_many :schools, through: :mentor_memberships
+end
```
Nitemaeric, line 16: This might be OCD, but:

```ruby
class Claims::Mentor < Mentor
  has_many :mentor_memberships
  has_many :schools, through: :mentor_memberships

  default_scope { joins(:mentor_memberships).distinct }
end
```

😆

db/migrate/20240214090208_remove_provider_id_from_mentor_training.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveProviderIdFromMentorTraining < ActiveRecord::Migration[7.1]
+  def change
+    remove_column :mentor_trainings, :provider_id, :uuid
```
Nitemaeric, line 3: To avoid coupling mentor trainings to claims, could we keep this field in? This way we can get the context of which mentor trainings happened between which providers and mentors without the need to involve claims made.

spec/factories/mentor_memberships.rb
```diff
@@ -0,0 +1,37 @@
...
+    association :mentor
+
+    trait :claims do
+      type { Claims::MentorMembership }
+    end
+
+    trait :placements do
+      type { Placements::MentorMembership }
+    end
+  end
```
Nitemaeric, line 36: Could we add `:claims_mentor_membership` and `:placements_mentor_membership` factories too please?

Most of the other factories are called with the prefix, rather than the trait.

db/migrate/20240214090208_remove_provider_id_from_mentor_training.rb
CatalinVoineag: We can but why did we want to say a claim belongs to a provider? https://github.com/DFE-Digital/itt-mentor-services/pull/231

Right now, we technically have two ways of linking a claim and a provider.

Do you want me to revert this then? https://github.com/DFE-Digital/itt-mentor-services/pull/231

Nitemaeric: A claim creates a requirement of all MentorTrainings be scoped to a single `provider_id` by adding `Claim#provider_id`.

The concept of a `MentorTraining` can be standalone.

Here's a link to our discussion on Slack: https://ukgovernmentdfe.slack.com/archives/C06CUPM2HR8/p1707403465382709

app/models/placements/mentor.rb
ollietreend: @Nitemaeric possibly a bit 😂 

But it's worth pointing out that there's a RuboCop rule specifically for this: [Layout/ClassStructure](https://docs.rubocop.org/rubocop/cops_layout.html#layoutclassstructure)

If you think the structure of this class is worth changing, would it be worth switching on this rule so there's a consistent style across the board? (presumably out of scope for this PR though)

db/migrate/20240214090208_remove_provider_id_from_mentor_training.rb
CatalinVoineag: OK, I'll keep the provider_id on the mentor training

app/models/mentor_membership.rb
ollietreend: @CatalinVoineag what do you think about adding a model validation rule, too? So we catch any uniqueness violations in the app/validation layer rather than as an exception from the database.

```ruby
validates_uniqueness_of :mentor_id, scope: [:type, :school_id]
```

app/models/placements/mentor.rb
Nitemaeric: The RuboCop is definitely for another PR :)

Happy to keep them different, but it just jumped out when going through the code.

app/models/mentor_membership.rb
CatalinVoineag: 0dfcc333083a61294db041c68cc2b7f4627cf2cb

db/migrate/20240214090208_remove_provider_id_from_mentor_training.rb
CatalinVoineag: Removed the commit that was removing the provider id

spec/factories/mentor_memberships.rb
CatalinVoineag: 0dfcc333083a61294db041c68cc2b7f4627cf2cb

app/models/placements/mentor.rb
CatalinVoineag: 0dfcc333083a61294db041c68cc2b7f4627cf2cb

spec/models/placements/mentor_membership_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+# Foreign Keys
+#
+#  fk_rails_...  (mentor_id => mentors.id)
+#  fk_rails_...  (school_id => schools.id)
+#
+require "rails_helper"
+
+RSpec.describe Placements::MentorMembership, type: :model do
+  context "with associations" do
+    it { is_expected.to belong_to(:mentor) }
```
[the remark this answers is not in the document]
Nitemaeric: It performs a `class_name.constantize.find(id)`, which passes with both `Placements::Mentor` and `Mentor`. It _does_ fail with `Claims::Mentor` here though.

Nitemaeric (approved):

PR #269, opened 2024-02-14
Add view component previews and tests for persona sign in component

JamieCleare2525:
## Context

Add tests for PersonaSignInComponent.

- Setup ViewComponent Previews
- Add unit tests for PersonaSignInComponent

## Guidance to review

- Visit http://placements.localhost:3000/rails/view_components
- Click through the links to view previews of the view components

## Link to Trello card

https://trello.com/c/DQ1ixmF7/206-add-tests-for-personasignincomponent-to-increase-coverage

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/17162488/e0ca3728-540a-4cc7-b7a2-1f9d37609466


spec/components/previews/persona_sign_in_component_preview.rb
```diff
@@ -0,0 +1,27 @@
...
+  def support_user_colin
+    persona = FactoryBot.build(:placements_support_user, :colin)
+    render_with_template(template:, locals: { persona: })
+  end
+
+  private
+
+  def template
+    "templates/persona_sign_in_component_preview"
+  end
```
Nitemaeric, line 26: Would it be more simple / less code to follow the default directory?

https://viewcomponent.org/guide/previews.html#preview-templates

> Given a preview test/components/previews/cell_component_preview.rb, template files can be defined at test/components/previews/cell_component_preview/:

Nitemaeric (commented):
Previews! ❤️

spec/components/previews/persona_sign_in_component_preview.rb
JamieCleare2525: So for that example it works because the method is `default` and the template is `test/components/previews/cell_component_preview/default.html.erb`.
Which means you need a different template per-method.
In my implementation, the templates are stored in a `templates` folder, and you only need to define the one template (i.e `templates/persona_sign_in_component_preview`).
This is only necessary in this case because for the component to look correct it needs some additional html. This will not be the case for every component preview.

elceebee (approved):

spec/components/previews/persona_sign_in_component_preview.rb
Nitemaeric: I see what you mean. We'd need a copy of our template _per_ preview.

Hmm, initially, it feels like we're not _directly_ interacting with the component for the previews.

But it doesn't really matter either.

Nitemaeric (approved):

PR #279, opened 2024-02-19
Upgrade Ruby to 3.3.0

CatalinVoineag:
## Context

We want to use the new chars keyword argument in the SecureRandom#alphanumeric method introduced in Ruby 3.3.

Also updated bundler to 2.5.6

Gemfile
Dockerfile
Github actions

## Guidance to review

If you're using asdf, you will need to update the ruby plugin `asdf plugin update ruby`


CatalinVoineag:
[Update Ruby to 3.3](https://trello.com/c/QQ1yRDwG/221-update-ruby-to-33)

.ruby-version
```diff
@@ -1 +1 @@
-ruby-3.2.2
+ruby-3.3.0
```
Nitemaeric, line 1: `rbenv` seems to fail to automatically detect the version with this prefix.

```suggestion
3.3.0
```

Running `rbenv install` raised the following error:

```
❯ rbenv install
ruby-build: definition not found: ruby-3.3.0

See all available versions with `rbenv install --list-all'.

If the version you need is missing, try upgrading ruby-build:

  brew upgrade ruby-build
```

Gemfile.lock
```diff
@@ -660,7 +660,7 @@ DEPENDENCIES
   webmock
 
 RUBY VERSION
-   ruby 3.2.2p53
+   ruby 3.3.0p0
 
 BUNDLED WITH
    2.4.22
```
Nitemaeric, line 666: After installing 3.3.0, my local bundler version was also updated. I wonder if we should update the bundler version here as well? `bundle update --bundler`. Likely worth a separate commit, potentially a separate PR.

Nitemaeric (approved):
Updated `Gemfile.lock` and app work without any issues.

I did run into an issue with `.ruby-version` though. Running `rbenv local 3.3.0`, it updated the `.ruby-version` content to `3.3.0` locally.

Gemfile.lock
CatalinVoineag: I'll do it in a separate commit

elceebee (approved):

PR #296, opened 2024-02-20
Refactor: Create a single source of truth for `current_service`

ollietreend:
## Context

This PR is a refactor which aims to clarify and simplify the way `current_service` is passed and defined within the app.

I'm doing this as a precursory refactor ahead of [setting up GoodJob](https://trello.com/c/9uy2ArZJ/217-draft-set-up-active-job). I'm doing this so I can access the current user within `routes.rb` – which has access to `request.session`, but cannot access the `current_user` controller helper method – so a suitable constraint can be applied for accessing the GoodJob admin dashboard.

It makes two main changes:

1. Define a single source of truth – by removing a duplicate implementation of `current_service`.
2. Stop passing `current_service` into the user's session – it was unnecessary, and was the source of a potential bug.

## Guidance to review

This is a refactor with no functional changes, so there's not much to review. The most important thing is that tests still pass.

If you want more details about the changes made, I recommend reading the commit messages.

## Link to Trello card

This is a precursory refactor for: https://trello.com/c/9uy2ArZJ/217-draft-set-up-active-job

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,12 +2,7 @@ module ApplicationHelper
   include Pagy::Frontend
 
   def current_service
-    case request.host
-    when ENV["CLAIMS_HOST"]
-      :claims
-    when ENV["PLACEMENTS_HOST"]
-      :placements
-    end
+    HostingEnvironment.current_service(request)
   end
 
   def service_name
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,11 +13,7 @@ class ApplicationController < ActionController::Base
   private
 
   def sign_in_user
-    @sign_in_user ||=
-      begin
-        session["service"] = current_service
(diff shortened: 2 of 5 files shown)


elceebee (approved):
Nice!

CatalinVoineag (approved):

app/helpers/application_helper.rb
```diff
@@ -2,12 +2,7 @@ module ApplicationHelper
   include Pagy::Frontend
 
   def current_service
-    case request.host
-    when ENV["CLAIMS_HOST"]
-      :claims
-    when ENV["PLACEMENTS_HOST"]
-      :placements
-    end
+    HostingEnvironment.current_service(request)
```
Nitemaeric, line 5: I got to this too slow, but I'm not sure this logic belongs in the `HostingEnvironment` class.

I think it might be more appropriate for us to add a `Service` class/module even.

We can then add maybe `Service.extract_from_request(request)`.

PR #297, opened 2024-02-20
Set up GoodJob for Active Job

ollietreend:
## Context

We need to be able to run asynchronous background tasks and scheduled jobs. To do this, we can use [GoodJob](https://github.com/bensheldon/good_job) which provides a backend adapter for [Active Job](https://guides.rubyonrails.org/active_job_basics.html).

GoodJob is similar to [Sidekiq](https://github.com/sidekiq/sidekiq) except it's backed by the existing Postgres database that the app already depends on. It keeps our stack simple by avoiding the need to set up Redis instances, as Sidekiq would require.

GoodJob also supports [cron-style scheduled jobs](https://github.com/bensheldon/good_job?tab=readme-ov-file#cron-style-repeatingrecurring-jobs) out of the box, akin to what you'd get from adding [sidekiq-scheduler](https://github.com/sidekiq-scheduler/sidekiq-scheduler) or [sidekiq-cron](https://github.com/sidekiq-cron/sidekiq-cron) on top of a Sidekiq setup.

- Set up GoodJob and configure Active Job to use it
- Enable cron functionality so we can define scheduled jobs
- Create `config/scheduled_jobs.yml` with an example job definition
- Mount the GoodJob admin dashboard ([demo](https://goodjob-demo.herokuapp.com/)) at `/good_job` – this is only accessible to support users

### In development mode

GoodJob will execute jobs asynchronously in the main `rails server` process. There's no need to run an external worker process if the `rails server` is running.

Scheduled jobs will not run in development mode.

### In production mode

A GoodJob worker process needs to be run for jobs to execute. They will not be run within the main server process.

Scheduled jobs will run in production.

Start a worker process by running:

```
bundle exec good_job start
```

## Guidance to review

This is probably easiest reviewed commit-by-commit. I've tried to chunk up each commit logically, and add descriptive commit messages where I think they'd be helpful.

There are no Active Job jobs defined in our app yet, so there isn't anything to use if you wanted to test it locally. I did have a dummy job which I used while working on this PR, but I haven't committed it into this branch.

## Link to Trello card

https://trello.com/c/9uy2ArZJ/217-set-up-active-job

### DIFF
--- a/config/initializers/good_job.rb
+++ b/config/initializers/good_job.rb
@@ -1,0 +1,21 @@
+# Configure GoodJob, an Active Job adapter backed by the Postgres database
+# More details at: https://github.com/bensheldon/good_job
+
+Rails.application.configure do
+  # In development, execute jobs within the main `rails server` process
+  config.good_job.execution_mode = :async if Rails.env.development?
+
+  # In production, execute jobs in a separate worker process (i.e. `good_job start`)
+  config.good_job.execution_mode = :external if Rails.env.production?
+
+  # Read scheduled jobs from config file
+  scheduled_jobs = YAML.load_file(
+    Rails.root.join("config/scheduled_jobs.yml"),
+    symbolize_names: true,
+  )
+
+  # Enable cron when running in production mode
+  # Disabled in development and test so scheduled jobs don't run automatically
+  config.good_job.enable_cron = Rails.env.production?
+  config.good_job.cron = scheduled_jobs || {}
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,3 +1,11 @@
+class SupportUserConstraint
+  def self.matches?(request)
+    service = HostingEnvironment.current_service(request)
+    current_user = DfESignInUser.load_from_session(request.session, service:)&.user
+    current_user&.support_user?
+  end
+end
+
 Rails.application.routes.draw do
   scope via: :all do
     get "/404", to: "errors#not_found", as: :not_found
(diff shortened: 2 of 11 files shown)


config/routes.rb
```diff
@@ -30,4 +38,8 @@
   draw :claims
 
   get :healthcheck, controller: :heartbeat
+
+  # GoodJob admin interface – only accessible to support users
```
elceebee, line 42: Very cool

JamieCleare2525 (approved):
LGTM

config/initializers/good_job.rb
```diff
@@ -0,0 +1,21 @@
+# Configure GoodJob, an Active Job adapter backed by the Postgres database
+# More details at: https://github.com/bensheldon/good_job
+
+Rails.application.configure do
+  # In development, execute jobs within the main `rails server` process
+  config.good_job.execution_mode = :async if Rails.env.development?
```
Nitemaeric, line 6: This looks like something we could put in `config/environments/development.rb`.

config/initializers/good_job.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  # In production, execute jobs in a separate worker process (i.e. `good_job start`)
+  config.good_job.execution_mode = :external if Rails.env.production?
```
Nitemaeric, line 9: This looks like something we could put in `config/environments/production.rb`.

spec/support/dfe_sign_in_user_helper.rb
```diff
@@ -1,4 +1,10 @@
 module DfESignInUserHelper
+  def sign_in_as(user)
```
Nitemaeric, line 2: 👍

PR #301, opened 2024-02-21
Remove School Users

Nitemaeric:
## Context

We need to be able to remove users from schools.

- Add remove flows for both Users and Support Users.
- Updated `ApplicationPolicy` to create sensible defaults.
  - The `update?` method is not used within the `UsersController` scope and `undercover` is blocking changes, therefore, we will keep the default `update?` as `false` for now.

## Guidance to review

- Try adding a User and removing them.
- You shouldn't be able to remove yourself from a school when signed in as a non-support User.

## Screenshots

| Change | Screenshot |
| ------ | ---------- |
| Added "Remove user" link | CleanShot 2024-02-22 at 10 55 54 |
| Added `/schools/:school_id/users/:user_id/remove` path | CleanShot 2024-02-22 at 10 56 16 |
| Removal alert | CleanShot 2024-02-22 at 10 56 59 |

### DIFF
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -7,19 +7,23 @@ class ApplicationPolicy
   end
 
   def index?
-    false
+    true
   end
 
   def show?
-    false
+    true
   end
 
   def create?
-    false
+    true
   end
 
   def new?
     create?
+  end
+
+  def check?
+    new?
   end
 
   def update?
@@ -32,6 +36,10 @@ class ApplicationPolicy
 
   def destroy?
     false
+  end
+
+  def remove?
+    destroy?
   end
 
   class Scope
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -6,42 +6,6 @@ class ApplicationPolicy
     @record = record
   end
 
-  def index?
-    true
-  end
-
-  def show?
-    true
-  end
-
-  def create?
-    true
-  end
-
-  def new?
-    create?
-  end
-
-  def check?
-    new?
-  end
-
-  def update?
-    false
-  end
-
-  def edit?
-    update?
-  end
-
-  def destroy?
-    false
-  end
-
-  def remove?
-    destroy?
-  end
-
   class Scope
     def initialize(user, scope)
       @user = user
--- a/app/views/claims/schools/users/remove.html.erb
+++ b/app/views/claims/schools/users/remove.html.erb
@@ -16,7 +16,7 @@
       <%= govuk_button_to t(".remove_user"), claims_school_user_path(@school, @user), warning: true, method: :delete %>
 
       <p class="govuk-body">
-        <%= govuk_link_to(t(".cancel"), claims_school_user_path(@school, @user)) %>
+        <%= govuk_link_to(t(".cancel"), claims_school_user_path(@school, @user), no_visited_state: true) %>
       </p>
     </div>
   </div>
--- a/app/views/claims/schools/users/remove.html.erb
+++ b/app/views/claims/schools/users/remove.html.erb
@@ -1,0 +1,23 @@
+<% content_for(:page_title) { sanitize(t(".page_title", user_name: @user.full_name)) } %>
+<% render "claims/support/primary_navigation", current: :organisations %>
+
+<%= content_for(:before_content) do %>
+  <%= govuk_back_link(href: claims_school_user_path(@school, @user)) %>
+<% end %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-l"><%= t(".caption", user_name: @user.full_name) %></span>
+      <h2 class="govuk-heading-l"><%= t(".heading") %></h2>
+
+      <%= govuk_warning_text(text: t(".warning", school_name: @school.name)) %>
+
+      <%= govuk_button_to t(".remove_user"), claims_school_user_path(@school, @user), warning: true, method: :delete %>
+
+      <p class="govuk-body">
+        <%= govuk_link_to(t(".cancel"), claims_school_user_path(@school, @user)) %>
+      </p>
+    </div>
+  </div>
+</div>
--- a/app/views/claims/support/schools/users/remove.html.erb
+++ b/app/views/claims/support/schools/users/remove.html.erb
@@ -16,7 +16,7 @@
       <%= govuk_button_to t(".remove_user"), claims_support_school_user_path(@school, @user), warning: true, method: :delete %>
 
       <p class="govuk-body">
-        <%= govuk_link_to(t(".cancel"), claims_support_school_user_path(@school, @user)) %>
+        <%= govuk_link_to(t(".cancel"), claims_support_school_user_path(@school, @user), no_visited_state: true) %>
       </p>
     </div>
   </div>
--- a/app/views/claims/support/schools/users/remove.html.erb
+++ b/app/views/claims/support/schools/users/remove.html.erb
@@ -1,0 +1,23 @@
+<% content_for(:page_title) { sanitize(t(".page_title", user_name: @user.full_name, school_name: @school.name)) } %>
+<% render "claims/support/primary_navigation", current: :organisations %>
+
+<%= content_for(:before_content) do %>
+  <%= govuk_back_link(href: claims_support_school_user_path(@school, @user)) %>
+<% end %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-l"><%= t(".caption", user_name: @user.full_name, school_name: @school.name) %></span>
+      <h2 class="govuk-heading-l"><%= t(".heading") %></h2>
+
+      <%= govuk_warning_text(text: t(".warning", school_name: @school.name)) %>
+
+      <%= govuk_button_to t(".remove_user"), claims_support_school_user_path(@school, @user), warning: true, method: :delete %>
+
+      <p class="govuk-body">
+        <%= govuk_link_to(t(".cancel"), claims_support_school_user_path(@school, @user)) %>
+      </p>
+    </div>
+  </div>
+</div>
--- a/app/components/primary_navigation_component.html.erb
+++ b/app/components/primary_navigation_component.html.erb
@@ -1,13 +1,11 @@
-<div class="app-primary-navigation">
-  <div class="app-primary-navigation govuk-!-display-none-print">
-    <div class="govuk-width-container app-primary-navigation--justify-between">
-      <nav class="app-primary-navigation__nav" aria-label="main menu">
-        <ul class="app-primary-navigation__list">
-          <% navigation_items.each do |navigation_item| %>
-            <%= navigation_item %>
-          <% end %>
-        </ul>
-      </nav>
-    </div>
+<div class="app-primary-navigation govuk-!-display-none-print">
+  <div class="govuk-width-container app-primary-navigation--justify-between">
+    <nav class="app-primary-navigation__nav" aria-label="main menu">
+      <ul class="app-primary-navigation__list">
+        <% navigation_items.each do |navigation_item| %>
+          <%= navigation_item %>
+        <% end %>
+      </ul>
+    </nav>
   </div>
 </div>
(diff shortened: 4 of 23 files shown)


app/policies/application_policy.rb
```diff
@@ -7,21 +7,25 @@ def initialize(user, record)
...
   def create?
-    false
+    true
   end
 
   def new?
     create?
   end
 
+  def check?
```
JamieCleare2525, line 25: If this is the same as `new?` you should use `alias_method`.
https://github.com/varvet/pundit?tab=readme-ov-file#plain-old-ruby

app/policies/application_policy.rb
```diff
@@ -34,6 +38,10 @@ def destroy?
     false
   end
 
+  def remove?
```
JamieCleare2525, line 41: Similar to above, use `alias_method` if this is the same as `:destroy`

Nitemaeric: Okay, this was an interesting find.

So, `alias_method` works as a direct alias to the _existing_ definition. This means, it references its own definition to the definition _at the time_. It does not delegate the call like `def remove?; destroy?; end` does.

This means, when we override the initial method in a subclass, it does not reference the _new_ definition.

```ruby
class ApplicationPolicy
  def destroy?
    puts "original destroy?"
    false
  end

  alias_method :remove?, :destroy?
  # This is _kind_ of like doing
  # define_method(:remove?) do
  #   false
  # end
end

class Claims::UserPolicy < ApplicationPolicy
  def destroy?
    puts "new destroy?"
    user != record
  end

  # remove? is still referencing the original definition
end
```

When calling `claims/user_policy#remove?` it will print out "original destroy?" and `claims/user_policy#destroy?` will print out "new destroy?".

-----

As we want to keep `remove?` coupled to `destroy?` by default, we should chain calls instead of aliasing them.

JamieCleare2525: To be honest, I wouldn't expect any inherited, aliased or chained methods within the ApplicationPolicy. As really the default behaviour should all be specifically defined per policy method.
```
def destroy?
  false
end

def remove?
  false
end
```
I would also expect all default behaviours to be `false`. That way users are prevented from accessing pages, unless their is a specific policy for that controller method. Having default `true` show the irrelevance and unnecessary addition of these methods.
Furthermore, I don't think that any policy methods belong in the `ApplicationPolicy `. Policy methods should be defined purely in the policies related to their specific controllers, as not all controllers will require a methods such as `remove` or `check` policy, and adding default policies per new controller method isn't scalable.

CatalinVoineag (approved):

app/policies/application_policy.rb
ollietreend: @Nitemaeric Sorry to chime in on this, but I see the [issues you mentioned](https://github.com/DFE-Digital/itt-mentor-services/actions/runs/8003289848/job/21858315579#step:8:14) with getting undercover to accept uncovered changes to this file. I think this could be the code & test suite telling us that these methods shouldn't be here.

Maybe we should be thinking of our ApplicationPolicy similarly to ApplicationController. Rather than defining a bunch of methods here that we think might be needed by children, maybe we should leave that up to the child classes themselves. For example, you wouldn't declare a `show`, `check` or `destroy` action in the ApplicationController – even though you're confident your individual models will probably all need them.

I think it's also telling that the ApplicationPolicy methods aren't covered by even our system tests. We aren't actually using any of these untested methods within the real-world implementation – because most of our auth needs are covered by "[are you logged in?](https://github.com/DFE-Digital/itt-mentor-services/blob/c815941b5245ee49be1ec0d96ad3726fd59ed87a/app/controllers/application_controller.rb#L39-L45)" and "[are you a support user?](https://github.com/DFE-Digital/itt-mentor-services/blob/bf8bde00122b6b4d52552d97f8e71423e667e51b/app/controllers/placements/support/application_controller.rb#L6-L10)" which are already established and inherited behaviours of our controllers.

**TL;DR:** I think our codebase & test coverage is telling us that these methods defined in ApplicationPolicy are unused and not valuable. Their existence could even be misleading to any other dev who picks up the codebase – seeing the existence of them could suggest we use policies for more than we do. Right now our only use of policies is [this policy scope](https://github.com/DFE-Digital/itt-mentor-services/blob/795c63f86763a4875cac9f5d50e1048a5817d327/app/controllers/claims/schools_controller.rb#L5) and [this `destroy?` policy](https://github.com/DFE-Digital/itt-mentor-services/blob/d99a09868755bc85bf570d5ee7707ac20d409ad3/app/controllers/placements/organisations/users_controller.rb#L45-L50) – we do not use the boilerplate `show?`, `update?`, etc. methods defined in this ApplicationPolicy class.

Nitemaeric: I believe I see your points, but I'm still inclined to move forward with the inheritance, default permissions, and `remove?` and `check?` permissions.

> To be honest, I wouldn't expect any inherited, aliased or chained methods within the ApplicationPolicy. As really the default behaviour should all be specifically defined per policy method.
>
> ```ruby
> def destroy?
>   false
> end
>
> def remove?
>   false
> end
> ```

The chained methods are by design, a pattern that pundit [encourages](https://github.com/varvet/pundit/blob/main/lib/generators/pundit/install/templates/application_policy.rb) with its generated `ApplicationPolicy`. See `new?` and `edit?`.

We would ~never~ _likely never_ want `remove?` and `destroy?` to go out of sync. Therefore the `remove?` permission is equal to the `destroy?` permission by design. This architecture is open enough to allow de-coupling the permissions though.

> I would also expect all default behaviours to be false. That way users are prevented from accessing pages, unless their is a specific policy for that controller method. Having default true shows the irrelevance and unnecessary addition of these methods.

The default CRUD permissions are defined at `ApplicationPolicy` as a way to configure global _base_ permissions. For example, if we were to require that _only_ Support Users can "create" records, we can do that here with minimal effort.

As for defaulting to `true`, by our application design, we allow read permissions to all records a user can _reach_ by default. We currently restrict access to unowned records by scoping the top-level query, e.g. `@user.schools.find(params.require(:id))`. This design hides unowned records from malicious actors. An alternative would be to open up the records and define something like:

```ruby
def show?
  record.respond_to?(:user) ? record.user == user : false # or true, depending on _unowned_ record defaults.
end
```

Defaulting to `false` and redefining every policy's permission to `true` would also defeat the purpose of having the base permissions, which are open to extension.

I'm typically one to vouch for implementing allow-lists rather than deny-lists, but the pattern of base `true` or `false` permissions with inherited policies defining explicit permissions, makes more sense in my experience.

We would never need to re-define a permission with `false`:

```ruby
class SomePolicy < ApplicationPolicy
  def action?
    false
  end
end
```

as it makes more sense to not provide the action (and route) at all instead. Which situation would merit us to introduce some code that no users have access to?

> Furthermore, I don't think that any policy methods belong in the ApplicationPolicy . Policy methods should be defined purely in the policies related to their specific controllers, as not all controllers will require a methods such as remove or check policy, and adding default policies per new controller method isn't scalable. For example, if we were to add a publish method for the PlacementsController, if a policy is required it should only be necessary to add it in a PlacementsPolicy, and not required to be added in the ApplicationPolicy

Not all controllers implement all CRUD actions, but the base permission methods added to `ApplicationPolicy` provide a common ancestry _if_ required. The `remove` and `check` actions are common enough (7 and 11 controllers respectively. We have 21 controllers in total, for reference of pattern frequency) that we can benefit from having them in our **base** policy class.

For an uncommon action or off-pattern action, we should definitely define them within their own scopes.

Nitemaeric: @ollietreend I've just seen your message after refreshing. Here we go... 😆 

> Sorry to chime in on this

Please do! I'm happy to hear everyone's thoughts!

> but I see the [issues you mentioned](https://github.com/DFE-Digital/itt-mentor-services/actions/runs/8003289848/job/21858315579#step:8:14) with getting undercover to accept uncovered changes to this file. I think this could be the code & test suite telling us that these methods shouldn't be here.

This is more telling that this particular controller does not implement the `update` action. Given a controller with the `update` action, undercover would be happy with the change.

This did trigger a conscious thought about what permissions should be allowed by default though. Should a user _always_ be allowed to update _any_ data? Probably not. The same goes for deleting data. As for reading and creating data, I don't think we want to restrict these actions for our users. A user should always be able to create data, given a flow is available, and read data they have access to, i.e. we allow them to find.

> For example, you wouldn't declare a `show`, `check` or `destroy` action in the ApplicationController – even though you're confident your individual controllers will probably all need them.

Rails _already_ does this in `ActionController::Base`, which `ApplicationController` inherits from. It's default behaviour is to render the relevant template. i.e. `show` will render `show.html.erb`.

> I think it's also telling that the ApplicationPolicy methods aren't covered by even our system tests. We aren't actually using any of these untested methods within the real-world implementation – because most of our auth needs are covered by "[are you logged in?](https://github.com/DFE-Digital/itt-mentor-services/blob/c815941b5245ee49be1ec0d96ad3726fd59ed87a/app/controllers/application_controller.rb#L39-L45)" and "[are you a support user?](https://github.com/DFE-Digital/itt-mentor-services/blob/bf8bde00122b6b4d52552d97f8e71423e667e51b/app/controllers/placements/support/application_controller.rb#L6-L10)" which are already established and inherited behaviours of our controllers.

Yes, the majority of our controllers have not adopted the usage of policies. This is something we should remedy, not pick and choose when to use a policy. This PR introduces those policy checks for the effected controller here, `Claims::Schools::UsersController`.

app/views/claims/schools/users/remove.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-l"><%= t(".caption", user_name: @user.full_name) %></span>
+      <h2 class="govuk-heading-l"><%= t(".heading") %></h2>
+
+      <%= govuk_warning_text(text: t(".warning", school_name: @school.name)) %>
+
+      <%= govuk_button_to t(".remove_user"), claims_school_user_path(@school, @user), warning: true, method: :delete %>
+
+      <p class="govuk-body">
+        <%= govuk_link_to(t(".cancel"), claims_school_user_path(@school, @user)) %>
```
elceebee, line 19: Just a small thing, but I just fixed all these yesterday. This link should have the `no_visited_state: true` attribute

app/views/claims/support/schools/users/remove.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-l"><%= t(".caption", user_name: @user.full_name, school_name: @school.name) %></span>
+      <h2 class="govuk-heading-l"><%= t(".heading") %></h2>
+
+      <%= govuk_warning_text(text: t(".warning", school_name: @school.name)) %>
+
+      <%= govuk_button_to t(".remove_user"), claims_support_school_user_path(@school, @user), warning: true, method: :delete %>
+
+      <p class="govuk-body">
+        <%= govuk_link_to(t(".cancel"), claims_support_school_user_path(@school, @user)) %>
```
elceebee, line 19: As above `no_visited_state: true`

app/policies/application_policy.rb
ollietreend: @Nitemaeric, @JamieCleare2525 and I have just talked this through and have agreed on the following steps forward:

- **Remove all methods from ApplicationPolicy** – to avoid making assumptions about which policies individual controllers will need, and to remove a potentially misleading/confusing piece of code that could make it look like granular policies are used across the board.
- **Explicitly implement policies when they're needed** – we'll take a 'permissive' model to reflect the fact that most actions are allowed most of the time. And in the specific cases where that's not true, we'll write a policy for it and use it in the `before_action` for those particular actions.

This is a decision that we may choose to revisit down the line. We may even find it necessary to handle policies differently between the two different services – for example, if they generally tend to have different permission models. But for now this is the approach we'll take.

app/views/claims/schools/users/remove.html.erb
Nitemaeric: Resolved here: https://github.com/DFE-Digital/itt-mentor-services/pull/301/commits/81bca1e5ea66ffc12ccd1bd0fc39956344bc93bc

app/views/claims/support/schools/users/remove.html.erb
Nitemaeric: Resolved here: https://github.com/DFE-Digital/itt-mentor-services/pull/301/commits/81bca1e5ea66ffc12ccd1bd0fc39956344bc93bc

JamieCleare2525 (approved):

olioldfield:
Looks good to me

PR #305, opened 2024-02-22
As a User, I want to remove a Mentor from a School

gms-gs:
## Context

To allow a user to remove a mentor

To add the remove mentor flow for a user

## Guidance to review

- Sign in as a user
- navigate to a schools mentor page
- Select a mentor
- Follow the remove mentor flow

## Link to Trello card

https://trello.com/c/ZDSoLcZv/164-as-a-user-i-want-to-remove-a-mentor-from-a-school

## Things to check

### DIFF
--- a/app/controllers/claims/schools/mentors_controller.rb
+++ b/app/controllers/claims/schools/mentors_controller.rb
@@ -1,7 +1,25 @@
 class Claims::Schools::MentorsController < ApplicationController
   include Claims::BelongsToSchool
+  before_action :set_mentor, only: %i[show remove destroy]
 
   def index
     @pagy, @mentors = pagy(@school.mentors.order(:first_name, :last_name))
+  end
+
+  def show; end
+
+  def remove; end
+
+  def destroy
+    mentor_membership = @mentor.mentor_memberships.find_by!(school: @school)
+    mentor_membership.destroy!
+
+    redirect_to claims_school_mentors_path(@school), flash: { success: t(".success") }
+  end
+
+  private
+
+  def set_mentor
+    @mentor = @school.mentors.find(params.require(:id))
   end
 end
--- a/app/views/claims/schools/mentors/index.html.erb
+++ b/app/views/claims/schools/mentors/index.html.erb
@@ -18,7 +18,7 @@
           <% table.with_body do |body| %>
             <% @mentors.each do |mentor| %>
               <% body.with_row do |row| %>
-                <% row.with_cell(text: govuk_link_to(mentor.full_name, "#")) %>
+                <% row.with_cell(text: govuk_link_to(mentor.full_name, claims_school_mentor_path(@school, mentor))) %>
                 <% row.with_cell(text: mentor.trn) %>
               <% end %>
             <% end %>
--- a/app/views/claims/schools/mentors/show.html.erb
+++ b/app/views/claims/schools/mentors/show.html.erb
@@ -1,8 +1,8 @@
-<% content_for :page_title, "#{@mentor.full_name} - #{t(".heading")} - #{@school.name}" %>
+<% content_for :page_title, "#{@mentor.full_name} - #{t(".heading")}" %>
 
 <% render "claims/support/primary_navigation", current: :organisations %>
 <% content_for(:before_content) do %>
-  <%= govuk_back_link(href: claims_support_school_mentors_path(@school)) %>
+  <%= govuk_back_link(href: claims_school_mentors_path(@school)) %>
 <% end %>
 
 <div class="govuk-width-container">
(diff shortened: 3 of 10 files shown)


app/controllers/claims/schools/mentors_controller.rb
```diff
@@ -1,7 +1,26 @@
...
+  def show; end
+
+  def remove; end
+
+  def destroy
+    mentor_membership = @mentor.mentor_memberships.find_by(school: @school)
+    mentor_membership.destroy!
+
+    redirect_to claims_school_mentors_path(@school)
+    flash[:success] = t(".mentor_removed")
```
Nitemaeric, line 18: I've been doing things _slightly_ different here. I wonder what your thoughts are?

```suggestion
    redirect_to claims_school_mentors_path(@school), flash: { success: t(".success") }
```

1. Inline the flash as it acts like a modifier to the redirection, not logically a step to perform _after_ redirection.
2. Use the `success` key as a behavioural identifier, rather than direct translation. The expanded form is `claims.schools.mentors.destroy.success` which provides the context required for the `.success` key.

app/controllers/claims/schools/mentors_controller.rb
```diff
@@ -1,7 +1,26 @@
...
     @pagy, @mentors = pagy(@school.mentors.order(:first_name, :last_name))
   end
+
+  def show; end
+
+  def remove; end
+
+  def destroy
+    mentor_membership = @mentor.mentor_memberships.find_by(school: @school)
+    mentor_membership.destroy!
```
Nitemaeric, line 15: If a membership isn't found, the following line will fail with a "`nil` does not respond to `destroy!`".

Let's escape that early.

```suggestion
    mentor_membership = @mentor.mentor_memberships.find_by!(school: @school)
    mentor_membership.destroy!
```

We'll now render a 404 page if the membership doesn't exist.

app/views/claims/schools/mentors/show.html.erb
```diff
@@ -0,0 +1,32 @@
...
+          <% end %>
+          <% summary_list.with_row do |row| %>
+            <% row.with_key(text: t(".attributes.mentors.trn")) %>
+            <% row.with_value(text: @mentor.trn) %>
+          <% end %>
+        <% end %>
+    </div>
+  </div>
+
+  <%= govuk_link_to t(".remove_mentor"), remove_claims_school_mentor_path, class: "app-link app-link--destructive" %>
```
Nitemaeric, line 31: I think this should go into the 2/3 column above.

I think you're also missing the arguments, right?

```suggestion
  <%= govuk_link_to t(".remove_mentor"), remove_claims_school_mentor_path(@school, @mentor), class: "app-link app-link--destructive" %>
```

CatalinVoineag (approved):

app/views/claims/schools/mentors/show.html.erb
Nitemaeric: This seems to implicitly work, but I'm not sure _why_. Can you explain why we don't need to pass through the IDs?

app/views/claims/schools/mentors/index.html.erb
```diff
@@ -18,7 +18,7 @@
           <% table.with_body do |body| %>
             <% @mentors.each do |mentor| %>
               <% body.with_row do |row| %>
-                <% row.with_cell(text: govuk_link_to(mentor.full_name, "#")) %>
+                <% row.with_cell(text: govuk_link_to(mentor.full_name, claims_school_mentor_path(id: mentor.id))) %>
```
Nitemaeric, line 21: We don't need to explicitly call `id` here. Rails handles Active Record instances into params for us.

```suggestion
                <% row.with_cell(text: govuk_link_to(mentor.full_name, claims_school_mentor_path(mentor))) %>
```

Side note, Rails calls `#to_param` behind the scenes. Therefore, if you were to add slugged urls, passing `_path(record)` would require no changes, whereas `_path(record.id)` would always pass through the `id` to the url builder and not the custom _param_ value.

app/views/claims/schools/mentors/show.html.erb
```diff
@@ -0,0 +1,31 @@
...
+          <% end %>
+          <% summary_list.with_row do |row| %>
+            <% row.with_key(text: Mentor.human_attribute_name(:last_name)) %>
+            <% row.with_value(text: @mentor.last_name) %>
+          <% end %>
+          <% summary_list.with_row do |row| %>
+            <% row.with_key(text: t(".attributes.mentors.trn")) %>
+            <% row.with_value(text: @mentor.trn) %>
+          <% end %>
+        <% end %>
```
Nitemaeric, line 27: This is indented 1 level too deep.

app/views/claims/support/schools/mentors/index.html.erb
```diff
@@ -23,7 +23,7 @@
           <% table.with_body do |body| %>
             <% @mentors.each do |mentor| %>
               <% body.with_row do |row| %>
-                <% row.with_cell(text: govuk_link_to(mentor.full_name, claims_support_school_mentor_path(id: mentor.id))) %>
+                <% row.with_cell(text: govuk_link_to(mentor.full_name, claims_school_mentor_path(@school, mentor))) %>
```
Nitemaeric, line 26:
```suggestion
                <% row.with_cell(text: govuk_link_to(mentor.full_name, claims_support_school_mentor_path(@school, mentor))) %>
```

Nitemaeric (approved):

PR #315, opened 2024-02-26
Add Page Footer Content

Nitemaeric:
## Context

We want to show "help" information and useful links.

- Add a custom `FooterComponent` that extends the `GovukComponent::FooterComponent` to add a `meta_pre_content` slot.
  - The component html template is copied over directly from the component library. This means we will not receive automatic updates to this component's layout moving forward.
  - I will reach out to Peter Yates to query whether we could add this additional slot in officially or not.
- Update the Footer content.

## Guidance to review

- View the review app.
- See the new footer content.
- The customised Footer component is copied from the `GovukComponent::FooterComponent`.
  - [build_meta_links](https://github.com/x-govuk/govuk-components/blob/main/app/components/govuk_component/footer_component.rb#L65) is overridden to avoid calling view helpers within the `#initialize` method.
  - The `footer_component.html.erb` is taken from https://github.com/x-govuk/govuk-components/blob/main/app/components/govuk_component/footer_component.html.erb with `meta_pre_content` added and `@meta_links` calling `govuk_footer_link_to` instead of rendering themselves as url helper components.

## Screenshots

CleanShot 2024-02-26 at 13 48 45

### DIFF
--- a/app/components/footer_component.html.erb
+++ b/app/components/footer_component.html.erb
@@ -1,0 +1,56 @@
+<%= tag.footer(role: "contentinfo", **html_attributes) do %>
+  <%= tag.div(**container_html_attributes) do %>
+    <% if navigation.present? %>
+      <div class="<%= brand %>-footer__navigation">
+        <%= navigation %>
+      </div>
+
+      <%= tag.hr(class: "#{brand}-footer__section-break") %>
+    <% end %>
+
+    <%= tag.div(class: meta_classes, **meta_html_attributes) do %>
+      <% if meta.present? %>
+        <%= meta %>
+      <% else %>
+        <div class="<%= brand %>-footer__meta-item <%= brand %>-footer__meta-item--grow">
+          <% if meta_pre_content.present? %>
+            <div class="<%= brand %>-footer__meta-custom">
+              <%= meta_pre_content %>
+            </div>
+          <% end %>
+
+          <% if meta_items.any? %>
+            <h2 class="<%= brand %>-visually-hidden"><%= meta_items_title %></h2>
+
+            <ul class="<%= brand %>-footer__inline-list">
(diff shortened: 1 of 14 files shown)


app/components/footer_component.html.erb
```diff
@@ -0,0 +1,55 @@
...
+
+      <%= tag.hr(class: "#{brand}-footer__section-break") %>
+    <% end %>
+
+    <%= tag.div(class: meta_classes, **meta_html_attributes) do %>
+      <% if meta.present? %>
+        <%= meta %>
+      <% else %>
+        <div class="<%= brand %>-footer__meta-item <%= brand %>-footer__meta-item--grow">
+          <% if meta_pre_content.present? %>
```
CatalinVoineag, line 16: What meta data are in these things? `meta_pre_content` and `meta`

app/components/footer_component.rb
```diff
@@ -0,0 +1,3 @@
+class FooterComponent < GovukComponent::FooterComponent
```
CatalinVoineag, line 1: Should we create a spec for this?

app/components/footer_component.html.erb
Nitemaeric: This is mostly a hard copy from the https://github.com/x-govuk/govuk-components/blob/main/app/components/govuk_component/footer_component.html.erb.

I've only just slotted in that additional `meta_pre_content` section.

CatalinVoineag: Ah OK, thanks 👍🏻

JamieCleare2525 (approved):
Approved, given @CatalinVoineag has requested spec for the component.

CatalinVoineag (approved):
Nice

PR #316, opened 2024-02-26
As a Support User, I want to add a Mentor to a School

gms-gs:
## Context

To allow support users to add a mentor to an organisation

Add the add mentor flow for a support user

## Guidance to review

- Sign in as support user
- Go to mentors
- Follow the flow to add a user to a school
- Sign in as a user to that school and check the mentor is there

## Link to Trello card

https://trello.com/c/1kWR8Luj/189-claims-as-a-support-user-i-want-to-add-a-mentor-to-a-school

## Things to check

### DIFF
--- a/app/controllers/claims/support/schools/mentors_controller.rb
+++ b/app/controllers/claims/support/schools/mentors_controller.rb
@@ -4,8 +4,24 @@ class Claims::Support::Schools::MentorsController < Claims::Support::Application
   before_action :set_mentor, only: %i[show remove destroy]
   before_action :authorize_mentor
 
+  helper_method :mentor_form
+
   def index
     @pagy, @mentors = pagy(@school.mentors.order(:first_name, :last_name))
+  end
+
+  def new; end
+
+  def create
+    mentor_form.save!
+
+    redirect_to claims_support_school_mentors_path(@school), flash: { success: t(".success") }
+  end
+
+  def check
+    render :new unless mentor_form.valid?
+  rescue TeachingRecord::RestClient::TeacherNotFoundError
+    render :no_results
   end
 
   def show; end
@@ -27,5 +43,24 @@ class Claims::Support::Schools::MentorsController < Claims::Support::Application
 
   def authorize_mentor
     authorize @mentor || Claims::Mentor
+  end
+
+  def default_params
+    { school: @school }
+  end
+
+  def mentor_params
+    params.require(:claims_mentor_form)
+          .permit(:first_name, :last_name, :trn)
+          .merge(default_params)
+  end
+
+  def mentor_form
+    @mentor_form ||=
+      if params[:claims_mentor_form].present?
+        Claims::MentorForm.new(mentor_params)
+      else
+        Claims::MentorForm.new(default_params)
+      end
   end
 end
--- a/app/forms/claims/mentor_form.rb
+++ b/app/forms/claims/mentor_form.rb
@@ -1,0 +1,42 @@
+class Claims::MentorForm < ApplicationForm
+  FORM_PARAMS = %i[trn].freeze
+
+  attr_accessor :school, :trn, :first_name, :last_name
+
+  validate :validate_mentor
+  validate :validate_membership
+
+  def persist
+    ActiveRecord::Base.transaction do
+      mentor.save! unless mentor.persisted?
+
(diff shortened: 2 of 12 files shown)


app/controllers/claims/support/schools/mentors_controller.rb
```diff
@@ -8,6 +8,22 @@ def index
     @pagy, @mentors = pagy(@school.mentors.order(:first_name, :last_name))
   end
 
+  def new
+    mentor_form
```
CatalinVoineag, line 12: This is a personal opinion, not really a blocker

Can we declare a ` helper_method :mentor_form` at the top of this controller and have the new method empty? `def new; end` It will be consistent with the claim controller

app/forms/claims/mentor_form.rb
```diff
@@ -0,0 +1,42 @@
+class Claims::MentorForm < ApplicationForm
+  FORM_PARAMS = %i[trn].freeze
+
+  attr_accessor :school, :trn, :first_name, :last_name
+
+  validate :validate_mentor
```
CatalinVoineag, line 6: I think the error message is a bit wrong, the message on the prototype appears to be `The mentor has already been added`

Screenshot from 2024-02-27 09-37-05

CatalinVoineag (commented):
Just a few things, almost ready to go 👍🏻

app/controllers/claims/support/schools/mentors_controller.rb
```diff
@@ -3,11 +3,26 @@ class Claims::Support::Schools::MentorsController < Claims::Support::Application
 
   before_action :set_mentor, only: %i[show remove destroy]
   before_action :authorize_mentor
+  before_action :mentor_form, only: %i[new]
```
Nitemaeric, line 6: Can we use the `helper_method` pattern here instead? https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/controllers/claims/schools/claims_controller.rb

`mentor_form` acts more like a memoized instance variable than code that performs an action.

Calling it here is just to make the `@mentor_form` variable available in the view.

Instead, we can make `mentor_form` a helper method and then call that helper method directly in the views. No need to pre-set it here. If we did want to predefine it in the controller, it would make more set to name it `set_mentor_form`.

app/controllers/claims/support/schools/mentors_controller.rb
```diff
@@ -3,11 +3,26 @@ class Claims::Support::Schools::MentorsController < Claims::Support::Application
...
   def index
     @pagy, @mentors = pagy(@school.mentors.order(:first_name, :last_name))
   end
 
+  def new; end
+
+  def create
+    mentor_form.save!
+
+    redirect_to claims_support_school_mentors_path(@school), flash: { success: t(".mentor_added") }
```
Nitemaeric, line 17: Previously, we've started using `.success`.

```suggestion
    redirect_to claims_support_school_mentors_path(@school), flash: { success: t(".success") }
```

app/views/claims/support/schools/mentors/new.html.erb
```diff
@@ -0,0 +1,41 @@
...
+          <p>
+            <%= t(".if_you_dont_have_a_trn") %>
+            <%= govuk_link_to(
+                  t(".trn_guidance"),
+                  "https://www.gov.uk/guidance/teacher-reference-number-trn",
+                  new_tab: true,
+                  no_visited_state: true,
+                ) %>
+            <%= t(".to_find_a_list_trn") %>
+          </p>
```
Nitemaeric, line 31: Instead of splitting this into 3 distinct pieces of text, you can use a single key and pass through the `a` tag.

Although not immediately applicable, this is a better general practice, as when working with other languages, they may not order words the same way as English.

```suggestion
          <p>
            <%= t(".guidance", guidance_link: govuk_link_to(t(".guidance_link_text", "https://www.gov.uk/guidance/teacher-reference-number-trn", new_tab: true, no_visited_state: true)) %>
          </p>
```

```yaml
en:
  ...:
    guidance: If you don’t have a TRN, read the %{guidance_link} to find a lost TRN, or apply for one.
    guidance_link_text: Teacher reference number (TRN) guidance
```

Nitemaeric (approved):

PR #317, opened 2024-02-26
Update CODEOWNERS for Placement-related models

ollietreend:
I've updated CODEOWNERS to consider Placement-related models as being owned by the School Placements team. This covers the models:

- app/models/placement.rb
- app/models/placement_mentor_join.rb
- app/models/placement_subject_join.rb

As well as the associated spec files and factory:

- spec/models/placement_spec.rb
- spec/models/placement_mentor_join_spec.rb
- spec/models/placement_subject_join_spec.rb
- spec/factories/placements.rb

The intention is to reduce noise for the Track & Pay team as we build out Placement related features.


Nitemaeric (approved):
I'm cautious that globbing `placement*.rb` might create false results some time down the line, but not a blocker!

🚀

PR #330, opened 2024-02-27
Refactors the placement route helpers in routes_helper.rb

Kizr:
## Context

Whilst looking through the application logic I noticed that we had a lot of specific route helpers that would ideally be refactored to avoid creating _n_ helpers whenever new routes are required, however to refactor the logic we should first make sure the logic is separated and tested.

Moved the relevant route helpers into their own file and added tests. One invalid route was found whilst adding tests and addressed.

## Guidance to review

The application should continue to work as expected, I identified that one helper `check_placements_organisation_users_path(organisation)` was returning an error upon testing and have updated the logic in what I believe to be the correct way, this should be checked.

## Link to Trello card

N/A I do not have access to Trello yet.

## Things to check

### DIFF
--- a/app/helpers/routes_helper.rb
+++ b/app/helpers/routes_helper.rb
@@ -1,4 +1,6 @@
 module RoutesHelper
+  include Placements::Routes::OrganisationsHelper
+
   def root_path
     {
       claims: claims_root_path,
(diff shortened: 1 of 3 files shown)


app/helpers/routes_helper.rb
```diff
@@ -1,4 +1,6 @@
 module RoutesHelper
+  include Placements::Routes::OrganisationsHelper
```
ollietreend, line 2: My Rails helper & Zeitwerk knowledge is not sharp enough to know if/why this `include` is needed.

Other helper modules in the codebase aren't explicitly included – so I assume Rails autoloads them in. Why is this one different?

spec/helpers/placements/routes/organisations_helper_spec.rb
ollietreend, line 1: Nice! Thanks for adding these tests ⭐

ollietreend (commented):
This is looking good! Just one thing that I want to understand better. Maybe we could talk it through tomorrow in a call.

app/helpers/routes_helper.rb
Kizr: It's nothing fancy, you're right that they're all available in the views by default but we're using them here:

https://github.com/DFE-Digital/itt-mentor-services/blob/03ca825ebbd1425170fa1f78d1edcb764a2d6f19/app/controllers/placements/organisations_controller.rb#L7-L13

In the original implementation we added the include here:

https://github.com/DFE-Digital/itt-mentor-services/blob/03ca825ebbd1425170fa1f78d1edcb764a2d6f19/app/controllers/application_controller.rb#L3

I wanted to keep the disruption to the code as minimal as possible as I prepared to refactor so I decided to add the include inside the existing `RoutesHelper` instead of adding a new one to `ApplicationController`.

ollietreend: Makes sense! Thanks for explaining.

ollietreend (approved):

CatalinVoineag (approved):
👋🏻 Looks good

PR #332, opened 2024-02-28
Add mentor training hours to claim form

CatalinVoineag:
## Context

The mentor training hours form will come after the user selects the mentors for the claim. It will cycle through the mentors who don't have mentor training hours.

A column has been added to the mentor_training table, `custom_hours_completed`. This is for capturing the custom hours the user can input.

After inputting all the hours for the mentors the form will redirect to the check claim page.

The form also handles the paths of the backlinks and the success redirects.

`Claims::Schools::Claims::MentorTrainingsController`
`Claim::MentorTrainingForm`
Changes to `Claim::MentorsForm`
Changes to claims#check page

## Guidance to review

Pull or go on review app
Create a claim with a few mentors
Add mentor hours
Change the mentor hours on check page
Change the mentors on check page
Leave inputs blank to get errors
Play around with the back links
Thank you 🥇 

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/5a02f6ec-ec7a-42e2-bbe6-f03a5f38f417

### DIFF
--- a/app/controllers/claims/schools/claims/mentor_trainings_controller.rb
+++ b/app/controllers/claims/schools/claims/mentor_trainings_controller.rb
@@ -1,0 +1,49 @@
+class Claims::Schools::Claims::MentorTrainingsController < Claims::ApplicationController
+  include Claims::BelongsToSchool
+  before_action :authorize_claim
+  helper_method :mentor_training_form
+
+  def edit; end
+
+  def update
+    if mentor_training_form.save
+      redirect_to mentor_training_form.success_path
+    else
+      render :edit
+    end
+  end
+
+  private
+
+  def mentor_training_form
+    @mentor_training_form ||=
+      if params[:claim_mentor_training_form].present?
+        Claim::MentorTrainingForm.new(mentor_training_params)
+      else
+        Claim::MentorTrainingForm.new(default_params)
+      end
+  end
+
+  def mentor_training_params
+    params.require(:claim_mentor_training_form).permit(
+      :hours_completed,
+      :custom_hours_completed,
+    ).merge(default_params)
+  end
+
+  def default_params
+    { claim:, mentor_training: }
+  end
+
+  def mentor_training
+    @mentor_training ||= @claim.mentor_trainings.find(params.require(:id))
+  end
+
+  def claim
+    @claim ||= @school.claims.find(params.require(:claim_id))
+  end
+
+  def authorize_claim
+    authorize claim
+  end
+end
--- a/app/forms/claim/mentor_training_form.rb
+++ b/app/forms/claim/mentor_training_form.rb
@@ -1,0 +1,72 @@
+class Claim::MentorTrainingForm < ApplicationForm
+  DEFAULT_HOURS_COMPLETED = %w[6 20].freeze
+  attr_accessor :claim, :mentor_training, :hours_completed, :custom_hours_completed
+
+  validates :hours_completed, presence: true, unless: :custom_hours?
+  validates :custom_hours_completed, presence: true, if: :custom_hours?
+
+  validate :validate_mentor_training_hours
+
+  delegate :mentor_full_name, to: :mentor_training
+
+  def persist
+    updated_mentor_training.save!
+  end
+
+  def back_path
+    if mentor_trainings.index(mentor_training).zero?
+      edit_claims_school_claim_mentor_path(claim.school, claim, mentor_training.mentor_id)
+    else
+      previous_mentor_training = mentor_trainings[mentor_trainings.index(mentor_training) - 1]
+      edit_claims_school_claim_mentor_training_path(
+        claim.school,
+        claim,
+        previous_mentor_training,
+        params: {
+          claim_mentor_training_form: { hours_completed: previous_mentor_training.hours_completed },
+        },
+      )
+    end
+  end
+
+  def success_path
+    if claim.mentor_trainings.without_hours.any?
+      edit_claims_school_claim_mentor_training_path(
+        claim.school,
+        claim,
+        claim.mentor_trainings.without_hours.first,
+      )
+    else
+      check_claims_school_claim_path(claim.school, claim)
+    end
+  end
+
+  def custom_hours?
+    @custom_hours ||= hours_completed == "on" || custom_hours_completed?
+  end
+
+  private
+
+  def custom_hours_completed?
+    if hours_completed.present?
+      !DEFAULT_HOURS_COMPLETED.include?(hours_completed)
+    end
+  end
+
+  def mentor_trainings
+    @mentor_trainings ||= claim.mentor_trainings.order_by_mentor_full_name
+  end
+
+  def updated_mentor_training
+    @updated_mentor_training ||= begin
+      mentor_training.hours_completed = custom_hours? ? custom_hours_completed : hours_completed
+      mentor_training
+    end
+  end
+
+  def validate_mentor_training_hours
+    if updated_mentor_training.invalid? && updated_mentor_training.errors[:hours_completed]
+      errors.add(:custom_hours_completed, updated_mentor_training.errors.first.message)
+    end
+  end
+end
--- a/app/forms/claim/mentors_form.rb
+++ b/app/forms/claim/mentors_form.rb
@@ -11,6 +11,18 @@ class Claim::MentorsForm < ApplicationForm
     claim
   end
 
+  def update_success_path
+    if claim.mentor_trainings.without_hours.any?
+      edit_claims_school_claim_mentor_training_path(
+        claim.school,
+        claim,
+        claim.mentor_trainings.without_hours.first,
+      )
+    else
+      check_claims_school_claim_path(claim.school, claim)
+    end
+  end
+
   private
 
   def updated_claim
@@ -22,7 +34,11 @@ class Claim::MentorsForm < ApplicationForm
 
   def mentor_trainings
     @mentor_trainings ||= Array.wrap(mentor_ids).map do |mentor_id|
-      claim.mentor_trainings.build(mentor_id:, provider_id: claim.provider_id)
+      claim.mentor_trainings.find_or_initialize_by(
+        mentor_id:,
+        provider_id: claim.provider_id,
+        claim_id: claim.id,
+      )
     end
   end
 
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -26,7 +26,5 @@ class Claim < ApplicationRecord
   has_many :mentor_trainings
   has_many :mentors, through: :mentor_trainings
 
-  accepts_nested_attributes_for :mentor_trainings
-
   delegate :name, to: :provider, prefix: true, allow_nil: true
 end
--- a/app/models/mentor_training.rb
+++ b/app/models/mentor_training.rb
@@ -28,4 +28,15 @@ class MentorTraining < ApplicationRecord
   belongs_to :claim
   belongs_to :mentor, optional: true
   belongs_to :provider, optional: true
+
+  validates(
+    :hours_completed,
+    numericality: { greater_than: 0, less_than_or_equal_to: 20, only_integer: true },
+    allow_nil: true,
+  )
+
+  scope :without_hours, -> { where(hours_completed: nil).order_by_mentor_full_name }
+  scope :order_by_mentor_full_name, -> { joins(:mentor).merge(Mentor.order_by_full_name) }
+
+  delegate :full_name, to: :mentor, prefix: true, allow_nil: true
 end
--- a/app/views/claims/schools/claims/mentor_trainings/edit.html.erb
+++ b/app/views/claims/schools/claims/mentor_trainings/edit.html.erb
@@ -1,0 +1,61 @@
+<% content_for(
+  :page_title,
+  t(".page_title", mentor: mentor_training_form.mentor_training.mentor_full_name),
+) %>
+
+<% render "claims/schools/primary_navigation", school: @school, current: :claims %>
+
+<%= content_for(:before_content) do %>
+  <%= govuk_back_link(href: mentor_training_form.back_path) %>
+<% end %>
+
+<div class="govuk-width-container">
+  <%= form_with(
+    model: mentor_training_form,
+    url: claims_school_claim_mentor_training_path(
+      @school,
+      mentor_training_form.claim,
+      mentor_training_form.mentor_training,
+    ),
+    method: :patch,
+  ) do |f| %>
+    <%= f.govuk_error_summary %>
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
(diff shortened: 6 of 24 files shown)


app/models/claim.rb
```diff
@@ -23,10 +23,17 @@ class Claim < ApplicationRecord
   belongs_to :school
   belongs_to :provider
 
-  has_many :mentor_trainings
-  has_many :mentors, through: :mentor_trainings
+  has_many(
+    :mentor_trainings,
+    -> { includes(:mentor).order("mentors.first_name", "mentors.last_name") },
```
CatalinVoineag, line 28: The prototype order the mentors by mentor first name and last name and this is also important when iterating over the mentors to add the hours, they should be ordered by something else than id as uuid ordering is not reliable

app/controllers/claims/schools/claims/mentor_trainings_controller.rb
```diff
@@ -0,0 +1,49 @@
+class Claims::Schools::Claims::MentorTrainingsController < ApplicationController
```
JamieCleare2525, line 1:
```suggestion
class Claims::Schools::Claims::MentorTrainingsController < Claims::ApplicationController
```

app/forms/claim/mentor_training_form.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def back_path
+    if claim.mentor_trainings.find_index(mentor_training).zero?
+      edit_claims_school_claim_mentor_path(claim.school, claim, mentor_training.mentor_id)
+    else
+      edit_claims_school_claim_mentor_training_path(
+        claim.school,
+        claim,
+        claim.mentor_trainings[claim.mentor_trainings.index(mentor_training) - 1].id,
```
JamieCleare2525, line 25: Same here. Not a blocker, but I don't think you need `.id`

app/forms/claim/mentor_training_form.rb
```diff
@@ -0,0 +1,81 @@
...
+      )
+    end
+  end
+
+  def success_path
+    if claim.mentor_trainings_without_hours.any?
+      edit_claims_school_claim_mentor_training_path(
+        claim.school,
+        claim,
+        claim.mentor_trainings_without_hours.first.id,
```
JamieCleare2525, line 35: Same here. Not a blocker, but I don't think you need .id

app/forms/claim/mentors_form.rb
```diff
@@ -11,6 +12,18 @@ def to_model
     claim
   end
 
+  def update_success_path
+    if claim.mentor_trainings_without_hours.any?
+      edit_claims_school_claim_mentor_training_path(
+        claim.school,
+        claim,
+        claim.mentor_trainings_without_hours.first.id,
```
JamieCleare2525, line 20: Same here. Not a blocker, but I don't think you need `.id`

app/models/claim.rb
```diff
@@ -23,10 +23,17 @@ class Claim < ApplicationRecord
   belongs_to :school
   belongs_to :provider
 
-  has_many :mentor_trainings
-  has_many :mentors, through: :mentor_trainings
+  has_many(
+    :mentor_trainings,
```
JamieCleare2525, line 27: Does this need to be `dependent: :destroy`?

app/models/claim.rb
```diff
@@ -23,10 +23,17 @@ class Claim < ApplicationRecord
...
+    -> { includes(:mentor).order("mentors.first_name", "mentors.last_name") },
+  )
+  has_many(
```
JamieCleare2525, line 30: does this need to be `dependent: :destroy`?

app/views/claims/schools/claims/mentors/_form.html.erb
```diff
@@ -7,7 +7,7 @@
 
         <%= f.govuk_collection_check_boxes(
           :mentor_ids,
-          school.mentors,
+          school.mentors.order(:first_name, :last_name),
```
JamieCleare2525, line 10: might be worth adding this order as a scope. Something like `order_by_name`

app/forms/claim/mentor_training_form.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def custom_hours_checked?
+    custom_hours? || mentor_training.custom_hours_completed.present?
+  end
+
+  private
+
+  def custom_hours?
+    hours_completed == "on"
```
JamieCleare2525, line 49: It's worth memoizing this method, as it's hit 3 times in one process.

app/models/claim.rb
CatalinVoineag: Yeah, though we probably won't destroy a claim but I'll add it for completion

app/forms/claim/mentor_training_form.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def updated_mentor_training
+    @updated_mentor_training ||= begin
+      mentor_training.hours_completed = custom_hours? ? nil : hours_completed
```
JamieCleare2525, line 54: Rather than doing the `custom_hours?` check twice. You could move this to and `if else`, so you only need to do the check one.

app/models/claim.rb
CatalinVoineag: Actually, I don't think we should add this. If we do, it would be easier to destroy claims in the console. Currently without this you cannot do `Claim.last.destroy`

app/models/claim.rb
JamieCleare2525: I don't think this was done.
But @ollietreend made a point on my PR that we can protect with `restrict_with_exception` instead https://github.com/DFE-Digital/itt-mentor-services/pull/334#discussion_r1507718744

spec/models/mentor_spec.rb
```diff
@@ -43,6 +43,20 @@
...
   end
 
+  context "with scopes" do
+    describe "#order_by_full_name" do
+      it "returns the mentors ordered by full name" do
+        mentor1 = create(:mentor, first_name: "Anne", last_name: "Doe")
+        mentor2 = create(:mentor, first_name: "Anne", last_name: "Smith")
+        mentor3 = create(:mentor, first_name: "John", last_name: "Doe")
+
+        expect(described_class.order_by_full_name).to match_array(
```
JamieCleare2525, line 53: `match_array` doesn't care about order. You need to use `eq`

app/models/claim.rb
CatalinVoineag: Yeah I made the same point here https://github.com/DFE-Digital/itt-mentor-services/pull/332#discussion_r1507754121 I don't think we need `restrict_with_exception` as you cannot delete a claim if it has mentor trainings anyway

CatalinVoineag: You'll get something like this 
Screenshot from 2024-02-29 15-53-31

app/views/claims/schools/claims/mentor_trainings/edit.html.erb
```diff
@@ -0,0 +1,59 @@
...
+
+          <%= f.govuk_radio_button(
+            :hours_completed,
+            "on",
+            checked: mentor_training_form.custom_hours_checked?,
+            label: { text: t(".other_amount") },
+          ) do %>
+            <h5 class="govuk-heading-s govuk-!-margin-bottom-1"><%= t(".number_of_hours") %></h5>
+            <%= f.govuk_number_field(
+              :custom_hours_completed,
```
Nitemaeric, line 46: Why does this need to be `custom_hours_completed`?

app/models/claim.rb
Nitemaeric, line 30: While this is possible, it feels like a code smell to me.

Calling `claim.mentor_trainings.without_hours` seems more intuitive, `[resource].[association].[scope]`. What do you think?

app/forms/claim/mentors_form.rb
```diff
@@ -1,4 +1,5 @@
 class Claim::MentorsForm < ApplicationForm
+  include Rails.application.routes.url_helpers
```
Nitemaeric, line 2: Personally, I would add this to `ApplicationForm`. Components, Views, Controller, etc, all have access to url helpers.

app/controllers/claims/schools/claims/mentor_trainings_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    ).merge(default_params)
+  end
+
+  def default_params
+    { claim:, mentor_training: }
+  end
+
+  def mentor_training
+    @claim.mentor_trainings.find(params.require(:id))
+  end
```
Nitemaeric, line 40: Should this be memoized?

app/models/claim.rb
Nitemaeric: This would re-use the `Mentor` scope.

```suggestion
    -> { includes(:mentor).merge(Mentor.order_by_full_name) },
```

app/views/claims/schools/claims/mentor_trainings/edit.html.erb
CatalinVoineag: I think we need to distinguish between the field where the user inputs his custom hours and the preselected hours Don't think rails will magically know, unless I missed something. This is a dirty demo with all fields named `hours_completed`. The error is not targeting the field corectly and when you come and update the form it will not work as you expected.


https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/63028c54-1026-4e21-b8e2-ccedf2afecfa

Nitemaeric: Simon has given the 👌 for us to go with the singular `hours_completed` attribute. https://ukgovernmentdfe.slack.com/archives/C0657JE64HX/p1709291417171859?thread_ts=1709291213.242919&cid=C0657JE64HX

Nitemaeric: Base data-wise, this distinction isn't preferred, as it adds presentational concerns to the raw data.

If we want to learn more about the user behaviour, we can add in analytics tools.

app/forms/claim/mentor_training_form.rb
```diff
@@ -0,0 +1,90 @@
...
+    if mentor_trainings.find_index(mentor_training).zero?
+      edit_claims_school_claim_mentor_path(claim.school, claim, mentor_training.mentor_id)
+    else
+      edit_claims_school_claim_mentor_training_path(
+        claim.school,
+        claim,
+        mentor_trainings[mentor_trainings.index(mentor_training) - 1],
+      )
+    end
+  end
```
Nitemaeric, line 27: > [Array#find_index](https://docs.ruby-lang.org/en/3.2/Array.html#method-i-find_index) is an alias for [Array#index](https://docs.ruby-lang.org/en/3.2/Array.html#method-i-index).

Using both here made me wonder if they did different things.

```suggestion
  def back_path
    current_mentor_training_index = mentor_trainings.find_index(mentor_training)
    
    if current_mentor_training_index.zero?
      edit_claims_school_claim_mentor_path(claim.school, claim, mentor_training.mentor)
    else
      edit_claims_school_claim_mentor_training_path(
        claim.school,
        claim,
        mentor_trainings[current_mentor_training_index - 1],
      )
    end
  end
```

app/models/mentor_training.rb
```diff
@@ -28,4 +29,24 @@ class MentorTraining < ApplicationRecord
...
+    allow_nil: true,
+  )
+  validates(
+    :custom_hours_completed,
+    numericality: { greater_than: 0, less_than_or_equal_to: 20, only_integer: true },
+    allow_nil: true,
+  )
+
+  scope :without_hours, -> { where(hours_completed: nil, custom_hours_completed: nil).order_by_mentor_full_name }
+  scope :order_by_mentor_full_name, -> { includes(:mentor).order("mentors.first_name", "mentors.last_name") }
```
JamieCleare2525, line 45: use your `order_by_mentor_full_name` scope here

app/views/claims/schools/claims/mentor_trainings/edit.html.erb
CatalinVoineag: I think we still need the `custom_hours_completed` but it will be just a attribute on the form object not on the mentor training model. This is because we need to target it for errors

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_02_22_135527) do
+ActiveRecord::Schema[7.1].define(version: 2024_02_27_150645) do
```
Nitemaeric, line 13: This is likely a remainder from the deleted migration.

spec/models/mentor_spec.rb
```diff
@@ -43,6 +43,20 @@
...
+      it "returns the mentors ordered by full name" do
+        mentor1 = create(:mentor, first_name: "Anne", last_name: "Doe")
+        mentor2 = create(:mentor, first_name: "Anne", last_name: "Smith")
+        mentor3 = create(:mentor, first_name: "John", last_name: "Doe")
+
+        expect(described_class.order_by_full_name).to eq(
+          [mentor1, mentor2, mentor3],
+        )
+      end
+    end
```
Nitemaeric, line 57: To ensure that this ordering is explicit, can we switch the order of definition compared to assertion please?

```suggestion
      it "returns the mentors ordered by full name" do
        mentor1 = create(:mentor, first_name: "Anne", last_name: "Smith")
        mentor2 = create(:mentor, first_name: "Anne", last_name: "Doe")
        mentor3 = create(:mentor, first_name: "John", last_name: "Doe")

        expect(described_class.order_by_full_name).to eq(
          [mentor2, mentor1, mentor3],
        )
      end
    end
```

This removes any ambiguity that it's just returning in the order it was defined.

app/forms/claim/mentor_training_form.rb
```diff
@@ -0,0 +1,80 @@
+class Claim::MentorTrainingForm < ApplicationForm
+  DEFAULT_HOURS_COMPLETED = %w[6 20].freeze
+  attr_accessor :claim, :mentor_training, :hours_completed, :custom_hours_completed
+
+  validate :hours_completed_presence
```
Nitemaeric, line 5: It looks like you need to add a few pieces of custom logic for this validation. Could you do?

```suggestion
  validates :hours_completed, presence: true, unless: :custom_hours_completed?
  validates :custom_hours_completed, presence: true, unless: :hours_completed?
```

Nitemaeric (commented):
Some minor comments.

app/forms/claim/mentor_training_form.rb
CatalinVoineag: Good idea but this needs to be 
```
  validates :hours_completed, presence: true, unless: :custom_hours?
  validates :custom_hours_completed, presence: true, if: :custom_hours?

Nitemaeric (approved):
Nice! 🚀

JamieCleare2525 (approved):

PR #333, opened 2024-02-29
As a Support User, I want to download a CSV of claims / payment details

gms-gs:
## Context

Allow a support use to download all claims

Adds a button on the Claims page to allow all claims to be downloaded

## Guidance to review

- Sign in as support user
- Go to claims
- Click Download claims

## Link to Trello card

https://trello.com/c/znGU9fJy/191-as-a-support-user-i-want-to-download-a-csv-of-claims-payment-details

## Things to check

## Screenshots
Screenshot 2024-03-04 at 20 22 45

### DIFF
--- a/app/services/claims/generate_claims_csv.rb
+++ b/app/services/claims/generate_claims_csv.rb
@@ -1,0 +1,22 @@
+require "csv"
+
+class Claims::GenerateClaimsCsv
+  include ServicePattern
+
+  HEADERS = %w[school_name school_urn amount claim_reference_number].freeze
+
+  def call
+    CSV.generate(headers: true) do |csv|
+      csv << HEADERS
+
+      Claim.find_each do |claim|
+        csv << [
+          claim.school.name,
+          claim.school.urn,
+          claim.amount.format(symbol: false, decimal_mark: ".", no_cents: false),
+          claim.reference,
+        ]
+      end
+    end
+  end
+end
--- a/app/views/claims/support/claims/index.html.erb
+++ b/app/views/claims/support/claims/index.html.erb
@@ -7,6 +7,8 @@
       <h1 class="govuk-heading-l"><%= t(".heading") %></h1>
 
       <% if @claims.any? %>
+        <%= govuk_button_to(t(".download_csv"), download_csv_claims_support_claims_path, method: :get) %>
+
         <%= govuk_table do |table| %>
           <% table.with_head do |head| %>
             <% head.with_row do |row| %>
--- a/spec/system/claims/support/claims/download_claims_csv_spec.rb
+++ b/spec/system/claims/support/claims/download_claims_csv_spec.rb
@@ -1,0 +1,43 @@
+require "rails_helper"
+
+RSpec.describe "Download claims CSV", type: :system, service: :claims do
+  let(:support_user) { create(:claims_support_user) }
+
+  before do
+    create_claim
+    user_exists_in_dfe_sign_in(user: support_user)
+    given_i_sign_in
+  end
+
+  scenario "Support user visits the claims index page and downloads CSV" do
+    given_i_am_on_the_claims_index_page
+    when_i_click_download_claims
+    then_i_receive_a_csv_file
+  end
+
+  private
+
+  def given_i_sign_in
+    visit sign_in_path
+    click_on "Sign in using DfE Sign In"
+  end
+
+  def given_i_am_on_the_claims_index_page
+    click_on("Claims")
+  end
+
+  def when_i_click_download_claims
+    click_on("Download claims")
+  end
+
+  def then_i_receive_a_csv_file
+    expect(response_headers["Content-Type"]).to eq "text/csv"
+  end
+
+  def create_claim
+    region = create(:region, name: "Inner London", claims_funding_available_per_hour: Money.from_amount(53.60, "GBP"))
+    school = create(:claims_school, :claims, name: "School name 1", region_id: region.id, urn: "1234")
+
+    create(:claim, school_id: school.id, reference: "12345678")
+  end
+end
--- a/app/controllers/claims/schools/claims_controller.rb
+++ b/app/controllers/claims/schools/claims_controller.rb
@@ -36,7 +36,7 @@ class Claims::Schools::ClaimsController < Claims::ApplicationController
   def confirm; end
 
   def submit
-    Claim::Submit.call(claim: @claim)
+    Claims::Submit.call(claim: @claim)
 
     redirect_to confirm_claims_school_claim_path(@school, @claim)
   end
--- a/app/controllers/claims/support/claims_controller.rb
+++ b/app/controllers/claims/support/claims_controller.rb
@@ -8,6 +8,12 @@ class Claims::Support::ClaimsController < Claims::Support::ApplicationController
 
   def show; end
 
+  def download_csv
+    csv_data = Claims::GenerateClaimsCsv.call
+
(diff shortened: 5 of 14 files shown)


spec/services/claim/calculate_amount_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+describe Claim::CalculateAmount do
+  let!(:claim) { create(:claim, school_id: school.id, reference: "12345678") }
+
+  let!(:region) { create(:region, name: "Inner London", claims_funding_available_per_hour: Money.from_amount(53.60, "GBP")) }
+  let!(:school) { create(:claims_school, :claims, name: "School name 1", region_id: region.id, urn: "1234") }
+
+  describe "#call" do
+    it "calculates the claim amount" do
+      create(:mentor_training, claim_id: claim.id, hours_completed: 10)
```
CatalinVoineag, line 11: You can just pass claim here as the factory should know about the association.
```suggestion
      create(:mentor_training, claim:, hours_completed: 10)
```

spec/services/generate_claims_csv_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  before do
+    region1 = create(:region, name: "Inner London", claims_funding_available_per_hour: Money.from_amount(53.60, "GBP"))
+    region2 = create(:region, name: "Outer London", claims_funding_available_per_hour: Money.from_amount(48.25, "GBP"))
+    region3 = create(:region, name: "Fringe", claims_funding_available_per_hour: Money.from_amount(45.10, "GBP"))
+
+    school1 = create(:claims_school, :claims, name: "School name 1", region_id: region1.id, urn: "1234")
+    school2 = create(:claims_school, :claims, name: "School name 2", region_id: region2.id, urn: "5678")
+    school3 = create(:claims_school, :claims, name: "School name 3", region_id: region3.id, urn: "5679")
+
+    claim1 = create(:claim, school_id: school1.id, reference: "12345678")
```
CatalinVoineag, line 15: Same here you can just pass the school

app/views/claims/support/claims/index.html.erb
```diff
@@ -6,6 +6,8 @@
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l"><%= t(".heading") %></h1>
 
+      <%= govuk_button_to(t(".download_csv"), download_csv_claims_support_claims_path, method: :get) %>
```
CatalinVoineag, line 9: We probably should not show this button if there are no claims

CatalinVoineag (commented):
Minor comments

CatalinVoineag (approved):
Nice, thank you

app/services/generate_claims_csv.rb
```diff
@@ -0,0 +1,19 @@
...
+  include ServicePattern
+
+  HEADERS = %w[school_name school_urn amount claim_reference_number].freeze
+
+  def call
+    CSV.generate(headers: true) do |csv|
+      csv << HEADERS
+
+      Claim.find_each do |claim|
+        claim_amount = Claim::CalculateAmount.call(claim:)
```
ollietreend, line 13: I realise @Nitemaeric [asked you](https://github.com/DFE-Digital/itt-mentor-services/pull/333#discussion_r1511406640) to move `claim_amount` into a Service.

My personal instinct would have been to implement it as a method on the `Claim` model directly, rather than as a standalone service. It feels like that would be a more 'obvious' way to get the claim amount.

For example, so I could call `claim.amount` and receive a `Money` object. And then call `.format` within the `GenerateClaimsCsv` service since that's a formatting/output concern. Something like:

```suggestion
        claim_amount = claim.amount.format(symbol: false, decimal_mark: ".", no_cents: false)
```

But this is just my opinion – no need to make any changes if you're happy with the current approach! 🙂

spec/services/claim/calculate_amount_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+describe Claim::CalculateAmount do
```
ollietreend, line 3:
```suggestion
describe Claims::CalculateAmount do
```

spec/services/generate_claims_csv_spec.rb
```diff
@@ -0,0 +1,44 @@
+require "rails_helper"
+
+RSpec.describe GenerateClaimsCsv do
```
ollietreend, line 3:
```suggestion
RSpec.describe Claims::GenerateClaimsCsv do
```

spec/system/claims/support/claims/download_claims_csv_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    create_claim
+    user_exists_in_dfe_sign_in(user: support_user)
+    given_i_sign_in
+  end
+
+  scenario "Support user visits the claims index page and downloads CSV" do
+    when_i_visit_claim_index_page
+    then_i_click_call_to_action
+    check_correct_path_called
+  end
```
ollietreend, line 16: This isn't strictly correct usage of [Given/When/Then](https://www.agilealliance.org/glossary/given-when-then/) syntax.

I think something like this could be a better fit for what you're trying to prove with this test:

```suggestion
  scenario "Support user visits the claims index page and downloads CSV" do
    given_i_am_on_the_claims_index_page
    when_i_click_download_claims
    then_i_receive_a_csv_file
  end
```

Plus I think the test might be more valuable if we check that a CSV file is sent back to the user. It would prove that the endpoint actually produces a CSV, rather than just checking the page path:

```ruby
def then_i_receive_a_csv_file
  expect(response_headers["Content-Type"]).to eq "text/csv"
end
```

ollietreend (changes requested):
Nice work on this @gms-gs.

I've added some comments, but some of them are just opinions. So please don't feel like you need to act upon every comment.

I've clicked "request changes" because I think it's important that the `Claim` module is changed to `Claims` (and the `claim` directory renamed to `claims`), so it's consistent with the existing convention.

app/services/claims/generate_claims_csv.rb
```diff
@@ -0,0 +1,17 @@
...
+  include ServicePattern
+
+  HEADERS = %w[school_name school_urn amount claim_reference_number].freeze
+
+  def call
+    CSV.generate(headers: true) do |csv|
+      csv << HEADERS
+
+      Claim.find_each do |claim|
+        csv << [claim.school.name, claim.school.urn, claim.amount.format(symbol: false, decimal_mark: ".", no_cents: false), claim.reference]
```
ollietreend, line 13: Minor readability nitpick. This line's getting a bit long now. It's absolutely not a deal-breaker, but @CatalinVoineag might appreciate it as a Vim/Emacs(??) user.

```suggestion
        csv << [
          claim.school.name,
          claim.school.urn,
          claim.amount.format(symbol: false, decimal_mark: ".", no_cents: false),
          claim.reference,
        ]
```

ollietreend (approved):
Excellent work 🎉  thanks for addressing my comments so quickly @gms-gs! ✅ 

I've left one final suggestion (sorry! 😅) but I've approved because it's purely a stylistic nitpick. Feel free to ignore and merge ahead if you'd prefer – I won't be offended! 😄

app/services/claims/generate_claims_csv.rb
CatalinVoineag: Neovim user to be specific 😄 :neckbeard:

PR #337, opened 2024-03-04
Claim Submit and generate reference

CatalinVoineag:
## Context

This is the last step in the create claim flow, the submit a claim functionality. For now we decided to use a Random number of 8 digits for the reference. This PR also adds a submitted_at datetime column to claim model

Migration
New Claims::Submit service
New Submit Page
Random 8 digits reference

## Guidance to review

Pull or go on review app
Go through the `Add claim` flow and submit a claim
You should get a reference back
And the claim you just created should be on the index page of the school/claims

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/4451e9eb-906c-496e-af53-38abf645964e

### DIFF
--- a/app/controllers/claims/schools/claims_controller.rb
+++ b/app/controllers/claims/schools/claims_controller.rb
@@ -1,6 +1,6 @@
 class Claims::Schools::ClaimsController < Claims::ApplicationController
   include Claims::BelongsToSchool
-  before_action :set_claim, only: %i[show check]
+  before_action :set_claim, only: %i[show check confirm submit]
   before_action :authorize_claim
 
   helper_method :claim_provider_form
@@ -31,6 +31,14 @@ class Claims::Schools::ClaimsController < Claims::ApplicationController
     else
       render :edit
     end
+  end
+
+  def confirm; end
+
+  def submit
+    Claim::Submit.call(claim: @claim)
+
+    redirect_to confirm_claims_school_claim_path(@school, @claim)
   end
 
   private
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -2,16 +2,19 @@
 #
 # Table name: claims
 #
-#  id          :uuid             not null, primary key
-#  draft       :boolean          default(FALSE)
-#  created_at  :datetime         not null
-#  updated_at  :datetime         not null
-#  provider_id :uuid
-#  school_id   :uuid             not null
+#  id           :uuid             not null, primary key
+#  draft        :boolean          default(FALSE)
+#  reference    :string
+#  submitted_at :datetime
+#  created_at   :datetime         not null
+#  updated_at   :datetime         not null
+#  provider_id  :uuid
+#  school_id    :uuid             not null
 #
 # Indexes
 #
 #  index_claims_on_provider_id  (provider_id)
+#  index_claims_on_reference    (reference) UNIQUE
 #  index_claims_on_school_id    (school_id)
 #
 # Foreign Keys
@@ -26,5 +29,11 @@ class Claim < ApplicationRecord
   has_many :mentor_trainings
   has_many :mentors, through: :mentor_trainings
 
+  validates :reference, uniqueness: true, allow_nil: true
+
   delegate :name, to: :provider, prefix: true, allow_nil: true
+
+  def submitted_on
+    submitted_at&.to_date
+  end
 end
--- a/app/views/claims/schools/claims/confirm.html.erb
+++ b/app/views/claims/schools/claims/confirm.html.erb
@@ -1,0 +1,21 @@
+<% content_for :page_title, t(".page_title") %>
+<% render "claims/schools/primary_navigation", school: @school, current: :claims %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <%= govuk_panel(title_text: t(".claim_submitted")) do %>
+        <div><%= t(".your_reference_number_id") %></div>
+        <div><strong><%= @claim.reference %></strong></div>
+      <% end %>
+
+      <h2 class="govuk-heading-m"><%= t(".what_happens_next") %></h2>
+      <p class="govuk-body"><%= t(".guidance") %></p>
+
+      <p class="govuk-body">
+        <%= govuk_link_to t(".view_claims"), claims_school_claims_path(@school), no_visited_state: true %>
+      </p>
+    <div>
+  <div>
+</div>
--- a/app/views/claims/schools/claims/index.html.erb
+++ b/app/views/claims/schools/claims/index.html.erb
@@ -26,7 +26,7 @@
       <% table.with_body do |body| %>
         <% @claims.each do |claim| %>
           <% body.with_row do |row| %>
-            <% row.with_cell(text: govuk_link_to(claim.id.first(8), claims_school_claim_path(id: claim.id))) %>
+            <% row.with_cell(text: govuk_link_to(claim.reference, claims_school_claim_path(id: claim.id))) %>
             <% row.with_cell(text: claim.provider_name) %>
             <% row.with_cell do %>
               <ul class="govuk-list">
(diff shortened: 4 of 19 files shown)


app/forms/claim/submit_form.rb
```diff
@@ -0,0 +1,23 @@
...
+      claim.draft = false
+      claim.submitted_at = Time.current
+      claim.reference = generate_reference
+      claim
+    end
+  end
+
+  def generate_reference
+    reference = SecureRandom.random_number(99_999_999) while Claim.where(reference:).present?
+    reference
```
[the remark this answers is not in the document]
CatalinVoineag: No, because we need to call the variable in here `Claim.where(reference:)`

CatalinVoineag:
[Update the "Add a claim" flow](https://trello.com/c/nWP6Fttf/168-update-the-add-a-claim-flow)

app/forms/claim/submit_form.rb
```diff
@@ -0,0 +1,23 @@
...
+    @updated_claim ||= begin
+      claim.draft = false
+      claim.submitted_at = Time.current
+      claim.reference = generate_reference
+      claim
+    end
+  end
+
+  def generate_reference
+    reference = SecureRandom.random_number(99_999_999) while Claim.where(reference:).present?
```
Nitemaeric, line 20:
```suggestion
    reference = SecureRandom.random_number(99_999_999) while Claim.exists?(reference:)
```

db/schema.rb
```diff
@@ -29,7 +29,10 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.uuid "provider_id"
+    t.string "reference"
+    t.datetime "submitted_at"
```
Nitemaeric, line 33: Since this field is added, do we need the `draft` attribute anymore? Could a record without a `submitted_at` value be considered a draft?

spec/models/claim_spec.rb
```diff
@@ -29,6 +32,13 @@
     it { is_expected.to have_many(:mentors).through(:mentor_trainings) }
   end
 
+  context "with validations" do
+    it "validates uniqueness of reference" do
+      claim = build(:claim)
+      expect(claim).to validate_uniqueness_of(:reference).allow_nil
+    end
```
Nitemaeric, line 39: You can use an implicit subject here.

```suggestion
    it { is_expected.to validate_uniqueness_of(:reference).allow_nil }
```

app/controllers/claims/schools/claims_controller.rb
```diff
@@ -33,6 +33,15 @@ def update
...
   end
 
+  def confirm; end
+
+  def submit
+    submit_form = Claim::SubmitForm.new(claim: @claim)
+    submit_form.save!
+
+    redirect_to confirm_claims_school_claim_path(@school, @claim)
+  end
```
Nitemaeric, line 43: This isn't a part of a HTML form. Should we make this a service instead? `Claim::Submit.call(claim: @claim)`

app/views/claims/schools/claims/confirm.html.erb
```diff
@@ -0,0 +1,22 @@
...
+<% render "claims/schools/primary_navigation", school: @school, current: :claims %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+
+      <%= govuk_panel(title_text: t(".claim_submitted")) do %>
+        <%= t(".your_reference_number_id") %>
+        <br>
+        <strong><%= @claim.reference %></strong>
```
Nitemaeric, line 11: Instead of using a `<br>` to add a line break, could we do:

```suggestion
        <div><%= t(".your_reference_number_id") %></div>
        <div><strong><%= @claim.reference %></strong></div>
```

app/views/claims/schools/claims/index.html.erb
```diff
@@ -35,7 +35,7 @@
                 <% end %>
               </ul>
             <% end %>
-            <% row.with_cell(text: l(claim.created_at.to_date, format: :short)) %>
+            <% row.with_cell(text: l(claim.submitted_at.to_date, format: :short)) %>
```
Nitemaeric, line 38: To follow the law of Demeter, could we add this to the model?

```ruby
class Claim
  def submitted_on
    submitted_at&.to_date
  end
end
```

```suggestion
            <% row.with_cell(text: l(claim.submitted_on, format: :short)) %>
```

This also handles `nil` calls.

There's also something I'll note, the `l` method doesn't like `nil` values being passed in. I think we only show records _with_ a `submitted_at` value, but I thought I'd mention it.

db/schema.rb
CatalinVoineag: I agree that the draft attribute is not suitable for our requirements anymore. I do think we should leave it like this for this PR and refactor the claim to use a status enum

CatalinVoineag: A draft claim cannot have a submitted at

app/views/claims/schools/claims/index.html.erb
Nitemaeric: @CatalinVoineag I edited the above _slightly_.

spec/models/claim_spec.rb
CatalinVoineag: I need to build the claim so that it links to a school. The claim cannot exist without a link to a school. I get a PG error saying school_id needs to exist as the validate uniqueness saves the record to db to check if the validation is actually working.

Nitemaeric: If you override `subject`, `is_expected` will automatically reference it too.

app/controllers/claims/schools/claims_controller.rb
Nitemaeric: Sorry, I think I wasn't too clear.

```suggestion
  def submit
    Claim::Submit.call(claim: @claim)

    redirect_to confirm_claims_school_claim_path(@school, @claim)
  end
```

I think the redirection and controller logic should stay put, but the stuff within the `SubmitForm` doesn't handle a form submission, there isn't a form at all actually, it marks a draft as submitted. That logic looks to belong within a service.

app/controllers/claims/schools/claims_controller.rb
```diff
@@ -33,6 +33,14 @@ def update
     end
   end
 
+  def confirm; end
+
+  def submit
+    Claims::Submit.call(claim: @claim)
```
Nitemaeric, line 39: This should reference the model, `Claim`, rather than the service namespace, `Claims`.

This is because we're attempting to `Submit` the `Claim`.

```suggestion
    Claim::Submit.call(claim: @claim)
```

app/services/claims/submit.rb
```diff
@@ -0,0 +1,29 @@
...
+    @updated_claim ||= begin
+      claim.draft = false
+      claim.submitted_at = Time.current
+      claim.reference = generate_reference
+      claim
+    end
+  end
+
+  def generate_reference
+    reference = SecureRandom.random_number(99_999_999) while Claim.exists?(reference:)
```
Nitemaeric, line 26: I'm wondering if this is the _logical_ way to do this... Should we be treating the reference number as a **string** of 8 random numbers instead of a number, then converting it to a string?

```ruby
8.times.map { ("0".."9").to_a.sample }.join
```

Just a thought, but likely less optimal. No changes needed. :)

app/models/claim.rb
```diff
@@ -2,16 +2,19 @@
...
 # Table name: claims
 #
-#  id          :uuid             not null, primary key
-#  draft       :boolean          default(FALSE)
-#  created_at  :datetime         not null
-#  updated_at  :datetime         not null
-#  provider_id :uuid
-#  school_id   :uuid             not null
+#  id           :uuid             not null, primary key
+#  draft        :boolean          default(FALSE)
```
JamieCleare2525, line 6: Why is a `Claim` `default: false` shouldn't this be `default: true`?

PR #338, opened 2024-03-04
Use scopes to order records

JamieCleare2525:
## Context

- Use scopes to order records, so ordering behaviour is consistent between services.

- Add scopes for ordering to Subject/Placement/User models
- Changes to controllers to order records using scopes

## Link to Trello card

https://trello.com/c/wTT6q4NX/160-implement-shared-scopes-for-ordering

### DIFF
--- a/spec/models/placement_spec.rb
+++ b/spec/models/placement_spec.rb
@@ -38,4 +38,21 @@ RSpec.describe Placement, type: :model do
     it { is_expected.to validate_presence_of(:end_date) }
     it { is_expected.to validate_comparison_of(:end_date).is_greater_than(:start_date) }
   end
+
+  describe "scopes" do
+    describe "#order_by_subject_name" do
+      it "returns the placements ordered by subject name" do
+        subject1 = create(:subject, name: "English")
+        placement1 = create(:placement, subjects: [subject1])
+        subject2 = create(:subject, name: "Art")
+        placement2 = create(:placement, subjects: [subject2])
+        subject3 = create(:subject, name: "Science")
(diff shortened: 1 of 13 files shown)


spec/models/placement_spec.rb
```diff
@@ -38,4 +38,21 @@
...
+        subject2 = create(:subject, name: "Art")
+        placement2 = create(:placement, subjects: [subject2])
+        subject3 = create(:subject, name: "Science")
+        placement3 = create(:placement, subjects: [subject3])
+
+        expect(described_class.order_by_subject_name).to eq(
+          [placement2, placement1, placement3],
+        )
+      end
+    end
```
ollietreend, line 56: How about when the Placement has multiple Subjects?

It would be good to see a test which covers this scenario, because I'm not sure if the database level join and sort is going to work as elegantly compared to when it's a 'has one' association 😕 unless Rails is doing some query magic behind the scenes...?

Either way, I think it could be good to have that scenario covered here.

ollietreend (commented):
This is looking great. It's a nice way to remove DB-level sorting from the controller code.

I just have one question about if/how this works for Placements being sorted by Subject name.

spec/models/placement_spec.rb
JamieCleare2525: score removed in [71d6ac4](https://github.com/DFE-Digital/itt-mentor-services/pull/338/commits/71d6ac49b1ab52d327a283a3c6232b8e71fba0c2). Will explore complex ordering in another PR

ollietreend (approved):
Looks good to me 👍🏻

PR #345, opened 2024-03-05
Improve seed data efficiency

JamieCleare2525:
## Context

- Improve speed of Seed Data running
- Improve GIAS CSV Importer when assigning region

## Changes proposed in this pull request

- Improve speed of Seed Data running
- Schools are assigned regions is bulk, not iterated over.

## Guidance to review

- Run `bundle exec rake db:drop db:create db:setup`
- Running seed data should be quicker.

## Link to Trello card

https://trello.com/c/LnEuYhc0/165-seed-data-efficiency-and-clean-up

PR #347, opened 2024-03-05
Add Trust

Kizr:
## Things to check

### DIFF
--- a/app/services/gias_csv_importer.rb
+++ b/app/services/gias_csv_importer.rb
@@ -3,6 +3,7 @@ require "csv"
 class GiasCsvImporter
   include ServicePattern
   OPEN_SCHOOL = "1".freeze
+  SUPPORTED_BY_A_TRUST = "1".freeze
   NON_ENGLISH_ESTABLISHMENTS = %w[8 10 25 24 26 27 29 30 32 37 49 56 57].freeze
   INNER_LONDON_AREAS = [
     "E09000002", # Barking and Dagenham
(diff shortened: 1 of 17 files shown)


app/services/gias_csv_importer.rb
```diff
@@ -87,54 +88,68 @@ def initialize(csv_path)
 
   def call
     invalid_records = []
-    records = []
+    school_records = []
+    trust_records = []
+    trust_associations = Hash.new { |h, k| h[k] = [] }
```
ollietreend, line 93: This is clever! 🤓 I've never seen this before.

app/services/gias_csv_importer.rb
```diff
@@ -171,6 +186,17 @@ def determine_region(district_admin_code)
...
   end
 
+  def associate_schools_to_trusts(trust_data)
+    Rails.logger.debug "Associating schools to trusts... "
+
+    trust_data.each do |uid, urns|
+      trust = Trust.find_by(uid:)
+      next unless trust
+
+      School.where(urn: urns).update_all(trust_id: trust.id)
```
ollietreend, line 196: It's only syntactic sugar, but I think you can simplify this to:

```suggestion
      School.where(urn: urns).update_all(trust:)
```

ollietreend (commented):
Just a couple of little things from me – but I'm not approving because this PR is still in draft.

app/services/gias_csv_importer.rb
JamieCleare2525: I don't think `update_all` accepts this.
As I replied in https://github.com/DFE-Digital/itt-mentor-services/pull/345#discussion_r1513041291, `update_all(region: inner_london)` returns a `TypeError: can't cast Region (TypeError)`

Kizr: Yeah, update_all needs the `id` here, but if I could use the nicer syntax I would!

app/services/gias_csv_importer.rb
Kizr: I've only learnt it myself recently, I think it's an improvement over adding a guard clause to create the empty array should it be missing.

spec/models/trust_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+#
+require "rails_helper"
+
+RSpec.describe Trust, type: :model do
+  describe "associations" do
+    it { is_expected.to have_many(:schools) }
+  end
+
+  describe "validations" do
+    subject { create(:trust) }
```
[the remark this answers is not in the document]
Kizr: Great spot, thank you!

app/services/gias_csv_importer.rb
ollietreend: Oops. I missed @JamieCleare2525's reply on the other PR comment 🙃

PR #370, opened 2024-03-12
Leave date submitted blank for support claim

CatalinVoineag:
## Context

`Claim::Submit` service now accepts params hash. This way we can leave the `submitted_at` nil when a support user creates a claim. The support user cannot submit a claim.

This PR also fixes an edge case when there is a claim with a reference already in place and we submit that claim, instead of creating another reference we use the existing one.

Claim::Submit service
ApplicationHelper
Views
Specs

## Guidance to review

Create a claim as a support user
The date submitted should be blank

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/d8a01577-a25d-477e-b528-d2208b3ee39a

### DIFF
--- a/app/controllers/claims/schools/claims_controller.rb
+++ b/app/controllers/claims/schools/claims_controller.rb
@@ -47,7 +47,10 @@ class Claims::Schools::ClaimsController < Claims::ApplicationController
   def confirm; end
 
   def submit
-    Claims::Submit.call(claim: @claim, draft: false)
+    Claims::Submit.call(
+      claim: @claim,
+      claim_params: { draft: false, submitted_at: Time.current },
+    )
 
     redirect_to confirm_claims_school_claim_path(@school, @claim)
   end
--- a/app/controllers/claims/support/schools/claims_controller.rb
+++ b/app/controllers/claims/support/schools/claims_controller.rb
@@ -45,7 +45,10 @@ class Claims::Support::Schools::ClaimsController < Claims::Support::ApplicationC
   end
 
   def submit
-    Claims::Submit.call(claim: @claim, draft: true)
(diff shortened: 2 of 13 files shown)


JamieCleare2525 (commented):
Something might be going wrong with the reference link for the claim.
https://track-and-pay-370.test.teacherservices.cloud/support/schools/fff9ddf4-9fde-4c29-b817-fdd67af8929a/claims
screencapture-track-and-pay-370-test-teacherservices-cloud-support-schools-fff9ddf4-9fde-4c29-b817-fdd67af8929a-claims-2024-03-13-09_43_10

CatalinVoineag:
> Something might be going wrong with the reference link for the claim. My assumption is this claim doesn't have a reference number, but given when this review app was spun up, it might be worth investigating. https://track-and-pay-370.test.teacherservices.cloud/support/schools/fff9ddf4-9fde-4c29-b817-fdd67af8929a/claims screencapture-track-and-pay-370-test-teacherservices-cloud-support-schools-fff9ddf4-9fde-4c29-b817-fdd67af8929a-claims-2024-03-13-09_43_10

This is a problem we are aware of. It's because we create a draft claim in DB when the provider is added to the claim, the first step in the claim form journey. 

The reference is generated when you submit the claim. After the check page so there is a edge case where you add a provider to the claim and leave the form going back to the index page. We show the draft claims on the support console  so the claim will not have a reference.

We are planning to fix this by updating the statuses the claim has, basically we need an intermediate status here, an `internal` status when the claim is in this edge case and will transition to draft when the support user finishes creating a claim.

JamieCleare2525 (approved):

PR #372, opened 2024-03-13
Add Import register placement schools into support schools rake task

gms-gs:
## Context

For private beta we want to onboard over 300 schools into our service, this is different from the schools in GIAS(School model). We are talking about the Claims::School model

Add a rake task that allows us to upload these schools

## Guidance to review

Run `'import_register_placement_schools:import[CSV_LOCATION]'`

## Link to Trello card

https://trello.com/c/eTZ8OzLD/237-private-beta-import-register-placement-schools-into-support-schools

## Things to check

### DIFF
--- a/app/services/claims/import_schools.rb
+++ b/app/services/claims/import_schools.rb
@@ -1,0 +1,21 @@
+require "csv"
+
+class Claims::ImportSchools
+  include ServicePattern
+
+  def initialize(csv_file_path:)
+    @csv_file_path = csv_file_path
+  end
+
+  def call
+    urns = CSV.read(csv_file_path, headers: true).map { |row| row["urn"] }
+
+    updated_count = School.where(urn: urns).update_all(claims_service: true)
+
+    Rails.logger.info "#{updated_count} schools updated."
+  end
+
+  private
+
+  attr_reader :csv_file_path
+end
--- a/lib/tasks/import_schools.rake
+++ b/lib/tasks/import_schools.rake
@@ -1,0 +1,8 @@
+namespace :import_schools do
+  desc "Import of schools into the claims domain"
+  task :import, [:csv_file_path] => :environment do |_task, args|
+    csv_file_path = args[:csv_file_path]
+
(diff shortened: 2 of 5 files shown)


app/services/claims/import_register_placement_schools.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  attr_reader :csv_file_path
+
+  def call
+    urns = CSV.read(csv_file_path, headers: true).map { |row| row["urn"] }
+
+    updated_count = School.where(urn: urns).update_all(claims_service: true)
+
+    Rails.logger.debug "#{updated_count} schools updated."
```
CatalinVoineag, line 17:
```suggestion
    Rails.logger.info "#{updated_count} schools updated."
```

I think debug will only print when you have the debug flag enabled but I think this information should be printed regardless of the debug flag

CatalinVoineag (commented):
This looks very good. I'm wondering if the name `Claims::ImportRegisterPlacementSchools` is appropriate here.

To me this is just an import of schools into the claims domain via a csv that has urns in it.

So should we rename this to be more generic, something like `Claims::ImportSchools` or something similar? What do you think?

CatalinVoineag (approved):
Nice!

app/services/claims/import_schools.rb
```diff
@@ -0,0 +1,19 @@
+require "csv"
+
+class Claims::ImportSchools
+  include ServicePattern
+
+  def initialize(csv_file_path:)
+    @csv_file_path = csv_file_path
+  end
+
+  attr_reader :csv_file_path
```
Kizr, line 10: Could this be a private `attr_reader`?

Kizr: As a general rule we should always default to using `private` unless we need something (an attribute, method, etc) to be publicly accessible, this gives us confidence that we can make changes to the logic without breaking things outside the scope of the class.

Happy for others to weigh in here though!

Kizr (approved):

PR #373, opened 2024-03-13
Refactor claim statuses

CatalinVoineag:
## Context

Previously, the claim status is determined by a boolean draft column on the claims table. As requirements have changed this is not supporting the functionality we need.

We need an `internal` status because we create a claim in DB when the user adds a mentor to a claim. Those claims should not be shown to the user.

The user should see claims that are `draft` or `submitted`, the other two statuses added in this commit.

We could have used the claim values to calculate the status of the claim but the logic would not scale very well so an enum seemed like the correct choice here.

These statuses should also help us with implementing policies.

## Changes

Migrations to add status enum and remove draft boolean (There is no live data so we don't need to migrate data)
Set the correct status in the claim form
Change specs

## Guidance to review

- If you have claims on your local machine, you will need to delete all those claims by running: 
`MentorTraining.destroy_all`
`Claim.destroy_all`
- Create new claims as support and normal user
- Only draft and submitted claims should be shown on claim index pages.

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/a54ff6f0-96c1-46fc-867e-9d4e11827c14

### DIFF
--- a/app/components/claim/status_tag_component.rb
+++ b/app/components/claim/status_tag_component.rb
@@ -8,9 +8,19 @@ class Claim::StatusTagComponent < ApplicationComponent
   end
 
   def call
-    css_class = claim.draft ? "govuk-tag--grey" : "govuk-tag--blue"
-    text = claim.draft ? t(".draft") : t(".submitted")
+    content_tag(:p, claim.status.humanize, class: "govuk-tag #{css_class}")
+  end
 
-    content_tag(:p, text, class: "govuk-tag #{css_class}")
+  private
+
+  def css_class
+    style_status_classes.fetch(claim.status)
+  end
+
+  def style_status_classes
+    {
+      draft: "govuk-tag--grey",
+      submitted: "govuk-tag--blue",
+    }.with_indifferent_access
   end
 end
--- a/app/forms/claim/provider_form.rb
+++ b/app/forms/claim/provider_form.rb
@@ -20,7 +20,7 @@ class Claim::ProviderForm < ApplicationForm
   def updated_claim
     @updated_claim ||= begin
       claim.provider_id = provider_id
-      claim.draft = true
+      claim.status = :internal
       claim.created_by = current_user
       claim
     end
--- a/app/models/claim.rb
+++ b/app/models/claim.rb
@@ -4,8 +4,8 @@
 #
 #  id              :uuid             not null, primary key
 #  created_by_type :string
-#  draft           :boolean          default(FALSE)
 #  reference       :string
+#  status          :enum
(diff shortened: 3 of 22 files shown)


app/models/claim.rb
```diff
@@ -33,8 +33,15 @@ class Claim < ApplicationRecord
   has_many :mentor_trainings
   has_many :mentors, through: :mentor_trainings
 
+  validates :status, presence: true
   validates :reference, uniqueness: { case_sensitive: false }, allow_nil: true
 
+  scope :not_internal, -> { where(status: %w[draft submitted]) }
```
Kizr, line 39:
```suggestion
  scope :not_internal, -> { where.not(status: :internal) }
```

Could we use `where.not(...)` here?

app/forms/claim/provider_form.rb
```diff
@@ -20,7 +20,7 @@ def claim
   def updated_claim
     @updated_claim ||= begin
       claim.provider_id = provider_id
-      claim.draft = true
+      claim.status = "internal"
```
Kizr, line 23: Could this be a symbol?

app/components/claim/status_tag_component.rb
```diff
@@ -8,9 +8,17 @@ def initialize(claim:, classes: [], html_attributes: {})
   end
 
   def call
-    css_class = claim.draft ? "govuk-tag--grey" : "govuk-tag--blue"
-    text = claim.draft ? t(".draft") : t(".submitted")
+    css_class = style_status_classes.fetch(claim.status)
```
Kizr, line 11: Could this be moved into a private method?

spec/system/claims/schools/claims/view_claims_spec.rb
```diff
@@ -35,6 +44,10 @@
     )
   end
 
+  before do
+    create(:claim, school:) # internal claim
```
Kizr, line 48: Could we set the status explicitly here to remove the need for the comment?

Kizr (commented):
Couple of non-blocking bits, I noticed we're using a lot of strings for enums, is that a stylistic choice in the application?

CatalinVoineag:
[Show draft claims on the school claims index page](https://trello.com/c/vgtrC5YZ/231-show-draft-claims-on-the-school-claims-index-page)

CatalinVoineag:
[Refactor the claim status](https://trello.com/c/c1FaD4hC/248-refactor-the-claim-status)

Kizr (approved):
Awesome stuff!

PR #377, opened 2024-03-14
As a user I would like to receive a confirmation email that my claim has been submitted

gms-gs:
## Context

We want to email the current user with a confirmation email containing relevant information about the claim being made.

We send an email to the current user when a claim has been submitted

## Guidance to review

Submit a claim, and use mailcatcher to intercept the email. Or pull the branch locally and submit a claim and look in the console to see it being called/sent.

## Link to Trello card

https://trello.com/c/CWB019sD/247-as-a-user-i-would-like-to-receive-a-confirmation-email-that-my-claim-has-been-submitted

## Things to check

### DIFF
--- a/app/controllers/claims/schools/claims_controller.rb
+++ b/app/controllers/claims/schools/claims_controller.rb
@@ -50,6 +50,7 @@ class Claims::Schools::ClaimsController < Claims::ApplicationController
     Claims::Submit.call(
       claim: @claim,
       claim_params: { status: :submitted, submitted_at: Time.current },
+      user: current_user,
     )
 
     redirect_to confirm_claims_school_claim_path(@school, @claim)
--- a/app/controllers/claims/support/schools/claims_controller.rb
+++ b/app/controllers/claims/support/schools/claims_controller.rb
@@ -48,6 +48,7 @@ class Claims::Support::Schools::ClaimsController < Claims::Support::ApplicationC
     Claims::Submit.call(
       claim: @claim,
       claim_params: { status: :draft },
+      user: current_user,
     )
 
     redirect_to claims_support_school_claims_path(@school), flash: { success: t(".success") }
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -10,4 +10,15 @@ class UserMailer < ApplicationMailer
                  subject: t(".subject", organisation_name: organisation.name),
                  body: t(".body", user_name: user.full_name, organisation_name: organisation.name, service_name:)
   end
+
+  def claim_submitted_notification(user, claim)
+    link_to_claim = claims_school_claim_url(id: claim.id, school_id: claim.school.id)
+
+    notify_email to: user.email,
+                 subject: t(".subject"),
+                 body: t(".body",
+                         reference: claim.reference,
+                         amount: claim.amount.format(symbol: true, decimal_mark: ".", no_cents: false),
+                         link_to_claim:)
+  end
 end
--- a/app/services/claims/submit.rb
+++ b/app/services/claims/submit.rb
@@ -1,18 +1,32 @@
 class Claims::Submit
   include ServicePattern
 
-  def initialize(claim:, claim_params:)
+  def initialize(claim:, claim_params:, user:)
     @claim = claim
     @claim_params = claim_params
+    @user = user
   end
 
   def call
     updated_claim.save!
+
+    # We will be implementing a different email for support users in another ticket
+    if !user.support_user? && status_changed_to_submitted?
+      send_claim_submitted_notification_email
+    end
   end
 
   private
 
-  attr_reader :claim, :claim_params
+  attr_reader :claim, :claim_params, :user
+
+  def send_claim_submitted_notification_email
+    UserMailer.with(service: user.service).claim_submitted_notification(user, claim).deliver_later
+  end
+
+  def status_changed_to_submitted?
+    updated_claim.saved_change_to_status? && updated_claim.submitted?
+  end
 
   def updated_claim
     @updated_claim ||= begin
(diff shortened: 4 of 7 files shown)


app/mailers/user_mailer.rb
```diff
@@ -10,4 +10,13 @@ def user_membership_destroyed_notification(user, organisation)
                  subject: t(".subject", organisation_name: organisation.name),
                  body: t(".body", user_name: user.full_name, organisation_name: organisation.name, service_name:)
   end
+
+  def user_claim_submitted_notification(user, claim, link_to_claim)
```
JamieCleare2525, line 14:
```suggestion
  def claim_submitted_notification(user, claim, link_to_claim)
```
According to the article @Nitemaeric send (https://katanacode.com/blog/ruby-on-rails-action-mailer-how-to-name-mailer-methods/), this should be the resource which is changed (`claim`). So you don't need the `user`.
In the existing examples it's because the `UserMembership` is being changed.

app/controllers/claims/schools/claims_controller.rb
```diff
@@ -52,6 +52,7 @@ def submit
       claim_params: { status: :submitted, submitted_at: Time.current },
     )
 
+    User::ClaimConfirmation.call(user: current_user, claim: @claim, link_to_claim: claims_school_claim_url(id: @claim.id))
```
JamieCleare2525, line 55: Might be worth considering moving this to an `after_save` in the `Claim` model, so that when the `submitted_at` changes from `nil` it sends a confirmation. 
This will prevent spamming emails when someone hits the submit button too many times.

ollietreend: Sorry to jump in here, but I'd caution against performing real-world side effects inside Active Record callbacks. It's usually not desirable. Instead, a service object is probably the more sensible place for it.

As per our [documented convention](https://github.com/DFE-Digital/itt-mentor-services/blob/main/docs/conventions/service_objects.md), we use service objects to perform data manipulation and associated side effects.

In this case, the data manipulation is _updating the claim record_ and the side effect is _sending a confirmation email_.

@gms-gs my suggestion would be that the user confirmation email is triggered from within the `Claims::Submit` service. Given its name, my understanding is that it's the service which handles the submission of claims, and this is a legitimate & desired side effect of that action.

What do you think?

spec/mailers/user_mailer_spec.rb
```diff
@@ -108,4 +108,28 @@
       end
     end
   end
+
+  describe "#user_claim_submitted_notification" do
+    subject(:claim_confirmation_email) { described_class.with(service: user.service).user_claim_submitted_notification(user, claim, link_to_claim) }
+
+    context "when a claim has been submitteds" do
```
CatalinVoineag, line 115: `submitteds` 😄

CatalinVoineag (approved):

app/services/claims/submit.rb
```diff
@@ -1,18 +1,25 @@
...
     @claim = claim
     @claim_params = claim_params
+    @user = user
+    @link_to_claim = link_to_claim
   end
 
   def call
     updated_claim.save!
+
+    unless user.type == "Claims::SupportUser"
```
JamieCleare2525, line 14: You'll still need to guard against spamming the same endpoint. This should return unless claim changed to `:submitted`

app/controllers/claims/schools/claims_controller.rb
```diff
@@ -50,6 +50,8 @@ def submit
     Claims::Submit.call(
       claim: @claim,
       claim_params: { status: :submitted, submitted_at: Time.current },
+      user: current_user,
+      link_to_claim: claims_school_claim_url(id: @claim.id),
```
JamieCleare2525, line 54: You shouldn't need to pass the link here. You can do this in the mailer.

app/controllers/claims/support/schools/claims_controller.rb
```diff
@@ -48,6 +48,8 @@ def submit
     Claims::Submit.call(
       claim: @claim,
       claim_params: { status: :draft },
+      user: current_user,
+      link_to_claim: claims_school_claim_url(id: @claim.id),
```
JamieCleare2525, line 52: You shouldn't need to pass the link here. You can do this in the mailer.

app/services/claims/submit.rb
```diff
@@ -1,18 +1,34 @@
...
+  attr_reader :claim, :claim_params, :user
+
+  def send_claim_submitted_notification_email
+    UserMailer.with(service: user.service).claim_submitted_notification(user, claim).deliver_later
+  end
+
+  def status_changed_to_submitted
+    return false if updated_claim.saved_change_to_status.nil?
+
+    true if updated_claim.saved_change_to_status[1] == "submitted"
```
JamieCleare2525, line 30: Unnecessary `true`, when `updated_claim.saved_change_to_status[1] == "submitted"` will return a boolean

ollietreend: It might be more elegant to write this logic as:

```ruby
  def status_changed_to_submitted?
    updated_claim.saved_change_to_status? && updated_claim.submitted?
  end
```

Note that the method name ends with a `?` since this is technically a [predicate method](https://pragmati.st/2012/03/24/the-elements-of-ruby-style-predicate-methods/).

app/services/claims/submit.rb
```diff
@@ -1,18 +1,34 @@
...
     @claim = claim
     @claim_params = claim_params
+    @user = user
   end
 
   def call
     updated_claim.save!
+
+    # We will be implementing a different email for support users in another ticket
+    if user.type != "Claims::SupportUser" && status_changed_to_submitted
```
ollietreend, line 14: You could use the `support_user?` method for this:

```suggestion
    if !user.support_user? && status_changed_to_submitted
```

ollietreend (approved):

PR #380, opened 2024-03-18
Accessibility, Cookies and Terms pages

CatalinVoineag:
## Context

Some of the content will change after content review but most of it will remain in place

- Accessibility & cookies & terms pages

## Guidance to review

Review the content on these pages.
Prototype: https://itt-mentoring-beta-dev-a238297da236.herokuapp.com/accessibility

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/c06401c2-2a1a-4f80-8d4f-0d56a5c05d09

### DIFF
--- a/app/helpers/footer_helper.rb
+++ b/app/helpers/footer_helper.rb
@@ -2,10 +2,10 @@ module FooterHelper
   def footer_meta_items
     [
       { text: t(".guidance"), href: "#" },
-      { text: t(".accessibility"), href: "#" },
-      { text: t(".cookies"), href: "#" },
+      { text: t(".accessibility"), href: accessibility_path },
+      { text: t(".cookies"), href: cookies_path },
       { text: t(".privacy_policy"), href: "#" },
-      { text: t(".terms_and_conditions"), href: "#" },
+      { text: t(".terms_and_conditions"), href: terms_path },
     ]
   end
 end
--- a/app/views/claims/pages/accessibility.html.erb
+++ b/app/views/claims/pages/accessibility.html.erb
@@ -1,0 +1,64 @@
+<% content_for :page_title, t(".page_title") %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l"><%= t(".page_title") %></h1>
+    </div>
+  </div>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <p class="govuk-body">
+        This service is run by the Becoming a Teacher team at the Department for Education. We want as many people as possible to be able to use the service. For example, that means users should be able to:
+      </p>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li>change colours, contrast levels and fonts</li>
+        <li>zoom in up to 300% without the text spilling off the screen</li>
+        <li>navigate most of the service using just a keyboard</li>
+        <li>navigate most of the service using speech recognition software></li>
+        <li>listen to most of the content on the service using a screen reader (including the most recent versions of JAWS, NVDA and VoiceOver)</li>
+      </ul>
+
+      <p class="govuk-body">We’ve also made the text on the service as simple as possible to understand.</p>
+      <p class="govuk-body">
+        <%= govuk_link_to("AbilityNet", "https://mcmw.abilitynet.org.uk/") %> has advice on making your device easier to use if you have a disability.
+      </p>
+
+      <h2 class="govuk-heading-m">Technical information about this service’s accessibility</h2>
+      <p class="govuk-body">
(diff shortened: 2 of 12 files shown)


app/views/claims/pages/accessibility.html.erb
```diff
@@ -0,0 +1,80 @@
...
+          ".web_accessibility_guidelines_html",
+          link_to: govuk_link_to(
+            t(".web_accessibility_guidelines"),
+            "https://www.w3.org/TR/WCAG21/",
+          ),
+        ) %>
+      </p>
+
+      <ul class="govuk-list govuk-list--bullet">
+        <li>Item 1</li>
```
CatalinVoineag, line 43: Placeholder until we get more content on this

CatalinVoineag:
[Footer links](https://trello.com/c/tBalE7NG/254-footer-links)

app/helpers/footer_helper.rb
```diff
@@ -2,10 +2,10 @@ module FooterHelper
   def footer_meta_items
     [
       { text: t(".guidance"), href: "#" },
-      { text: t(".accessibility"), href: "#" },
-      { text: t(".cookies"), href: "#" },
+      { text: t(".accessibility"), href: claims_accessibility_path },
```
JamieCleare2525, line 5: This isn't going to work, because Placements also use this `footer_helper` with the `FooterComponent` in the `layouts/_footer.html.erb`
Could you make this service conditional.

CatalinVoineag: Good catch

config/locales/en/claims/pages.yml
```diff
@@ -17,5 +17,108 @@ en:
...
+        text_on_service: We’ve also made the text on the service as simple as possible to understand.
+        ability_net: AbilityNet
+        ability_net_advice_html: "%{link_to} has advice on making your device easier to use if you have a disability."
+        technical_information: Technical information about this service’s accessibility
+        technical_information_guidance: The Becoming a Teacher team (DfE) is committed to making Claim funding for mentors accessible, in accordance with the Public Sector Bodies (Websites and Mobile Applications) (No. 2) Accessibility Regulations 2018.
+        web_accessibility_guidelines_html: This website is partially compliant with the  %{link_to} AA standard, due to the non-compliances listed below.
+        web_accessibility_guidelines: Web Content Accessibility Guidelines (WCAG) version 2.1
+        service_design_html: The service was designed using the %{link_to}, which conforms to the WCAG 2.1 AA standard. To ensure the service fully meets the WCAG 2.1 AA standard, we are currently arranging an accessibility audit. We will update this page following the audit, and publish details of any accessibility issues identified.
```
ollietreend, line 35: I'm not entirely sure that locale files are the right tool for these pages. These are static pages of long-form content, rather than typical 'web app' pages made up of buttons and structured strings of UI labels, etc.

I think this could introduce friction when it comes to content changes. For example, removing paragraphs, adding new ones, or restructuring them altogether will lead to lots of cognitive load. Key names could fall out of sync with the content the represent, and the file could easily end up with orphaned keys, etc.

Instead, have you considered using [Rails localised views](https://guides.rubyonrails.org/i18n.html#localized-views) for these content-heavy pages?

From the Rails guide:

> You can make use of this feature, e.g. when working with a large amount of static content, which would be clumsy to put inside YAML or Ruby dictionaries. Bear in mind, though, that any change you would like to do later to the template must be propagated to all of them.

Since we only actually support one locale (English), and have no plans to change that in the future, I think this could be the most pragmatic approach for these content-heavy pages.

What do you think?

spec/system/claims/terms_page_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+      "time or for any period. From time to time, we may restrict access to all "\
+      "or some parts of the Service to users who have registered with us.\n"\
+      "If you choose, or are provided with, a user verification code, password or "\
+      "any other piece of information as part of our security procedures, you must "\
+      "treat such information as confidential, and you must not disclose it to any third party.\n"\
+      "We reserve the right to restrict or deny you access to all or some part "\
+      "of the Service if in our opinion you have failed to comply with these Terms and Conditions.\n"\
+      "The Department for Education withholds the right to withdraw without notice "\
+      "listings that breach these terms and to withdraw the access of users who breach them.",
+    )
```
ollietreend, line 55: I'm not sure how valuable these tests are. Is it really necessary to test the static content of these pages?

I feel like we could probably get just as much value from these tests by just checking that the page loads and has the expected title.

ollietreend (commented):
@CatalinVoineag I've added a couple of comments which I think would be good to talk through.

I also understand that the content for these pages hasn't been finalised yet. Given that, I think it'd be good to keep these pages as flexible as possible so they can easily be updated when the actual content is ready.

config/locales/en/claims/pages.yml
CatalinVoineag: I've moved the content on the `.html` file

ollietreend (approved):
Chef kiss emoji

PR #391, opened 2024-03-19
Claim policies

CatalinVoineag:
## Context

Claim policies

The claim form didn't have any policies, meaning a claim could have been
updated or submitted twice by a user if they would know the url.

This PR fixes these edge cases.

It also will redirect the user to root path if he doesn't have access to a resource. This is because redirecting back to request.referrer can cause a redirection loop if the `request.referrer` is not a path the user has access to as well.

Claim policies
`Claims::Support::Schools::ClaimsController` submit action
Redirect to root_path if no access, only for claims domain

## Guidance to review

Try to submit the claim twice by clicking on the back button on the confirm page
Try to edit a submitted claim. 

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/065235b6-5b99-4f70-949b-d20542ceaeaa

### DIFF
--- a/app/controllers/claims/support/schools/claims_controller.rb
+++ b/app/controllers/claims/support/schools/claims_controller.rb
@@ -1,7 +1,7 @@
 class Claims::Support::Schools::ClaimsController < Claims::Support::ApplicationController
   include Claims::BelongsToSchool
 
-  before_action :set_claim, only: %i[check submit show]
+  before_action :set_claim, only: %i[check draft show edit update]
   before_action :authorize_claim
   helper_method :claim_provider_form
 
@@ -44,7 +44,7 @@ class Claims::Support::Schools::ClaimsController < Claims::Support::ApplicationC
     end
   end
 
-  def submit
+  def draft
     Claims::Submit.call(
       claim: @claim,
       claim_params: { status: :draft },
@@ -84,6 +84,6 @@ class Claims::Support::Schools::ClaimsController < Claims::Support::ApplicationC
   end
 
   def authorize_claim
-    authorize Claim
+    authorize @claim || Claim
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -51,6 +51,13 @@ class ApplicationController < ActionController::Base
   end
 
   def user_not_authorized
-    redirect_back fallback_location: root_path, alert: t("you_cannot_perform_this_action")
+    flash[:alert] = t("you_cannot_perform_this_action")
+    can_be_infinite_redirect = request.url == request.referer
+
+    if can_be_infinite_redirect
+      redirect_to root_path
+    else
+      redirect_back(fallback_location: root_path)
+    end
   end
 end
--- a/app/controllers/claims/schools/claims_controller.rb
+++ b/app/controllers/claims/schools/claims_controller.rb
@@ -1,6 +1,6 @@
 class Claims::Schools::ClaimsController < Claims::ApplicationController
   include Claims::BelongsToSchool
-  before_action :set_claim, only: %i[show check confirm submit]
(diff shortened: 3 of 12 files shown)


CatalinVoineag:
[Implement Pundit policies on the claim form](https://trello.com/c/KhUWNM5s/252-implement-pundit-policies-on-the-claim-form)

.rubocop.yml
```diff
@@ -39,3 +39,12 @@ RSpec/SpecFilePathFormat:
...
   Enabled: true
+
+RSpec/RepeatedExample:
+  Exclude:
+    - spec/policies/*.rb
+
+RSpec/RepeatedDescription:
+  Exclude:
+    - spec/policies/*.rb
+
```
CatalinVoineag, line 50: Testing policies with these rules is frustrating

app/controllers/claims/application_controller.rb
```diff
@@ -1,3 +1,7 @@
 class Claims::ApplicationController < ApplicationController
   after_action :verify_authorized
+
+  def user_not_authorized
```
JamieCleare2525, line 4: Do you need this given the `user_not_authorized` in `ApplicationController`?

JamieCleare2525: > Do you need this given the `user_not_authorized` in `ApplicationController`? I can see you've made a comment in your PR description, see an example were you wouldn't get a referrer. If this is a concern for claims, this will be a concern for both services and perhaps should just be rescued with a redirect back

app/controllers/claims/support/schools/claims_controller.rb
```diff
@@ -44,7 +44,7 @@ def update
     end
   end
 
-  def submit
+  def add
```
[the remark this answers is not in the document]
JamieCleare2525: I would consider renaming this to `draft` or `make_draft`, as this is more accurate to what is actually being done. `add` suggests the same as `create`.

app/controllers/claims/support/schools/claims_controller.rb
```diff
@@ -44,7 +44,7 @@ def update
...
     Claims::Submit.call(
```
JamieCleare2525, line 48: Given that the behaviour for `submitting/drafting/adding` a claim is now vastly different for a `SupportUser`, I would suggest creating a separate service for when a `SupportUser` performs this action.

app/controllers/claims/application_controller.rb
JamieCleare2525: My concern is that your change prevents the `redirect_back` completely within your service. I shouldn't redirect to the `root_path` in all case. Only in that specific use case where as you mention the `request.referrer` creates an infinite loop.

JamieCleare2525: https://medium.com/@infernalmaster/rails-infinite-redirect-problem-4a40d998e194

.rubocop.yml
JamieCleare2525: This change shouldn't be necessary if you wrap the test in context blocks:
```ruby
permissions :edit? do
  context "when user has an internal claim" do
    it "grants access" do
      expect(claim_policy).to permit(user, internal_claim)
    end
  end
  
  context "when the user has a draft claim" do
    it "grants access" do
      expect(claim_policy).to permit(user, draft_claim)
    end
  end

  context "when the user has a submitted claim" do
    it "denies access" do
      expect(claim_policy).not_to permit(user, submitted_claim)
    end
  end
end
```

app/controllers/claims/support/schools/claims_controller.rb
CatalinVoineag: This has been something I looked at. It's not for this PR but I will put a refactor PR on our board

JamieCleare2525 (approved):

PR #392, opened 2024-03-19
[DO NOT MERGE] Add a route that will trigger a stack trace

ollietreend:
So that @saliceti can test how multi-line log entries are handled by Logit.

To trigger the error, run the app and visit the path `/explode`.

When the server us running locally with `bundle exec rails server`, it produces a stack trace in the STDOUT logs like this:

Console output showing a Rails stack trace


ollietreend:
Closing this PR as it's no longer needed.

PR #394, opened 2024-03-20
Add grant conditions to claims

CatalinVoineag:
## Context

We need to add a guidance to our claims service

guidance
list with dash

## Guidance to review

Review content

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/bbd60811-9796-4f82-baf6-af2515a9e0eb

### DIFF
--- a/app/assets/stylesheets/lists.scss
+++ b/app/assets/stylesheets/lists.scss
@@ -1,0 +1,16 @@
+.list--dash {
+  li {
+    $contents-spacing: govuk-spacing(3);
+    position: relative;
+    padding-left: $contents-spacing;
+    padding-right: $contents-spacing;
+
+    &:before {
+      content: "\2013";
+      position: absolute;
+      left: 0;
+      width: govuk-spacing(3);
+      overflow: hidden;
+    }
+  }
+}
--- a/app/helpers/footer_helper.rb
+++ b/app/helpers/footer_helper.rb
@@ -1,7 +1,7 @@
 module FooterHelper
   def footer_meta_items
     [
-      { text: t(".guidance"), href: "#" },
+      { text: t(".#{current_service}_guidance"), href: guidance_path },
       { text: t(".accessibility"), href: accessibility_path },
       { text: t(".cookies"), href: cookies_path },
       { text: t(".privacy_policy"), href: privacy_path },
--- a/app/helpers/routes_helper.rb
+++ b/app/helpers/routes_helper.rb
@@ -74,4 +74,11 @@ module RoutesHelper
       placements: "#",
     }.fetch HostingEnvironment.current_service(request)
   end
+
+  def guidance_path
+    {
+      claims: claims_grant_conditions_path,
+      placements: "#",
+    }.fetch HostingEnvironment.current_service(request)
+  end
 end
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -9,3 +9,4 @@ $govuk-assets-path: "";
 @import "search-form";
 @import "filter-form";
 @import "app-related-aside";
+@import "lists";
--- a/app/views/claims/pages/grant_conditions.html.erb
+++ b/app/views/claims/pages/grant_conditions.html.erb
@@ -1,0 +1,175 @@
+<% content_for :page_title, t(".page_title") %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l"><%= t(".page_title") %></h1>
+
+      <h3 class="govuk-heading-s">Contents</h3>
+      <ul class="govuk-list list--dash govuk-!-margin-bottom-8">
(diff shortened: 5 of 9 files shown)


CatalinVoineag:
[Add grant conditions page](https://trello.com/c/mykN1M70/300-add-grant-conditions-page)

app/assets/stylesheets/lists.scss
```diff
@@ -0,0 +1,15 @@
...
+  li {
+    position: relative;
+
+    &:before {
+      color: govuk-colour("dark-grey");
+      content: "\2013";
+      left: -16px;
+      position: absolute;
+      top: 0;
+    }
```
Kizr, line 13: I have _zero_ knowledge on how we implement CSS correctly within our area, but as I can't easily check this myself due to a lack of domain knowledge, has this been tested on multiple screen sizes with the absolute positioning, e.g. mobile devices?

app/helpers/routes_helper.rb
```diff
@@ -67,4 +67,11 @@ def terms_path
       placements: "#",
     }.fetch HostingEnvironment.current_service(request)
   end
+
+  def guidance_path
+    {
+      claims: claims_grant_conditions_path,
+      placements: "#",
```
Kizr, line 74: Another domain related question (sorry), will the grant conditions need to be accessed from the placements side of the app?

app/helpers/footer_helper.rb
```diff
@@ -1,7 +1,7 @@
 module FooterHelper
   def footer_meta_items
     [
-      { text: t(".guidance"), href: "#" },
+      { text: t(".#{current_service}_guidance"), href: guidance_path },
```
Nitemaeric, line 4: I'm a bit hesitant towards naming this "guidance".

Would it make more sense to do something like:

```ruby
def footer_meta_items
  [
    { text: t(".claims.grant_conditions"), href: claims_grant_conditions_path } if current_service == :claims,
    { text: ..., href: ... },
    ...
  ].compact
end
```

and maybe as the next iteration:

```ruby
def footer_meta_items
  {
    claims: claims_footer_meta_items,
    placements: placements_footer_meta_items
  }.fetch current_service
end

def claims_footer_meta_items
  [
    ...
  ]
end
```

Also, the School Placements' guidance link doesn't have content just yet, https://school-placements-beta-4330a9a75ffc.herokuapp.com/guidance. It may end up changing, like ours did.

ollietreend: Yes please – let's not lock ourselves in to needing the same page names between each service. There's every chance they'll need different sets of static pages.

app/assets/stylesheets/lists.scss
ollietreend: It looks like we're trying to imitate the style of a [detailed guide on GOV.UK](https://www.gov.uk/guidance/content-design/content-types#detailed-guide).

I'd suggest that we use the [Contents list](https://components.publishing.service.gov.uk/component-guide/contents_list) component from the [govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components) gem. That way we'll be using the exact same markup and CSS as is used on the GOV.UK website.

It should be pretty simple to integrate, and then there's less custom CSS for us to maintain and test across browsers & devices, etc. Plus the styles will be consistent with GOV.UK.

CatalinVoineag: Ah, I didn't know there's another gem for components

app/helpers/footer_helper.rb
CatalinVoineag: I like the idea of refactoring this to have claim footer meta items and placements footer meta items.

Let's merge this as it is now and I'll follow this up with another PR. I don't want to do it here as it's a bit of a refactor.

app/assets/stylesheets/lists.scss
CatalinVoineag: I added the styling from the [govuk_publishing_compoenents ](https://github.com/alphagov/govuk_publishing_components). 

We cannot use this gem, it's not compatible with the lates govuk-frontend libarry

app/helpers/routes_helper.rb
CatalinVoineag: No, should be fixed by this https://github.com/DFE-Digital/itt-mentor-services/pull/394/files#r1535466462

app/assets/stylesheets/lists.scss
```diff
@@ -0,0 +1,16 @@
...
+    $contents-spacing: govuk-spacing(3);
+    position: relative;
+    padding-left: $contents-spacing;
+    padding-right: $contents-spacing;
+
+    &:before {
+      content: "\2013";
+      position: absolute;
+      left: 0;
+      width: govuk-spacing(3);
```
Nitemaeric, line 12: 👍 `govuk-spacing`

Nitemaeric (approved):

ollietreend (approved):
Looks good to me, on the understanding that we follow up with [the discussed refactor](https://github.com/DFE-Digital/itt-mentor-services/pull/394#discussion_r1535466462).

PR #396, opened 2024-03-21
Allow support users to remove claims

gms-gs:
## Context

We need to allow the support user to remove claims. Currently the only states the claim can be in are draft and submitted, at least user facing.

Implement archived and discarded status enums

## Guidance to review

- Creare a draft and submitted claim
- View the claim
- Follow the deleted flow

## Link to Trello card

https://trello.com/c/JUhoXxIJ/259-allow-support-users-to-remove-claims

## Things to check

### DIFF
--- a/app/controllers/claims/support/schools/claims_controller.rb
+++ b/app/controllers/claims/support/schools/claims_controller.rb
@@ -1,17 +1,26 @@
 class Claims::Support::Schools::ClaimsController < Claims::Support::ApplicationController
   include Claims::BelongsToSchool
 
-  before_action :set_claim, only: %i[check submit show]
+  before_action :set_claim, only: %i[check submit show remove destroy]
   before_action :authorize_claim
   helper_method :claim_provider_form
 
   def index
-    @pagy, @claims = pagy(@school.claims.not_internal.order_created_at_desc)
+    @pagy, @claims = pagy(@school.claims.not_internal_archived_or_discarded.order_created_at_desc)
   end
 
   def new; end
 
   def show; end
+
+  def remove; end
+
+  def destroy
+    destroy_draft_claim if @claim.draft?
+    discard_submitted_claim if @claim.submitted?
+
+    redirect_to claims_support_school_claims_path(@school, @claim), flash: { success: t(".success") }
+  end
 
   def create
     if claim_provider_form.save
@@ -55,6 +64,15 @@ class Claims::Support::Schools::ClaimsController < Claims::Support::ApplicationC
   end
 
   private
+
+  def destroy_draft_claim
+    @claim.mentor_trainings.destroy_all
+    @claim.destroy!
+  end
+
+  def discard_submitted_claim
+    @claim.update!(status: :discarded)
+  end
 
   def claim_params
     params.require(:claim)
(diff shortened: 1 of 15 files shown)


app/policies/claim_policy.rb
```diff
@@ -11,6 +11,10 @@ def confirm?
     true
   end
 
+  def destroy?
```
CatalinVoineag, line 14: Based on this PR https://github.com/DFE-Digital/itt-mentor-services/pull/391 that introduces policies, can we be a bit more strict here?

https://github.com/DFE-Digital/itt-mentor-services/blob/af7b37e6b1e39d9a72f8122423afcd55d3d2158c/app/policies/claim_policy.rb

I think only a support user can remove a claim.

Can we do something like?
```suggestion
  def destroy?
    user.support_user?
  end
  
  def remove?
    destroy?
  end
```

db/migrate/20240321065230_add_archived_status_to_claims.rb
```diff
@@ -0,0 +1,5 @@
+class AddArchivedStatusToClaims < ActiveRecord::Migration[7.1]
+  def change
+    add_enum_value :claim_status, "archived"
```
CatalinVoineag, line 3: I think we should not add this until we need it

config/locales/en/claims/support/schools/claims.yml
```diff
@@ -28,8 +28,17 @@ en:
             grant_funding: Grant funding
             hours: hours
             page_caption: Claims - %{school_name}
+            remove_claim: Remove claim
           new:
             page_title: Accredited provider - Add claim
+          remove:
+            page_title: Accredited provider - Remove claim
```
CatalinVoineag, line 35: Is the page title here correct? I'm not sure about the `Acredited provider bit` I think it needs to be `Are you sure you want to remove this claim?` At least that's what's on the prototype https://itt-mentoring-beta-dev-a238297da236.herokuapp.com/support/organisations/049fc44e-6876-4884-8a77-d363173ce8d2/claims/undefined/delete

JamieCleare2525 (commented):
Why aren't you using the `discard` gem to soft delete the removed claims?
This would keep the method of "discarding" records across the application consistent.

app/controllers/claims/support/schools/claims_controller.rb
```diff
@@ -56,6 +65,15 @@ def submit
 
   private
 
+  def destroy_draft_claim
+    @claim.mentor_trainings.destroy_all
```
JamieCleare2525, line 69: This should be handled by a dependancy on the `claim` being destroyed

app/models/claim.rb
```diff
@@ -36,11 +36,11 @@ class Claim < ApplicationRecord
   validates :status, presence: true
   validates :reference, uniqueness: { case_sensitive: false }, allow_nil: true
 
-  scope :not_internal, -> { where.not(status: :internal) }
+  scope :not_internal_archived_or_discarded, -> { where.not(status: %i[internal archived discarded]) }
```
JamieCleare2525, line 39: You'd be better off having an `active` scope. Possibly worth moving these to a constant. Or even better, reverse the logic and only include the statuses with are active `ACTIVE_STATUSES = %i[draft submitted]`

PR #399, opened 2024-03-21
Allow support users to remove claims

gms-gs:
## Context

We need to allow the support user to remove claims. Currently the only states the claim can be in are draft and submitted, at least user facing.

Implement the discarded gem for a Claim

## Guidance to review

- Creare a draft and submitted claim
- View the claim
- Follow the deleted flow
- Claim will not appear 
- Claim will be discarded

## Link to Trello card

https://trello.com/c/JUhoXxIJ/259-allow-support-users-to-remove-claims

## Things to check

### DIFF
--- a/app/controllers/claims/support/schools/claims_controller.rb
+++ b/app/controllers/claims/support/schools/claims_controller.rb
@@ -1,17 +1,25 @@
 class Claims::Support::Schools::ClaimsController < Claims::Support::ApplicationController
   include Claims::BelongsToSchool
 
-  before_action :set_claim, only: %i[check draft show edit update]
+  before_action :set_claim, only: %i[check draft show edit update remove destroy]
   before_action :authorize_claim
   helper_method :claim_provider_form
 
   def index
-    @pagy, @claims = pagy(@school.claims.not_internal.order_created_at_desc)
+    @pagy, @claims = pagy(@school.claims.visible.order_created_at_desc)
   end
 
   def new; end
 
   def show; end
+
+  def remove; end
+
+  def destroy
+    @claim.destroy!
+
+    redirect_to claims_support_school_claims_path(@school, @claim), flash: { success: t(".success") }
+  end
 
   def create
     if claim_provider_form.save
--- a/app/views/claims/support/schools/claims/show.html.erb
+++ b/app/views/claims/support/schools/claims/show.html.erb
@@ -76,4 +76,8 @@
         <% end %>
     </div>
   </div>
+
+  <% if policy(@claim).remove? %>
+    <%= govuk_link_to t(".remove_claim"), remove_claims_support_school_claim_path(@school, @claim), class: "app-link app-link--destructive" %>
+  <% end %>
 </div>
--- a/spec/system/claims/support/schools/claims/removing_a_claim_spec.rb
+++ b/spec/system/claims/support/schools/claims/removing_a_claim_spec.rb
@@ -1,0 +1,95 @@
+require "rails_helper"
+
+RSpec.describe "Create claim", type: :system, service: :claims do
+  let!(:school) { create(:claims_school, mentors: [mentor1, mentor2], name: "A School") }
+  let!(:colin) do
+    create(
+      :claims_support_user,
+      :colin,
+      user_memberships: [create(:user_membership, organisation: school)],
+    )
+  end
+  let!(:provider) { create(:provider, :best_practice_network) }
+
+  let(:claims_mentor) { create(:claims_mentor, first_name: "Barry", last_name: "Garlow") }
+
+  let!(:submitted_claim) do
+    create(
+      :claim,
+      :submitted,
+      school:,
+      reference: "12345678",
+      submitted_at: Time.new(2024, 3, 5, 12, 31, 52, "+00:00"),
+      provider:,
+    )
+  end
+
+  let!(:draft_claim) do
+    create(
+      :claim,
+      :draft,
+      school:,
+      reference: "111111111",
+      provider:,
+    )
+  end
+
+  let(:mentor1) { create(:mentor, first_name: "Anne") }
+  let(:mentor2) { create(:mentor, first_name: "Joe") }
+
+  let(:mentor_training) { create(:mentor_training, claim: submitted_claim, mentor: claims_mentor, hours_completed: 6) }
+  let(:mentor_training2) { create(:mentor_training, claim: draft_claim, mentor: claims_mentor, hours_completed: 6) }
+
+  before do
+    user_exists_in_dfe_sign_in(user: colin)
+    given_i_sign_in
+  end
+
+  scenario "When removing a draft claim" do
+    when_i_select_a_school
+    when_i_click_on_claims
+    when_i_visit_the_draft_claim_show_page
+    when_i_click_remove_claim
+    when_i_confirm_removal
+    then_the_claim_has_been_deleted
+  end
+
+  private
+
+  def given_i_sign_in
+    visit sign_in_path
+    click_on "Sign in using DfE Sign In"
+  end
+
+  def when_i_select_a_school
+    click_on "A School"
+  end
+
+  def when_i_click_on_claims
+    within(".app-secondary-navigation__list") do
+      click_on("Claims")
+    end
+  end
+
+  def when_i_visit_the_submitted_claim_show_page
+    click_on submitted_claim.reference
+  end
+
+  def when_i_visit_the_draft_claim_show_page
+    click_on draft_claim.reference
+  end
+
+  def when_i_click_remove_claim
+    click_on "Remove claim"
+  end
+
+  def when_i_confirm_removal
+    expect(page).to have_content("Are you sure you want to remove this claim?")
+    click_on "Remove claim"
+  end
+
+  def then_the_claim_has_been_deleted
+    expect(page).not_to have_content("111111111")
+    expect(page).to have_content("Claim has been removed")
+  end
+end
--- a/app/controllers/claims/schools/claims_controller.rb
+++ b/app/controllers/claims/schools/claims_controller.rb
@@ -6,7 +6,7 @@ class Claims::Schools::ClaimsController < Claims::ApplicationController
   helper_method :claim_provider_form
 
   def index
-    @pagy, @claims = pagy(@school.claims.not_internal.order_created_at_desc)
+    @pagy, @claims = pagy(@school.claims.visible.order_created_at_desc)
   end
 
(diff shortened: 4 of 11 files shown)


app/models/claim.rb
```diff
@@ -30,17 +30,17 @@ class Claim < ApplicationRecord
...
   validates :status, presence: true
   validates :reference, uniqueness: { case_sensitive: false }, allow_nil: true
 
-  scope :not_internal, -> { where.not(status: :internal) }
+  scope :not_internal_or_discarded, -> { where.not(status: %i[internal discarded]) }
   scope :order_created_at_desc, -> { order(created_at: :desc) }
 
   enum :status,
-       { internal: "internal", draft: "draft", submitted: "submitted" },
+       { internal: "internal", draft: "draft", submitted: "submitted", discarded: "discarded" },
```
JamieCleare2525, line 43: Why aren't you using the `discard` gem for making a record discarded? We already have this functionality for `Users`, and should keep the discarding mechanism consistent throughout the service/application

JamieCleare2525: A key feature of using `discarded` is that, should you need to `undiscard` the record you haven't lost it's original state.

app/models/claim.rb
```diff
@@ -30,17 +30,17 @@ class Claim < ApplicationRecord
...
 
-  has_many :mentor_trainings
+  has_many :mentor_trainings, dependent: :destroy
   has_many :mentors, through: :mentor_trainings
 
   validates :status, presence: true
   validates :reference, uniqueness: { case_sensitive: false }, allow_nil: true
 
-  scope :not_internal, -> { where.not(status: :internal) }
+  scope :not_internal_or_discarded, -> { where.not(status: %i[internal discarded]) }
```
JamieCleare2525, line 39: Might be better to have an `active` scope, rather than this very specifically named scope.
Could be beneficially to have a constant (e.g `ACTIVE_STATUSES = %i[draft submitted].freeze`), rather than doing the negative of that search. 
```suggestion
  scope :active, -> { where(status: ACTIVE_STATUSES) }
```

app/controllers/claims/support/schools/claims_controller.rb
```diff
@@ -1,18 +1,27 @@
...
 
   def new; end
 
   def show; end
 
+  def remove; end
+
+  def destroy
+    @claim.destroy! if @claim.draft?
+    discard_submitted_claim if @claim.submitted?
```
JamieCleare2525, line 20: Is there a case where the claim would not be either `draft` or `submitted`?
```suggestion
    @claim.submitted? ? @claim.discard : @claim.destroy!
```

app/policies/claim_policy.rb
```diff
@@ -11,6 +11,14 @@ def submit?
     !user.support_user? && !record.submitted?
   end
 
+  def destroy?
+    user.support_user?
```
JamieCleare2525, line 15: You could additionally ensure that the claim is in one of the states which will allow you do perform this action.
```suggestion
    user.support_user? && (record.submitted? || record.draft?)
```

app/controllers/claims/support/schools/claims_controller.rb
ollietreend: I'm wondering if it'd be worth using soft-deletion everywhere for the Claim model?

It feels a bit unusual to implement soft-delete only for 'submitted' claims, and instead to permanently delete 'draft' claims. I could imagine that being confusing behaviour – for example, if somebody accidentally deletes a 'submitted' claim, we'd be able to help them retrieve it by un-discarding it. But the same wouldn't be true if the claim happened to have been in a 'draft' state, since it would have been permanently deleted.

This is more of a data modelling decision rather than an implementation detail, so I'm wondering if @Nitemaeric has an opinion? It's the sort of thing that's worth getting right upfront.

And as @JamieCleare2525 said, we implemented the [discard](https://github.com/jhawthorn/discard) gem back in PR #222, so it'd be good to follow that approach as a consistent way to soft-delete records within the app.

Nitemaeric: This is a Product requirement we've received, https://trello.com/c/JUhoXxIJ/259-allow-support-users-to-remove-claims#comment-65fbedab17e1be89451976e0. We've decided that "archived" is equal to "soft-deleted" and opted towards using the `discard` gem to manage that.

The support interface will have a way to view "archived" records, so we want to make the distinction between hard-deleting drafts and soft-deleting (archiving) submitted claims.

We _may_ be incorrectly equating "archived" and "soft-deleted". This is open for debate.

app/models/claim.rb
```diff
@@ -30,13 +30,15 @@ class Claim < ApplicationRecord
   belongs_to :provider
   belongs_to :created_by, polymorphic: true
 
-  has_many :mentor_trainings
+  has_many :mentor_trainings, dependent: :destroy
```
Nitemaeric, line 33: Let's add the `shoulda-matchers` modifier, `.dependent`, in the model specs.

app/models/claims/school.rb
```diff
@@ -14,9 +14,9 @@
 #  gender                       :string
 #  group                        :string
 #  last_inspection_date         :date
+#  latitude                     :float
 #  local_authority_code         :string
 #  local_authority_name         :string
-#  latitude                     :float
 #  longitude                    :float
```
Nitemaeric, line 20: This looks unrelated to the PR scope.

app/models/placements/school.rb
```diff
@@ -14,9 +14,9 @@
 #  gender                       :string
 #  group                        :string
 #  last_inspection_date         :date
+#  latitude                     :float
 #  local_authority_code         :string
 #  local_authority_name         :string
-#  latitude                     :float
```
Nitemaeric, line 19: This looks unrelated to the PR scope.

app/models/school.rb
```diff
@@ -14,9 +14,9 @@
 #  gender                       :string
 #  group                        :string
 #  last_inspection_date         :date
+#  latitude                     :float
 #  local_authority_code         :string
 #  local_authority_name         :string
-#  latitude                     :float
```
Nitemaeric, line 19: This looks unrelated to the PR scope.

app/policies/claim_policy.rb
```diff
@@ -11,6 +11,14 @@ def submit?
...
   end
 
+  def destroy?
+    user.support_user? && record.draft?
+  end
+
+  def remove?
+    destroy?
+  end
+
```
Nitemaeric, line 21: This is already in `Claims::ApplicationPolicy`.

app/views/claims/support/schools/claims/show.html.erb
```diff
@@ -71,4 +71,8 @@
         <% end %>
     </div>
   </div>
+
+  <% if @claim.draft? %>
```
Nitemaeric, line 75: We should rely on the policy so we don't need to repeat ourselves.

```suggestion
  <% if policy(@claim).remove? %>
```

spec/factories/schools.rb
```diff
@@ -14,9 +14,9 @@
 #  gender                       :string
 #  group                        :string
 #  last_inspection_date         :date
+#  latitude                     :float
 #  local_authority_code         :string
 #  local_authority_name         :string
-#  latitude                     :float
```
Nitemaeric, line 19: This looks unrelated to the PR scope.

spec/factories/subjects.rb
```diff
@@ -1,3 +1,14 @@
...
+# Table name: subjects
+#
+#  id           :uuid             not null, primary key
+#  code         :string
+#  name         :string           not null
+#  subject_area :enum
+#  created_at   :datetime         not null
+#  updated_at   :datetime         not null
+#
 FactoryBot.define do
```
Nitemaeric, line 12: This looks unrelated to the PR scope.

spec/models/claims/school_spec.rb
```diff
@@ -14,9 +14,9 @@
 #  gender                       :string
 #  group                        :string
 #  last_inspection_date         :date
+#  latitude                     :float
 #  local_authority_code         :string
 #  local_authority_name         :string
-#  latitude                     :float
```
Nitemaeric, line 19: This looks unrelated to the PR scope.

spec/models/placements/school_spec.rb
```diff
@@ -14,9 +14,9 @@
 #  gender                       :string
 #  group                        :string
 #  last_inspection_date         :date
+#  latitude                     :float
 #  local_authority_code         :string
 #  local_authority_name         :string
-#  latitude                     :float
```
Nitemaeric, line 19: This looks unrelated to the PR scope.

spec/models/school_spec.rb
```diff
@@ -14,9 +14,9 @@
 #  gender                       :string
 #  group                        :string
 #  last_inspection_date         :date
+#  latitude                     :float
 #  local_authority_code         :string
 #  local_authority_name         :string
-#  latitude                     :float
```
Nitemaeric, line 19: This looks unrelated to the PR scope.

Nitemaeric (commented):
There is quite a bit of unrelated changes. Let's remove them from the PR - or - put them in their own commit.

Nitemaeric (approved):

spec/models/claim_spec.rb
```diff
@@ -59,13 +59,13 @@
   end
 
   describe "scopes" do
-    describe "#not_internal" do
+    describe "#visible" do
       it "returns the claims that dont have status internal" do
         create(:claim)
         claim1 = create(:claim, :draft)
         claim2 = create(:claim, :submitted)
```
JamieCleare2525, line 66: Can you add claims to this test which aren't `:draft` or `:submitted`. As currently it is returning `Claim.all`

spec/system/claims/support/schools/claims/removing_a_claim_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+  def when_i_click_remove_claim
+    click_on "Remove claim"
+  end
+
+  def when_i_confirm_removal
+    expect(page).to have_content("Are you sure you want to remove this claim?")
+    click_on "Remove claim"
+  end
+
+  def then_the_claim_has_been_deleted
```
JamieCleare2525, line 91: Could you also check for the flash message, to:
1. Show that the correct controller action has occurred.
2. Have confirmation that the deletion has happened that that we aren't just looking at a page without that content.

JamieCleare2525 (approved):

PR #401, opened 2024-03-22
Refactor footer helper to accommodate both services

CatalinVoineag:
## Context

This should make the footer helper a bit easier to scale for both services.

footer_helper.rb

### DIFF
--- a/app/helpers/footer_helper.rb
+++ b/app/helpers/footer_helper.rb
@@ -1,11 +1,30 @@
 module FooterHelper
   def footer_meta_items
+    {
+      claims: claims_footer_meta_items,
+      placements: placements_footer_meta_items,
+    }.fetch current_service
+  end
+
+  private
+
+  def claims_footer_meta_items
     [
-      { text: t(".#{current_service}_guidance"), href: guidance_path },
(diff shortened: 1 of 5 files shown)


CatalinVoineag:
[Refactor footer_helper](https://trello.com/c/ffUATsPW/307-refactor-footerhelper)

app/helpers/footer_helper.rb
```diff
@@ -1,11 +1,31 @@
 module FooterHelper
   def footer_meta_items
+    if current_service == :claims
+      claims_footer_meta_items
+    else
+      placements_footer_meta_items
+    end
```
Nitemaeric, line 7: Let's avoid using an `else` statement. If `current_service` is not either `:claims` or `:placements`, we want to immediately be able to see that.

```suggestion
    {
      claims: claims_footer_meta_items,
      placements: placements_footer_meta_items
    }.fetch current_service
```

Kizr (changes requested):
Please could we add a test for `footer_meta_items`? Given this app serves two purposes it would give us more confidence that we haven't mistakenly broken the part we can't see.

Kizr (approved):
Awesome, thank you!

PR #403, opened 2024-03-22
Move the Claim model into the Claims namespace

Nitemaeric:
## Context

We want to correctly namespace service-specific models. The `Claim` and `MentorTraining` models currently only belong to the "claims" service and should be namespaced accordingly.

The additional benefit of this change is the simplification of the `CODEOWNERS` file.

- Move the `Claim` model into `Claims::Claim`. The table name is unchanged.
- Move the `MentorTraining` model into `Claims::MentorTraining`. The table name is unchanged.
- Use fixtures for `Region` test data.

## Guidance to review

- App should function as is.


.github/CODEOWNERS
```diff
@@ -13,11 +13,9 @@ tmp                                   @DFE-Digital/school-placements  @DFE-Digit
 vendor                                @DFE-Digital/school-placements  @DFE-Digital/track-pay
 
 app/*/claims                          @DFE-Digital/track-pay
-app/policies/claim_policy.rb          @DFE-Digital/track-pay
 spec/*/claims                         @DFE-Digital/track-pay
 config/locales/*/claims               @DFE-Digital/track-pay
 config/locales/*/claims.yml           @DFE-Digital/track-pay
-config/routes/claims.rb               @DFE-Digital/track-pay
```
ollietreend, line 20: I'm assuming `config/routes/claims.rb` is covered by the existing `app/*/claims` rule, and that's why you've removed it. Is that right?

ollietreend (approved):
Looks good to me.

app/models/claims/claim.rb
```diff
@@ -28,8 +28,8 @@
 #  fk_rails_...  (provider_id => providers.id)
 #  fk_rails_...  (school_id => schools.id)
 #
-class Claim < ApplicationRecord
-  belongs_to :school, class_name: "Claims::School"
```
CatalinVoineag, line 32: I'm wondering why do we need a claim to belong to a school and not a claim school?

app/models/claims/claim.rb
```diff
@@ -28,8 +28,8 @@
...
+class Claims::Claim < ApplicationRecord
+  belongs_to :school
   belongs_to :provider
```
CatalinVoineag, line 33: If we have Claims::Provider, this be 
```suggestion
  belongs_to :provider, class_name: "Claims::Provider"
```

app/models/claims/claim.rb
Nitemaeric: This happens by default. Since the `Claims::Claim` model is a part of the `Claims` namespace, its associations lookup the same namespace first. `Claims::School` -> `School` -> "Not Found".

app/models/claims/claim.rb
Nitemaeric: Same as above, the namespace is implicit.

.github/CODEOWNERS
Nitemaeric: Whoops. That was my bad. It won't be included. Adding back...

app/models/claims/claim.rb
CatalinVoineag: Ah OK 👍🏻

app/models/claims/claim.rb
CatalinVoineag: Too much rails magic for me

CatalinVoineag (approved):

.github/CODEOWNERS
Nitemaeric: Added back.

PR #412, opened 2024-03-26
Refactor claim services

CatalinVoineag:
## Context

Previously we had one service that was used by normal users and support users to submit claims.

As the requirements have changed, each of these users will 'submit' the claim in different ways and will trigger different behaviours.

So this PR splits this service into 2

Claim service

## Guidance to review

Create a claim as normal user
Create a claim as support user


CatalinVoineag:
[Refactor Claims::Submit service](https://trello.com/c/PO7hUdHc/308-refactor-claimssubmit-service)

app/controllers/claims/support/schools/claims_controller.rb
```diff
@@ -53,11 +53,7 @@ def update
   end
 
   def draft
-    Claims::Submit.call(
-      claim: @claim,
-      claim_params: { status: :draft },
-      user: current_user,
-    )
+    Claims::Claim::SupportCreateDraft.call(claim: @claim)
```
Nitemaeric, line 56: Theoretically, we don't need to put emphasis on `Support`, right? That's controlled by this service only being called in a support controller.

Nitemaeric (commented):
Looking really good. Clean ✨

app/controllers/claims/support/schools/claims_controller.rb
CatalinVoineag: I emphasised the support part because the service will send emails to all school users. Previsously I called it `CreateDraft` but if we would want to use this service to create a draft in another controller I'm not sure if we want to send emails to all school users.

Basically I thought this should only be used by support. But if you want I can drop the `support bit`.

Nitemaeric: Right now, we don't have an alternative flow for creating drafts, we'd always send out an email notification to the school users that a draft has been created.

That being said, if we _do_ want to deem this `CreateDraft` action as within the scope of Support's action set, we should probably name it `Claims::Support::Claim::CreateDraft`... `<Service>[::UserGroup][::Resource]::<Action>`?

Nitemaeric (approved):

PR #416, opened 2024-03-27
Remove Claims::ClaimDecorator

Nitemaeric:
## Context

We previously moved the `Claims::ClaimDecorator#item_status_tag` logic into a view component. As we do not use this decorator, we should remove it.

- Remove the `Claims::ClaimDecorator`, accompanying spec, and `#decorate` calls on claims.

## Guidance to review

- Specs pass.


CatalinVoineag (approved):
Nice 🗑️ 
Screenshot from 2024-03-27 12-27-40

PR #423, opened 2024-03-27
Add Claims::ClaimsQuery for complex filtering

Nitemaeric:
## Context

We want to provide a maintainable approach to complex queries, such as Search and Filter capabilities.

- Add an `ApplicationQuery` class as a base class for creating query objects.

## Guidance to review

- Review the example `Claims::ClaimsQuery` class and it's drop-in replacement usage.
- Review the `ApplicationQuery`.
- Think of how you would use the existing `ApplicationQuery` and patterns in `Claims::ClaimsQuery` to add your own query objects.

### DIFF
--- a/app/queries/claims/claims_query.rb
+++ b/app/queries/claims/claims_query.rb
@@ -1,0 +1,22 @@
+class Claims::ClaimsQuery < ApplicationQuery
+  def call
+    scope = Claims::Claim.visible
+    scope = school_condition(scope)
+    scope = provider_condition(scope)
+    scope.order_created_at_desc
+  end
+
+  private
+
+  def school_condition(scope)
+    return scope if params[:school_ids].blank?
+
+    scope.where(school_id: params[:school_ids])
+  end
+
+  def provider_condition(scope)
+    return scope if params[:provider_ids].blank?
+
+    scope.where(provider_id: params[:provider_ids])
+  end
+end
--- a/spec/queries/claims/claims_query_spec.rb
+++ b/spec/queries/claims/claims_query_spec.rb
@@ -1,0 +1,41 @@
+require "rails_helper"
+
+describe Claims::ClaimsQuery do
+  subject(:claims_query) { described_class.call(params:) }
+
+  let(:params) { {} }
+
+  describe "#call" do
+    it "returns all visible claims, ordered by created at date descending" do
+      _internal_claim = create(:claim)
+      draft_claim = create(:claim, :draft, created_at: Date.parse("29 March 2024"))
+      submitted_claim = create(:claim, :submitted, created_at: Date.parse("28 March 2024"))
+
+      expect(claims_query).to eq([draft_claim, submitted_claim])
+    end
+
+    context "when given school ids" do
+      let(:school) { create(:claims_school) }
+      let(:params) { { school_ids: [school.id] } }
+
+      it "filters the results by provided school ids" do
+        claim_belonging_to_filtered_school = create(:claim, :submitted, school:)
+        _claim_not_belonging_to_filtered_school = create(:claim, :submitted)
+
+        expect(claims_query).to match_array([claim_belonging_to_filtered_school])
+      end
+    end
+
+    context "when given provider ids" do
+      let(:provider) { create(:claims_provider) }
+      let(:params) { { provider_ids: [provider.id] } }
+
+      it "filters the results by provided provider ids" do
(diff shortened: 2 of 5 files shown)


spec/queries/claims/claims_query_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  let(:params) { {} }
+
+  describe "#call" do
+    it "returns all visible claims, ordered by created at date" do
+      _internal_claim = create(:claim)
+      draft_claim = create(:claim, :draft)
+      submitted_claim = create(:claim, :submitted)
+
+      expect(claims_query).to match_array([draft_claim, submitted_claim])
```
JamieCleare2525, line 14: `match_array` doesn't consider order. To check order you need `eq`

Nitemaeric: I forgot we care about the order here...

app/queries/claims/claims_query.rb
```diff
@@ -0,0 +1,22 @@
+class Claims::ClaimsQuery < ApplicationQuery
+  def call
+    scope = Claims::Claim.visible
```
JamieCleare2525, line 3: Do you ever consider a case where you'd need to query claims which aren't visible?

Nitemaeric: I'd create separate query objects for each of the use-cases.

JamieCleare2525: But if given a scope you wouldn't have to.
Alternatively you could start with `Claims::Claim.all` and have a condition to filter by `status`.
Having 2 separate query objects for the only benefit the main scope seems overkill.

Nitemaeric: I think we should have specific query objects for specific use-cases. In this particular case, we want a query object for the support user's list of all submitted claims. We don't have a use-case for other variants.

Nitemaeric: > Alternatively you could start with Claims::Claim.all and have a condition to filter by status.

If status is a field that we can filter, we'd add a `status_condition` method to allow filtering.

Nitemaeric: From my past experience, re-using query objects for different use-cases doesn't lead to maintainable results.

JamieCleare2525: I disagree with the limitation of the approach, but as this is claims specific I won't hold this back.

ollietreend: > I think we should have specific query objects for specific use-cases. In this particular case, we want a query object for the support user's list of all submitted claims. We don't have a use-case for other variants.

I wonder if this could be made clearer through our choice of naming?

What if the query object was called `Claims::AllVisibleClaimsQuery`?

You would then presumably create a `Claims::AllSubmittedClaimsQuery` and anything else needed for each specific use-case you're describing.

The current name isn't intention-revealing, which I think is what makes it look like it should be a generic query object for finding claims.

JamieCleare2525 (commented):
My remaining thought is that @Kizr did a lot of work to ensure that things like the `ApplicationForm` were able to be tested in isolation to any other Form Object.
I would appreciate tests for the `ApplicationQuery` class and its methods.

app/queries/claims/claims_query.rb
Nitemaeric: So, the query is currently an example put together to show the pattern. The plan is to evolve it into a query object used for the following screen:

CleanShot 2024-03-28 at 15 26 39

A list of all claims a support user should see.

ollietreend: Perhaps this object is a `Claims::Support::ClaimsIndexQuery` then?

JamieCleare2525: Given that we can see a `status` filter in the screen, is it not worth starting with `Claims::Claim.all` and filtering it down by a given `status` (with an `else` of `scope.visible`). Rather than chaining `Claims::Claim.visible.where(status: "draft")`, as the initial `visible` query is unnecessary

Nitemaeric: Once the page design is fully live on the prototype, we'll make sure the filters are complete. This is mainly just a POC.

Nitemaeric:
@JamieCleare2525 Added specs for `ApplicationQuery`.

CatalinVoineag (approved):

Kizr (approved):

JamieCleare2525 (approved):

ollietreend (approved):

PR #424, opened 2024-03-28
Fix claim back links

CatalinVoineag:
## Context

Previously, if a user didn't input mentor hours after adding a mentor
from the check page of the claim form and click the back link. A mentor
without hours would be added to the claim.

This commit will remove any mentors without hours when the user lands on
the check page of a claim. Fixing the issue

Also, this commit will redirect back to the check page when trying to
edit mentor training hours and clicking the back link
if the claim is ready to be checked. Previously the user would need to
cycle through all the mentor hours.

Claim model methods
Claims controller check mentor
Back link redirect for mentor training hours form
Specs

## Guidance to review

Create a claim
On the check page change the mentors
Add a mentor and click continue
Hit the back link to not input any training hours
Hit the back link again
The Mentor you added without mentor training hours should not be on the check page again

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/3e7892e1-9134-4948-b5d1-1255ea0a828a

Problem:

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/fc5a59a5-45cc-4744-9a46-39495f26d8ee

### DIFF
--- a/app/forms/claim/mentor_training_form.rb
+++ b/app/forms/claim/mentor_training_form.rb
@@ -14,7 +14,9 @@ class Claim::MentorTrainingForm < ApplicationForm
   end
 
   def back_path
-    if mentor_trainings.index(mentor_training).zero?
+    if claim.ready_to_be_checked?
+      check_claims_school_claim_path(claim.school, claim)
+    elsif mentor_trainings.index(mentor_training).zero?
       edit_claims_school_claim_mentor_path(claim.school, claim, mentor_training.mentor_id)
     else
       previous_mentor_training = mentor_trainings[mentor_trainings.index(mentor_training) - 1]
--- a/app/models/claims/claim.rb
+++ b/app/models/claims/claim.rb
@@ -63,4 +63,8 @@ class Claims::Claim < ApplicationRecord
   def amount
     Claims::Claim::CalculateAmount.call(claim: self)
   end
+
+  def ready_to_be_checked?
+    mentors.present? && mentor_trainings.without_hours.blank?
+  end
 end
--- a/app/services/claims/claim/remove_empty_mentor_training_hours.rb
+++ b/app/services/claims/claim/remove_empty_mentor_training_hours.rb
@@ -1,0 +1,15 @@
+class Claims::Claim::RemoveEmptyMentorTrainingHours
+  include ServicePattern
+
+  def initialize(claim:)
+    @claim = claim
+  end
+
+  def call
+    claim.mentor_trainings.without_hours.destroy_all
+  end
+
+  private
+
+  attr_reader :claim
+end
--- a/spec/factories/mentor_trainings.rb
+++ b/spec/factories/mentor_trainings.rb
@@ -27,7 +27,7 @@
 FactoryBot.define do
   factory :mentor_training, class: "Claims::MentorTraining" do
     association :claim
-    association :mentor
+    association :mentor, factory: :claims_mentor
     association :provider
   end
 end
--- a/spec/models/claims/claim_spec.rb
+++ b/spec/models/claims/claim_spec.rb
@@ -110,4 +110,26 @@ RSpec.describe Claims::Claim, type: :model do
       end
     end
   end
+
+  describe "#ready_to_be_checked?" do
+    it "returns true if the claim's mentors all have their hours recorded" do
+      claim = create(:claim)
+      create(:mentor_training, hours_completed: 20, claim:)
+
+      expect(claim.ready_to_be_checked?).to eq(true)
+    end
+
+    it "returns false if the claim does have mentors" do
+      claim = build(:claim)
+
+      expect(claim.ready_to_be_checked?).to eq(false)
+    end
+
+    it "returns false if the claim does have mentor training hours" do
+      claim = create(:claim)
+      create(:mentor_training, hours_completed: nil, claim:)
+
+      expect(claim.ready_to_be_checked?).to eq(false)
+    end
+  end
 end
--- a/spec/services/claims/claim/remove_empty_mentor_training_hours_spec.rb
+++ b/spec/services/claims/claim/remove_empty_mentor_training_hours_spec.rb
@@ -1,0 +1,23 @@
+require "rails_helper"
+
+describe Claims::Claim::RemoveEmptyMentorTrainingHours do
+  subject(:service) { described_class.call(claim:) }
+
+  let!(:claim) { create(:claim, reference: nil, status: :internal, school:) }
+  let(:school) { create(:claims_school, urn: "1234") }
+
+  it_behaves_like "a service object" do
+    let(:params) { { claim: } }
+  end
+
+  describe "#call" do
+    it "removes the mentor trainings without hours from a claim" do
+      create(:mentor_training, hours_completed: nil, claim:)
+      create(:mentor_training, hours_completed: nil, claim:)
+      training_with_hours = create(:mentor_training, hours_completed: 20, claim:)
+      expect { service }.to change(Claims::MentorTraining, :count).by(-2)
+
+      expect(claim.mentor_trainings).to eq([training_with_hours])
+    end
+  end
+end
--- a/spec/system/claims/schools/claims/change_claim_on_check_page_spec.rb
+++ b/spec/system/claims/schools/claims/change_claim_on_check_page_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.describe "Change claim on check page", type: :system, service: :claims do
-  let!(:school) { create(:claims_school, mentors: [mentor1, mentor2], region: regions(:inner_london)) }
+  let!(:school) { create(:claims_school, mentors: [mentor1, mentor2, mentor3], region: regions(:inner_london)) }
   let!(:anne) do
     create(
       :claims_user,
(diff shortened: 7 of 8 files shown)


app/models/claims/claim.rb
```diff
@@ -63,4 +63,12 @@ def submitted_on
...
     Claims::Claim::CalculateAmount.call(claim: self)
   end
+
+  def ready_to_be_checked?
+    mentors.present? && mentor_trainings.without_hours.blank?
+  end
+
+  def remove_mentor_trainings_without_hours
+    mentor_trainings.without_hours.destroy_all
+  end
```
Nitemaeric, line 73: This looks harmless enough, but it sets a precedence of data-mutating methods within our models.

Can we move this into a service instead please?

spec/models/claims/claim_spec.rb
```diff
@@ -110,4 +110,41 @@
...
+
+      expect(claim.ready_to_be_checked?).to eq(false)
+    end
+  end
+
+  describe "#remove_mentor_trainings_without_hours" do
+    it "removes all mentor trainings from a claim" do
+      claim = create(:claim)
+      create(:mentor_training, claim:)
+      create(:mentor_training, claim:)
```
Nitemaeric, line 144: This test doesn't show that it keeps trainings _with_ hours.

If I changed

```diff
class Claims::Claim < ApplicationRecord
  def remove_mentor_trainings_without_hours
-   mentor_trainings.without_hours.destroy_all
+   mentor_trainings.destroy_all
  end
end
```

The unit test here would still pass.

spec/models/claims/claim_spec.rb
```diff
@@ -110,4 +110,41 @@
       end
     end
   end
+
+  describe "#ready_to_be_checked?" do
+    it "returns true if the claim is ready to be checked" do
```
Nitemaeric, line 115: This spec description doesn't provide the context of the intended behaviour. What does "ready to be checked" actually mean in terms of behaviour?

```suggestion
    it "returns true if the claim's mentors all have their hours recorded" do
```

spec/models/claims/claim_spec.rb
```diff
@@ -110,4 +110,41 @@
...
+      provider = create(:provider)
+      mentor = create(:claims_mentor)
```
Nitemaeric, line 117: Do these need to be explicit? I think we can rely on the `mentor_training` factory to create these.

We can pass the `claim.provider` through to `create(:mentor_training, provider: claim.provider)`.

app/services/claims/claim/remove_mentor_training_hours.rb
```diff
@@ -0,0 +1,15 @@
+class Claims::Claim::RemoveMentorTrainingHours
```
Nitemaeric, line 1:
```suggestion
class Claims::Claim::RemoveEmptyMentorTrainingHours
```

spec/services/claims/claim/remove_mentor_training_hours_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    context "when mentor trainings have hours completed" do
+      it "does not remove mentor trainings with hours completed from a claim" do
+        training1 = create(:mentor_training, hours_completed: 10, claim:)
+        training2 = create(:mentor_training, hours_completed: 10, claim:)
+        expect { service }.to change(Claims::MentorTraining, :count).by(0)
+
+        expect(claim.mentor_trainings).to eq([training1, training2])
+      end
+    end
+  end
```
Nitemaeric, line 31: Can we add a case where it removes empty mentor trainings but keeps those with hours belonging to a single claim?

These test cases give me confidence that they remove empty-hour records entirely, or not at all, but it doesn't give me confidence in the mixed records case.

spec/system/claims/schools/claims/change_claim_on_check_page_spec.rb
```diff
@@ -63,6 +64,19 @@
...
+  scenario "Anne changes the mentors on claim without inputting hours" do
+    given_i_visit_claim_check_page
+    when_i_click_change_mentors
+    then_i_expect_the_mentors_to_be_checked([mentor1, mentor2])
+    when_i_check_the_mentor(mentor3)
+    when_i_click("Continue")
+    when_i_click("Back")
+    when_i_click("Back")
+    when_i_click("Back")
+    when_i_click("Back")
```
Nitemaeric, line 76: Can we add in some test cases to check what the current page is?

i.e.

```
when_i_click "Back"
then_i_should_see_some_page
when_i_click "Back"
then_i_should_see_some_other_page
```

It helps provide the context of what we're clicking "Back" from.

spec/system/claims/schools/claims/change_claim_on_check_page_spec.rb
```diff
@@ -63,6 +64,19 @@
...
+    then_i_check_my_answers(provider1, [mentor1, mentor2], [20, 12])
```
Nitemaeric, line 77: Can we also be explicit that `mentor3` is _not_ included on the page please?

app/services/claims/claim/remove_mentor_training_hours.rb
Nitemaeric: I think this provides clearer context of what it does.

spec/system/claims/schools/claims/change_claim_on_check_page_spec.rb
```diff
@@ -73,6 +90,15 @@
     when_i_choose_other_amount_and_input_hours(6, with_error: true)
     when_i_click("Continue")
     then_i_check_my_answers(provider1, [mentor1, mentor2], [6, 12])
+    then_i_cant_see_the_mentor(mentor3)
```
Nitemaeric, line 93: I think this was meant to be in the scenario above?

spec/services/claims/claim/remove_empty_mentor_training_hours_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    context "when mentor trainings have hours completed" do
+      it "does not remove mentor trainings with hours completed from a claim" do
+        training1 = create(:mentor_training, hours_completed: 10, claim:)
+        training2 = create(:mentor_training, hours_completed: 10, claim:)
+        expect { service }.to change(Claims::MentorTraining, :count).by(0)
+
+        expect(claim.mentor_trainings).to eq([training1, training2])
+      end
+    end
```
Nitemaeric, line 31: I'm happy to drop this test. The case above shows:

1. It keeps mentor trainings _with_ hours completed.
2. It removes mentor trainings _without_ hours completed.
3. It handles the case where there are a mix of both complete and incomplete hours.

Not a blocker though.

Nitemaeric (approved):
I think the `then_i_cant_see_the_mentor(mentor3)` has been added to the wrong scenario, but once resolved, all good to go.

CatalinVoineag:
[Fix the back link on the claim form](https://trello.com/c/MyS9qbOM/321-fix-the-back-link-on-the-claim-form)

app/forms/claim/mentor_training_form.rb
ollietreend, line 1: Should this file and its siblings be relocated following #403?

```diff
- app/forms/claim/mentor_training_form.rb
+ app/forms/claims/claim/mentor_training_form.rb

- app/forms/claim/mentors_form.rb
+ app/forms/claims/claim/mentors_form.rb

- app/forms/claim/provider_form.rb
+ app/forms/claims/claim/provider_form.rb
```

Right now it's not namespaced under `app/forms/claims/*` and therefore GitHub is asking for a School Placements team member to review it.

spec/factories/mentor_trainings.rb
ollietreend, line 1: Likewise I think it'd be good if we could namespace factories to match the models they represent.

ollietreend (commented):
These comments aren't a blocker for this PR, since resolving them is likely out of scope for this piece of work.

But I think it'd be a good idea to do a sweep and tidy up the namespacing at some point.

app/services/claims/claim/remove_empty_mentor_training_hours.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def call
+    claim.mentor_trainings.without_hours.destroy_all
+  end
+
+  private
+
+  attr_reader :claim
+end
```
ollietreend, line 15: As a relative outsider without context on this side of the codebase, it seems a bit odd to me that MentorTraining records are being created without hours if they're deemed to be 'bad data' that subsequently needs cleaning up. I would have thought a more sensible approach would be to validate the data and avoid creating 'invalid' records in the first place.

ollietreend (approved):
Sorry, I forgot to hit 'approve' with my previous comments.

Without context on the specific problem you're trying to solve, it does seem a bit weird that undesirable records are created as part of 'normal' usage of the form. But 🤷🏻 I imagine there must be a good reason for that particular behaviour.

app/forms/claim/mentor_training_form.rb
CatalinVoineag: This is done as part of George's pr https://github.com/DFE-Digital/itt-mentor-services/pull/435

app/services/claims/claim/remove_empty_mentor_training_hours.rb
CatalinVoineag: I think this is a side effect of a multi-step form. We create a mentor training record when the user adds a mentor to a claim and then he needs to add hours to that mentor. There is an edge case where he can click the back link and back off from adding hours of training, so the mentor training needs to be removed

spec/factories/mentor_trainings.rb
CatalinVoineag: I'll create a separate PR for this

CatalinVoineag:
[Fix back links on the claim form before getting on the check page](https://trello.com/c/G1p0VgYj/456-fix-back-links-on-the-claim-form-before-getting-on-the-check-page)

PR #426, opened 2024-03-28
Update CSV file download name

gms-gs:
## Context

Update the file name to include todays date

Append todays date onto the filename 

## Guidance to review

- Download all claims
- View the filename which is downloaded

### DIFF
--- a/app/controllers/claims/support/claims_controller.rb
+++ b/app/controllers/claims/support/claims_controller.rb
@@ -11,7 +11,7 @@ class Claims::Support::ClaimsController < Claims::Support::ApplicationController
   def download_csv
     csv_data = Claims::GenerateClaimsCsv.call
 
-    send_data csv_data, filename: "claims.csv", type: "text/csv", disposition: "attachment"
+    send_data csv_data, filename: "claims-#{Date.current}.csv", type: "text/csv", disposition: "attachment"
   end
 
   private


app/controllers/claims/support/claims_controller.rb
```diff
@@ -11,7 +11,10 @@ def show; end
   def download_csv
     csv_data = Claims::GenerateClaimsCsv.call
 
-    send_data csv_data, filename: "claims.csv", type: "text/csv", disposition: "attachment"
+    today = Time.zone.today.strftime("%Y%m%d")
```
CatalinVoineag, line 14: I think you can just use `Date.current` here and should give you the same results https://apidock.com/rails/v3.2.1/Date/current/class
```suggestion
    today = Date.current.strftime("%Y%m%d")
```

app/controllers/claims/support/claims_controller.rb
```diff
@@ -11,7 +11,10 @@ def show; end
   def download_csv
     csv_data = Claims::GenerateClaimsCsv.call
 
-    send_data csv_data, filename: "claims.csv", type: "text/csv", disposition: "attachment"
+    today = Date.current.strftime("%Y%m%d")
+    filename = "claims-#{today}.csv"
```
Nitemaeric, line 15: Any reason to not just do `claims-#{Date.current}.csv`?

CatalinVoineag (approved):

PR #437, opened 2024-04-03
Remove internal_draft claims after midnight

CatalinVoineag:
## Context

Internal draft claims are claims that are incomplete, these claims can be created
by a user who doesn't submit/add a claim. Meaning they didn't go through the whole
claim form, they stopped half way.

These claims will just be in our DB if we don't remove them, they cannot be
picked up again.

This commit adds a job to remove the internal draft claims every day after
midnight. We only remove the internal draft claims that have been last updated
`1 day ago`, just in case we remove a claim that's in the process of being submitted.

Remove internal draft claims job and service
Specs

## Guidance to review

Sign in as Colin
Create a few internal draft claims with `updated_at` more than 1 day ago
Go to http://claims.localhost:3000/good_job/cron_entries?locale=en
Run the `Claims::RemoveInternalDraftClaimsJob` manually
The internal draft claims should be removed from your DB

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/c89a64b0-1fc5-4eb4-940d-386b95578952

### DIFF
--- a/config/scheduled_jobs.yml
+++ b/config/scheduled_jobs.yml
@@ -25,3 +25,8 @@ publish_subject_sync:
   cron: "0 0 * * *"
   class: "PublishTeacherTraining::Subject::SyncAllJob"
   description: "Sync all subjects with data imported from Publish Teacher Training"
+
+remove_internal_draft_claims:
+  cron: "10 0 * * *"
+  class: "Claims::RemoveInternalDraftClaimsJob"
+  description: "Remove all internal_draft claims"
--- a/spec/services/claims/claim/remove_internal_drafts_spec.rb
+++ b/spec/services/claims/claim/remove_internal_drafts_spec.rb
@@ -1,0 +1,22 @@
+require "rails_helper"
+
+describe Claims::Claim::RemoveInternalDrafts do
+  subject(:service) { described_class.call }
+
+  let!(:draft_claim) { create(:claim, status: :draft) }
+  let!(:active_internal_draft_claim) { create(:claim, status: :internal_draft) }
+  let!(:submitted_claim) { create(:claim, status: :submitted) }
+
+  it_behaves_like "a service object"
+
+  describe "#call" do
+    it "removes internal draft claims that haven't changed in at least 24 hours" do
+      create(:claim, status: :internal_draft, updated_at: 1.day.ago)
+
+      expect { service }.to change(Claims::Claim, :count).from(4).to(3)
+      expect(Claims::Claim.all).to eq(
+        [draft_claim, active_internal_draft_claim, submitted_claim],
+      )
+    end
+  end
+end
--- a/app/jobs/claims/remove_internal_draft_claims_job.rb
+++ b/app/jobs/claims/remove_internal_draft_claims_job.rb
@@ -1,0 +1,9 @@
+class Claims::RemoveInternalDraftClaimsJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    Audited.audit_class.as_user("System") do
+      Claims::Claim::RemoveInternalDrafts.call
+    end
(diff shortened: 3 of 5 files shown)


CatalinVoineag:
[Add job to remove the internal drafts at midnight](https://trello.com/c/v7mj9e5b/313-add-job-to-remove-the-internal-drafts-at-midnight)

spec/services/claims/claim/remove_internal_drafts_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  let!(:submitted_claim) { create(:claim, status: :submitted) }
+
+  it_behaves_like "a service object"
+
+  describe "#call" do
+    it "removes internal draft claims updated 2 hours ago" do
+      create(:claim, status: :internal_draft, updated_at: 3.hours.ago)
+
+      expect { service }.to change(Claims::Claim, :count).from(3).to(2)
+      expect(Claims::Claim.all).to eq([draft_claim, submitted_claim])
```
Nitemaeric, line 16: Can we add an `internal_draft` claim that was recently updated to this list too please?

Nitemaeric (approved):

config/scheduled_jobs.yml
```diff
@@ -25,3 +25,8 @@ publish_subject_sync:
   cron: "0 0 * * *"
   class: "PublishTeacherTraining::Subject::SyncAllJob"
   description: "Sync all subjects with data imported from Publish Teacher Training"
+
+remove_internal_draft_claims:
+  cron: "10 0 * * *"
+  class: "Claims::RemoveInternalDraftClaimsJob"
+  description: "Remove all internal_draft claims"
```
Nitemaeric, line 32: I wonder how we should test these cron schedules 🤔

CatalinVoineag: Not sure but I could have a look tomorrow after creating a few internal draft claims today to see if they have been removed 😄

app/models/claims/claim.rb
```diff
@@ -47,6 +47,7 @@ class Claims::Claim < ApplicationRecord
 
   scope :visible, -> { where(status: VISIBLE_STATUSES) }
   scope :order_created_at_desc, -> { order(created_at: :desc) }
+  scope :updated_over_2_hours_ago, -> { where("updated_at <= ?", 2.hours.ago) }
```
ollietreend, line 50: Does this really warrant a scope? 🤔 It seems to be very implementation specific.

What happens if you change your definition of a "stale" claim to be > 5 hours ago, rather than 2?

I would suggest either naming the scope something more meaningful and less tightly coupled – for example "stale". Or alternatively just move that query into the service object since that's the only place it's useful.

ollietreend (commented):
A couple of comments from me 🙂

But for what it's worth, I retract my previous statement about it being odd to store half-completed forms in the database 🙈 I think we are going to do the same thing in School Placements as it seems to be the least problematic approach!

app/models/claims/claim.rb
ollietreend: Also there's a slightly more Railsy way to perform this query. Active Record supports [range conditions](https://guides.rubyonrails.org/active_record_querying.html#range-conditions).

Therefore, the query could be written as:

```suggestion
  scope :updated_over_2_hours_ago, -> { where(updated_at: ..2.hours.ago) }
```

which requires no SQL snippets to be written.


The SQL query itself is still the same:

```ruby
Claims::Claim.where(updated_at: ..2.hours.ago).to_sql
```

produces:

```sql
SELECT "claims".* FROM "claims" WHERE "claims"."updated_at" <= '2024-04-04 07:21:53.320273'
```

PR #439, opened 2024-04-03
Rename 'Users' primary menu item in Support console to 'Support Users'

gms-gs:
## Context

Changed the primary nav for a support user

- Changed the order of the nav bar
- Renamed 'Users' primary menu item in Support console to 'Support Users'

## Guidance to review

- Sign in as a support user 

## Link to Trello card

https://trello.com/c/s0WdT1IS/328-rename-users-primary-menu-item-in-support-console-to-support-users

## Screenshots

Screenshot 2024-04-03 at 13 49 46

### DIFF
--- a/config/locales/en/claims/support/support_users.yml
+++ b/config/locales/en/claims/support/support_users.yml
@@ -3,7 +3,7 @@ en:
     support:
       support_users:
         index:
-          page_title: Users
+          page_title: Support users
           add_user: Add user
           heading: Users
         new:
--- a/spec/system/claims/support/support_users/view_support_users_spec.rb
+++ b/spec/system/claims/support/support_users/view_support_users_spec.rb
@@ -6,19 +6,20 @@ RSpec.describe "View support users", type: :system do
 
   scenario "View list of all support users" do
     given_i_sign_in_as(support_user)
-    and_i_visit_the_support_users_page
-    i_see_the_list_of_all_support_users_ordered_by_latest_created_at_first
+    when_i_visit_the_support_users_page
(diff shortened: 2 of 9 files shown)


Nitemaeric:
@gms-gs Could you also update the page's title to "Support Users" too? :)

config/locales/en/claims/support/support_users.yml
```diff
@@ -3,7 +3,7 @@ en:
     support:
       support_users:
         index:
-          page_title: Users
+          page_title: Support users
```
Nitemaeric, line 6: This is also an important piece of presentational content, we should have page title tests :)

Nitemaeric: Can we use the same format as in `spec/system/claims/support_users/add_a_support_user_spec.rb`?

```ruby
and_the_page_title_is("Error: Personal details - Add user - Claim funding for mentor training")
```

spec/system/claims/support/support_users/view_support_users_spec.rb
```diff
@@ -6,19 +6,20 @@
 
   scenario "View list of all support users" do
     given_i_sign_in_as(support_user)
-    and_i_visit_the_support_users_page
-    i_see_the_list_of_all_support_users_ordered_by_latest_created_at_first
+    when_i_visit_the_support_users_page
+    then_i_see_the_list_of_all_support_users_ordered_by_latest_created_at_first_and_page_title
```
Nitemaeric, line 10: We should revert this change to the method name.

Nitemaeric (approved):

PR #443, opened 2024-04-03
Split the claim forms into 2

CatalinVoineag:
## Context

Previously the claim form was used for the support user and normal user journey. These journeys start to diverge a bit, mainly when redirecting the user so to make our code a bit more manageable we decided to split the claim forms into 2.

This commit also fixes some issues with redirecting the support user to the correct page

Created new `::Support` claim forms with their redirect paths

## Guidance to review

Log in as Colin
Create a claim
On the check page change the mentors
Add a mentor and click continue
Hit the back link to not input any training hours
Hit the back link again
The Mentor you added without mentor training hours should not be on the check page again

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/1f9d44d3-7ecb-4ff0-a7aa-7039a736bdf7

### DIFF
--- a/app/controllers/claims/support/schools/claims/mentor_trainings_controller.rb
+++ b/app/controllers/claims/support/schools/claims/mentor_trainings_controller.rb
@@ -7,17 +7,7 @@ class Claims::Support::Schools::Claims::MentorTrainingsController < Claims::Appl
 
   def update
     if mentor_training_form.save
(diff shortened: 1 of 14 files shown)


CatalinVoineag:
[Create Support variants of the Claims-flow form objects](https://trello.com/c/mtuPiX5H/329-create-support-variants-of-the-claims-flow-form-objects)

PR #444, opened 2024-04-04
Use with_content_before_meta_items method instead of custom footer component and styles

gms-gs:
## Context

Due to the latest change to the `govuk_ components` gem , we now don't need to handroll our own component to add in custom content before meta items. The new change exposing a new method `with_content_before_meta_items`, which we can now utilise.

- Update footer to use `with_content_before_meta_items`
- Remove `footer_component`
- Remove dead SCSS

## Guidance to review

- Visit claims or placements, footer should look the same as before

## Link to Trello card

https://trello.com/c/6JsAAwQg/347-replace-existing-custom-footer-component-with-updated-govuk-footer-component-that-has-support-for-content-slots-above-meta-item


CatalinVoineag (approved):

Nitemaeric (approved):
🚀 Nice!

There's definitely _something_ up with the meta item link font sizes, but I'm not sure what.. 🤷 I thought it might have been me that introduced the weird sizes in the custom component.

PR #445, opened 2024-04-05
Allow support user to edit draft claim

gms-gs:
## Context

We want to allow a support user to edit a draft claim

Add change links.

As part of this PR we refactored the back links when making changes to a draft claim. Previously it was in an infinite loop with the back link. Now when a user clicks back, it will take the user back through the flow in the order of when you are creating a claim. This seems sensible and keeps the flow consistent when creating and editing a claim.

## Guidance to review

- Sign in as support user
- Create a claim for a school
- Submit it
- Visit that claims show page and use change links

## Link to Trello card

https://trello.com/c/IDz69VhK/271-allow-support-user-to-edit-draft-claim

### DIFF
--- a/app/views/claims/support/schools/claims/show.html.erb
+++ b/app/views/claims/support/schools/claims/show.html.erb
@@ -22,6 +22,13 @@
           <% summary_list.with_row do |row| %>
             <% row.with_key(text: Claims::Claim.human_attribute_name(:accredited_provider)) %>
             <% row.with_value(text: @claim.provider.name) %>
+            <% if policy(@claim).edit? %>
+              <% row.with_action(text: t("change"),
+                                 href: edit_claims_support_school_claim_path(@school, @claim),
+                                 html_attributes: {
+                                    class: "govuk-link--no-visited-state",
+                                    }) %>
+            <% end %>
           <% end %>
         <% else %>
           <% summary_list.with_row do |row| %>
@@ -38,6 +45,13 @@
                 <li><%= mentor.full_name %></li>
               <% end %>
             </ul>
+          <% end %>
+          <% if policy(@claim).edit? %>
+            <% row.with_action(text: t("change"),
+                               href: edit_claims_support_school_claim_mentor_path(@school, @claim),
+                               html_attributes: {
+                        class: "govuk-link--no-visited-state",
+                      }) %>
           <% end %>
         <% end %>
 
@@ -63,6 +77,20 @@
           <% summary_list.with_row do |row| %>
             <% row.with_key(text: mentor_training.mentor.full_name) %>
             <% row.with_value(text: "#{mentor_training.hours_completed} #{t(".hours")}") %>
+            <% if policy(@claim).edit? %>
+              <% row.with_action(text: t("change"),
+                                 href: edit_claims_support_school_claim_mentor_training_path(
+                                  @school,
+                                  @claim,
+                                  mentor_training,
+                                  params: {
+                                    claims_support_claim_mentor_training_form: { hours_completed: mentor_training.hours_completed },
+                                  },
+                                ),
+                                 html_attributes: {
+                                  class: "govuk-link--no-visited-state",
+                                }) %>
+            <% end %>
           <% end %>
         <% end %>
       <% end %>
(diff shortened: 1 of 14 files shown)


Nitemaeric:
I ran into 2 issues:

1. I think "Add claim" is misleading when making an edit.
2. When you get to the edit (check) page and make a "Change", pressing back goes into an infinite cycle and never leads you back to the "show" page.

https://github.com/DFE-Digital/itt-mentor-services/assets/4231530/d0834879-fe66-4035-9b31-1cb61b835dd4

CatalinVoineag:
> I ran into 2 issues:
> 
> 1. I think "Add claim" is misleading when making an edit.
> 2. When you get to the edit (check) page and make a "Change", pressing back goes into an infinite cycle and never leads you back to the "show" page.
> 
>  CleanShot.2024-04-05.at.11.34.34.mp4

I think the back link logic tailors for the adding a claim flow, not editing a claim. It's tricky because you expect different behaviors from one page 🤷🏻‍♂️ 

I'm wondering if we need a separate page for the editing of a draft claim

CatalinVoineag:
> > > I ran into 2 issues:
> > > 
> > > 1. I think "Add claim" is misleading when making an edit.
> > > 2. When you get to the edit (check) page and make a "Change", pressing back goes into an infinite cycle and never leads you back to the "show" page.
> > > 
> > > CleanShot.2024-04-05.at.11.34.34.mp4
> > 
> > 
> > I think the back link logic tailors for the adding a claim flow, not editing a claim. It's tricky because you expect different behaviors from one page 🤷🏻‍♂️
> > I'm wondering if we need a separate check page for the editing of a draft claim
> 
> Hmmm, yeah either a separate check page for the editing of a draft claim or we add _more_ conditional logic to the back buttons, could use a policy for this maybe?

Maybe, give it a go. I am a bit concerned about the logic of these buttons and having so many different paths to handle

CatalinVoineag (approved):
Nice

app/views/claims/support/schools/claims/show.html.erb
```diff
@@ -22,6 +22,13 @@
           <% summary_list.with_row do |row| %>
             <% row.with_key(text: Claims::Claim.human_attribute_name(:accredited_provider)) %>
             <% row.with_value(text: @claim.provider.name) %>
+            <% if policy(@claim).draft? %>
+              <% row.with_action(text: t("change"),
+                                 href: edit_claims_support_school_claim_path(@school, @claim),
+                                 html_attributes: {
+                                    class: "govuk-link--no-visited-state",
+                                    }) %>
+            <% end %>
```
Nitemaeric, line 31: This looks _strange_ to me. Possibly the policy is being misused.

`policy(_).draft?` means "Can we perform the action "draft" on this resource?". Is that how we're intending to use it here?

I'm worried we're checking if we're attempting to check if this record is a draft.

Nitemaeric (commented):
Can you provide more context in the PR description about the decisions made for this PR please?

app/views/claims/support/schools/claims/show.html.erb
Nitemaeric: Yeah, probably. It looks to me that this is checking whether a user can "edit" the claim, right?

I _think_ we already use the `edit` actions in the create flow too, unfortunately. Maybe we name this new one `change?` to correspond with the link's text?

cc: @CatalinVoineag

spec/policies/claims/claim_policy_spec.rb
```diff
@@ -101,6 +101,32 @@
...
+
+    context "when user has a submitted claim" do
+      it "denies access" do
+        expect(claim_policy).not_to permit(support_user, submitted_claim)
+      end
+    end
+
+    context "when user is not a support user" do
+      it "denies access" do
+        expect(claim_policy).not_to permit(user, submitted_claim)
```
Nitemaeric, line 125: If we're checking a `submitted_claim`, it will _always_ be denied regardless of the type of user.

```suggestion
        expect(claim_policy).not_to permit(user, draft_claim)
```

Nitemaeric (approved):

CatalinVoineag (approved):

PR #447, opened 2024-04-05
Support claims index page filters

CatalinVoineag:
## Context

We need to allow our support users to filter the claims.

Two filters, `School` and `Provider` filters have 2 search fields where
you can search by school name or provider to get the filter you want.
This is done by javascript in a stimulus controller.

The other 2 filters, `Submitted after` and `Submitted before` filters are
date filters and use the `govuk_date_field` form helper. You cannot set
the value of this helper as it's sending 3 params not just 1. So to
persist the value of these filters we need a form object.

`Claims::Support::Claims::FilterForm` is a form object that is meant to
just house the params and all the logic that these filters need.

All the params of the filters are in the url.

`Claims::Support::Claims::FilterForm`
Stimulus controller
Styling changes

## Guidance to review

Sign in as colin
Go to support claims index
Filter claims by all the filters
Use the search too

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/275610e4-c59c-4860-b9c4-d5849f8645d0


CatalinVoineag:
[As a Support User, I want to be able to see the list of all Claims and be able to filter through them](https://trello.com/c/OmdlnZnz/349-as-a-support-user-i-want-to-be-able-to-see-the-list-of-all-claims-and-be-able-to-filter-through-them)

app/controllers/claims/support/claims_controller.rb
```diff
@@ -3,7 +3,10 @@ class Claims::Support::ClaimsController < Claims::Support::ApplicationController
   before_action :authorize_claim
 
   def index
-    @pagy, @claims = pagy(Claims::ClaimsQuery.call(params:))
+    @pagy, @claims = pagy(Claims::ClaimsQuery.call(params: filter_form.query_params))
+    @schools = Claims::School.all
+    @providers = Claims::Provider.private_beta_providers
+    @filter_form = filter_form
```
Nitemaeric, line 9: Could we make this a helper method, similar to our other form objects?

app/views/claims/support/claims/_filter.html.erb
```diff
@@ -0,0 +1,153 @@
...
+            </ul>
+          <% end %>
+
+           <% if @filter_form.provider_ids.present? %>
+             <h3 class="govuk-heading-s govuk-!-margin-bottom-0"><%= t("claims.support.claims.index.accredited_provider") %></h3>
+             <ul class="app-filter-tags">
+               <% @filter_form.provider_ids.each do |provider_id| %>
+                 <li>
+                  <%= govuk_link_to(
+                    Claims::Provider.find_by_id(provider_id)&.name,
```
Nitemaeric, line 47: We should avoid calling database queries in views.

Let's add a `#providers` method to the `filter_form` that loads all providers by `provider_ids` list. This will also prevent N+1 queries.

```ruby
class Claims::ClaimQuery
  def providers
    Claims::Provider.find(provider_ids)
  end
end
```

app/views/claims/support/claims/_filter.html.erb
```diff
@@ -0,0 +1,153 @@
...
+            </div>
+          </div>
+
+          <% if @filter_form.school_ids.present? %>
+            <h3 class="govuk-heading-s govuk-!-margin-bottom-0"><%= t("claims.support.claims.index.school") %></h3>
+            <ul class="app-filter-tags">
+              <% @filter_form.school_ids.each do |school_id| %>
+                <li>
+                 <%= govuk_link_to(
+                   Claims::School.find_by_id(school_id)&.name,
```
Nitemaeric, line 31: This too. Let's avoid N+1 queries.

app/views/claims/support/claims/index.html.erb
```diff
@@ -8,20 +8,40 @@
...
-        clear_search_url: claims_support_claims_path,
-      } %>
+      <div class="govuk-form-group">
+        <%= form_with(
+          url: claims_support_claims_path,
+          scope: :claims_support_claims_filter_form,
+          class: "search-form",
+          method: :get,
+          data: { turbo: false },
+        ) do |f| %>
```
Nitemaeric, line 22: I think it might be easier to manage the filter/search states if this form encompasses both the filters and the search. Otherwise, submitting 1 form would reset the other form, no?

app/javascript/controllers/claims_support_filter_search_controller.js
```diff
@@ -0,0 +1,50 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Connects to data-controller="claims-support-filter-search"
+export default class extends Controller {
+  static targets = [ "schoolInput", "schoolList", "providerInput", "providerList" ]
+
+  connect() {
+    this.schoolInputTarget.addEventListener('input', this.searchSchool.bind(this))
```
Nitemaeric, line 8: There's a _memory leak_ with this code here, it won't be an issue for our app, as our app doesn't use Turbo and performs whole-page reloads, but nevertheless.

To register event listeners with Stimulus that are also cleaned up automatically, use the following approach: https://stimulus.hotwired.dev/reference/actions

In this case, it would be:

```erb
<%= form.govuk_text_field(
              :search_school,
              type: :search,
              data: { claims_support_filter_search_target: "schoolInput", action: "claims-support-filter-search#searchSchool" },
              label: { text: t("claims.support.claims.index.school"), size: "s" },
            ) %>
```

The default "event" type for `input` tags, is `input`.

CleanShot 2024-04-12 at 10 32 43

Nitemaeric:
This is also a pretty complex change. It would help reviewers if this PR was broken down into step-by-step commits.

app/views/claims/support/claims/index.html.erb
CatalinVoineag: I'll try to get both forms working as one

CatalinVoineag: It might be a bit strange for the user, as the filters will be applied when you click search or the search would be triggered when you click apply filters

Nitemaeric: Hmm, I'm not sure. I've previously run into a case where users were surprised that their search query wasn't being removed.

https://github.com/DFE-Digital/itt-mentor-services/assets/4231530/ef454d71-8554-48b8-833c-9604473a9b10

Shall we ask Product?

CatalinVoineag: Probably ask product, the video example you show is happening in the prototype as well. It's just a side effect of having to click buttons rather than live search. I'll ask

CatalinVoineag: Seems like they want search and filtering to be separate https://ukgovernmentdfe.slack.com/archives/C0657JE64HX/p1712929956668779?thread_ts=1712929049.555999&cid=C0657JE64HX

app/views/claims/support/claims/_filter.html.erb
Nitemaeric: We should be using `filter_form.schools` here, right?

app/views/claims/support/claims/_filter.html.erb
Nitemaeric: `filter_form.providers` here too. :)

app/views/claims/support/claims/_filter.html.erb
CatalinVoineag: Stashed the wrong changes

Nitemaeric (approved):

PR #451, opened 2024-04-05
Add Support Claims page

Nitemaeric:
## Things to check


app/queries/claims/claims_query.rb
```diff
@@ -1,13 +1,20 @@
...
     scope = provider_condition(scope)
     scope.order_created_at_desc
   end
 
   private
 
+  def search_condition(scope)
+    return scope if params[:search].blank?
+
+    scope.where(Claims::Claim.arel_table[:reference].matches("%#{params[:search]}%"))
```
CatalinVoineag, line 15: I think we should use PG search here. Don't know if we need postcode and name but something like this https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/models/school.rb#L97

Nitemaeric: Hmm, given the following scenario, would it be confusing to the user?

Search: "1234"

References returned:

- 12345678 (partial)
- 12356789 (_close_ match)

CatalinVoineag: I guess it depends on what the search should do. To me a search form can be used with incomplete information of what you're searching, so if you know only half the reference you can still find what you're looking for. 

But I guess it depends if this is a precise tool or not I guess 🤷🏻

I'm happy to just leave it as it is and if someone wants it we can make it more dynamic.

app/components/claim/card_component.rb
```diff
@@ -0,0 +1,9 @@
+class Claim::CardComponent < ApplicationComponent
```
CatalinVoineag, line 1: Does this need a spec and a preview?

CatalinVoineag (commented):
Overall this looks very good to me, nice 👍🏻

app/components/claim/card_component.rb
Nitemaeric: Ah yes, I'll add that. Nice catch.

Nitemaeric: Added a relatively basic spec for the component, along with a preview.

CatalinVoineag (approved):

JamieCleare2525 (approved):

PR #460, opened 2024-04-09
Refactor the GIAS importer to follow ETL pattern

ollietreend:
## Context

Currently, school data is imported from GIAS in two steps:

1. Download the CSV
2. Import the CSV (filtering out unwanted schools in the process)

We need to perform some 'massaging' of the data. For example, [the importer excludes schools][1] which aren't needed within our app. These are filtered out because either the school has closed, or it isn't in England (which is where the ITT policy currently applies).

We also have an upcoming need to convert the easting/northing fields provided by GIAS into equivalent latitude/longitude values, which will power our location-based search. This is another bit of 'massaging' which needs to happen before the data can be imported.

I've therefore decided to refactor the GIAS import process so it more explicitly follows an [ETL (extract, transform, load)][2] pattern:

1. Download the CSV
2. Transform the CSV (filter out unwanted schools)
3. Import the CSV

This reduces the responsibility of the 'import' step by extracting the filtering logic into a standalone 'transform' step. The importer now simply needs to import every row of the provided CSV. And it'll make it easier to change and add to the transformation logic in future.

### The same end result

This refactor doesn't affect the outcome of the import. The high-level integration test for the entire import still passes as before:

```
bundle exec rspec spec/jobs/gias/sync_all_schools_job_spec.rb
```

### Other changes of note

One change I have made is in testing for edge cases. Previously the importer handled edge cases such as "invalid" CSV files where schools were missing a name or URN. However in practice this never happens: the GIAS CSV file always has a name for each school. The edge-case handling was also not comprehensive – it didn't handle cases where the CSV was missing expected columns or other field values, for example.

Since we haven't seen "invalid" CSV files in practice, it seems like a low-risk change to remove tests for these edge cases. I'd rather keep things simple. And then if we end up receiving corrupt CSV files in the future, we'll be alerted by Sentry with unhandled exceptions when schools fail to save to the database. This will make the problem more obvious instead of quietly skipping those schools.

[1]: https://github.com/DFE-Digital/itt-mentor-services/blob/3a082ae94baf51f28931a6552a2f312768b6c223/app/services/gias/csv_importer.rb#L28
[2]: https://en.wikipedia.org/wiki/Extract,_transform,_load

## Guidance to review

I've tried to create a tidy commit history with useful commit messages. I recommend reviewing this PR commit-by-commit if you want to see the individual steps I took.

## Link to Trello card

This is an enabler for a proper implementation of: https://trello.com/c/3rYfhwuq/228-spike-using-the-easting-northing-values-from-gias-to-geocode-schools


app/services/gias/csv_transformer.rb
```diff
@@ -0,0 +1,55 @@
...
+      CSV.new(input_file, headers: true, return_headers: true).each do |row|
+        if row.header_row? || school_in_scope?(row)
+          output_csv << row
+        end
+      end
+
+      # Rewind the file so it's ready for reading
+      output_file.rewind
+      output_file
+    end
```
Nitemaeric, line 22: I think you can wrap this into a block to remove the for manually rewinding the file.

```suggestion
    def call
      CSV.open(output_file, "w") do |output_csv|
        CSV.foreach(input_file, headers: true, return_headers: true) do |row|
          if row.header_row? || school_in_scope?(row)
            output_csv << row
          end
        end
      end
    end
```

Nitemaeric (approved):

app/services/gias/csv_transformer.rb
ollietreend: For the sake of consistency, the `Gias::CsvDownloader`, `Gias::CsvTransformer` and `Gias::CsvImporter` services all accept and/or return [File] objects.

Internally, `Gias::CsvTransformer` uses a [CSV] object to wrap the `output_file` (a [File] object). But this is an internal implementation detail – externally, the service still returns a regular [File] object.

I don't think [CSV.open] will work in this case because it expects a filename (a String) rather than an actual file handle. And it returns a [CSV] object, where we need to return a [File].

I don't think there's a CSV method that will accept a [File] and a block, and return the file rewound. I appreciate the code looks less elegant than your suggestion, but I think it's what we need to use in this case.

[File]: https://ruby-doc.org/3.2.2/File.html
[CSV]: https://ruby-doc.org/3.2.2/stdlibs/csv/CSV.html
[CSV.open]: https://ruby-doc.org/3.2.2/stdlibs/csv/CSV.html#method-c-open

Kizr (approved):

PR #469, opened 2024-04-11
Revert "Silence deprecation warning by monkey-patching Rack for Capybara"

ollietreend:
Reverts DFE-Digital/itt-mentor-services#76

This monkey-patch was only required temporarily while we waited for teamcapybara/capybara#2706 to be packaged into a release. It was included in [Capybara release 3.40.0](https://github.com/teamcapybara/capybara/compare/3.39.2...3.40.0) which we've been using since #181.

So it should be safe to revert the mokey-patch now.


Nitemaeric (approved):
🐒

PR #470, opened 2024-04-11
Allow multiple hosts for each service

Nitemaeric:
## Context

We need to allow multiple entry domains per service. This is because we want to have access to the apps both behind Azure Front Door and directly to the application.

The app will function as-is without the new `#{service}_HOSTS` env vars, but to allow multiple entry domains, you must use the plural `_HOSTS` env var alongside the `_HOST` env var.

The `#{service}_HOST` env var is used for the base url of email link hrefs.

- Add `CLAIMS_HOSTS` and `PLACEMENTS_HOSTS` comma-separated env vars.

## Guidance to review

**Case 1**

- Don't add any additional env vars to your setup. It should "just work" as-is. The review app should also work without env var changes.

**Case 2**

- Pull down the code.
- Add the following env vars:

  ```dotenv
  CLAIMS_HOSTS=claims.localhost,claims-2.localhost
  PLACEMENTS_HOSTS=placements.localhost,placements-2.localhost
  ```

- You can now visit the apps at `claims.localhost:3000`, `claims-2.localhost:3000`, `placements.localhost:3000`, and `placements-2.localhost:3000`.
  - You may need to update your local "hosts" file to point these domains to your localhost.

### DIFF
--- a/config/routes/claims.rb
+++ b/config/routes/claims.rb
@@ -1,4 +1,6 @@
-scope module: :claims, as: :claims, constraints: { host: ENV["CLAIMS_HOST"] } do
+scope module: :claims, as: :claims, constraints: {
+  host: [ENV["CLAIMS_HOST"], *ENV.fetch("CLAIMS_HOSTS", "").split(",")],
+} do
   root to: "pages#start"
 
   scope module: :pages do
--- a/.env.test
+++ b/.env.test
@@ -6,10 +6,12 @@ PORT=3000
 DFE_SIGN_IN_ISSUER_URL=https://dev-oidc.signin.education.gov.uk
 
 CLAIMS_HOST=claims.localhost
+CLAIMS_HOSTS=claims.localhost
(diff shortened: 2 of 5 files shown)


config/routes/claims.rb
```diff
@@ -1,4 +1,6 @@
-scope module: :claims, as: :claims, constraints: { host: ENV["CLAIMS_HOST"] } do
+scope module: :claims, as: :claims, constraints: {
+  host: [ENV["CLAIMS_HOST"], *ENV.fetch("CLAIMS_HOST", "").split(",")],
```
CatalinVoineag, line 2: Isn't this suppose to be like?
```suggestion
  host: [ENV["CLAIMS_HOST"], *ENV.fetch("CLAIMS_HOSTS", "").split(",")],
```

Nitemaeric: Yes... 🙈

JamieCleare2525 (approved):
Tested locally. Works fine.

PR #471, opened 2024-04-11
Geocode schools using coordinates from GIAS

ollietreend:
## Context

We want to use coordinates from the GIAS dataset to geocode schools rather than using an external geocoding service. For the full context, see [ADR 7](https://github.com/DFE-Digital/itt-mentor-services/blob/geocode-schools-using-gias-coordinates/adr/00007-geocode-schools-using-coordinates-from-gias.md) in this PR.

- Update the `Gias::CsvTransformer` to convert the Easting/Northing coordinates from GIAS into Latitude/Longitude values that can be imported into the app.
- Remove now-redundant functionality for geocoding schools after import.
- Add [PROJ](https://proj.org) as a dependency of this app – including in the Dockerfile, CI workflow, and README documentation.
- An ADR to record this decision.

## Guidance to review

The database seeds file [runs `Gias::SyncAllSchoolsJob` inline](https://github.com/DFE-Digital/itt-mentor-services/blob/b62028e419b53251f2472c06d6b7106bf6d97d23/db/seeds.rb#L17) when no schools are present.

You can test this PR locally by pulling down the branch, installing the `proj` dependency, then re-seeding your database:

```
git checkout geocode-schools-using-gias-coordinates
brew bundle # will install proj using homebrew
bundle exec rails db:reset
```

> [!WARNING]  
> `rails db:reset` will drop and re-create your local database. This is usually safe to run on a dev machine – but don't do it if the database has anything important in it!

Then in a Rails console, run:

```
School.count
# => 23874

School.not_geocoded.count
# => 0
```

Every GIAS school should be in your database and geocoded.

## Link to Trello card

https://trello.com/c/3rYfhwuq/228-use-easting-northing-values-from-gias-to-geocode-schools


ollietreend:
The [tests are failing](https://github.com/DFE-Digital/itt-mentor-services/actions/runs/8650018108/job/23717577486?pr=471#step:7:95).

It looks like I need to build in some tolerance for variation in the latitude/longitude values returned by PROJ. Presumably my laptop and the CI server are running different versions of PROJ, resulting in _very slightly_ different values being returned.

I say _"very slightly"_ because we're talking about a 2 metre difference in the returned values. This is still well within acceptable limits for school locations.

CatalinVoineag (approved):
Looks good, very nice. I wonder if the import will be much faster because of this

JamieCleare2525 (approved):
Tested locally. Appears to be working as intended.

JamieCleare2525 (approved):

PR #472, opened 2024-04-12
Refactor dsi redirect in omniauth

CatalinVoineag:
## Context

We need the dsi redirect to be more flexible, to allow more hosts to user DFE sign in so this commit just gets the base url from the request i.e. http://host:port and we can append `auth/dfe/callback` to it.

Omniauth dsi redirect URL
HostingEnvironment methods cleanup 

## Guidance to review

Use DFE sign-in on your local
It should work as normal


Nitemaeric (approved):
Awesome! 🚀

ollietreend (approved):

PR #473, opened 2024-04-12
Calculate total mentor training hours for a mentor for a given provider

gms-gs:
## Context

We should not allow a mentor to submit over 20 hours of training hours. Currently a mentor can submit 20 hours per claim meaning they can submit over 20 hours in total. 

This change creates a service `Claims::CalculateTotalMentorTrainingHoursForProvider` which returns the total number of mentor training hours for a mentor. This logic will be used in the scenario where a mentor already has submitted 20 hours on a claim, and potentially attempts to submit another 20 hours on another claim for the same provider.

**NOTE: They can submit 20 hours in total per provider**

## Guidance to review

- Run the tests

## Link to Trello card

https://trello.com/c/B637194w/372-set-up-logic-to-only-allow-a-mentor-to-claim-up-to-20-hours-per-provider

### DIFF
--- a/app/services/claims/calculate_total_mentor_training_hours_for_provider.rb
+++ b/app/services/claims/calculate_total_mentor_training_hours_for_provider.rb
@@ -1,0 +1,16 @@
+class Claims::CalculateTotalMentorTrainingHoursForProvider
+  include ServicePattern
+
+  def initialize(mentor:, provider:)
+    @mentor = mentor
+    @provider = provider
+  end
+
+  def call
+    mentor.mentor_trainings.where(provider:).sum(:hours_completed)
+  end
+
+  private
+
+  attr_reader :mentor, :provider
+end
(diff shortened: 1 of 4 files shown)


app/services/claims/calculate_total_mentor_training_hours_for_provider.rb
```diff
@@ -0,0 +1,16 @@
+class Claims::CalculateTotalMentorTrainingHoursForProvider
+  include ServicePattern
+
+  def initialize(mentor:, provider:)
+    @mentor = mentor
+    @provider = provider
+  end
+
+  def call
+    mentor.mentor_trainings.where(provider:).sum(:hours_completed)
```
CatalinVoineag, line 10: I wonder if this service should also calculate hours per academic year. But for now this is enough I think

CatalinVoineag (approved):

app/services/claims/calculate_total_mentor_training_hours_for_provider.rb
Nitemaeric: Yeah, it _should_. But we'll ignore the concept of academic years for now. We'll be addressing academic years after the private beta.

Nitemaeric (approved):

PR #474, opened 2024-04-12
Allow DSI user to actually sign out

CatalinVoineag:
## Context

Previously, the sign out link didn't actually comunicated with DSI to
log out the user. It would still keep them logged in, on their end.

This commit will send a request to DSI to logout the user and redirect
back to our own sign out path

New logout URL in the app

The logout URL has changed in the DSI console, for both placements and claims
Screenshot from 2024-04-15 16-15-34

## Guidance to review

Sign in with dfe sign in on local
Sign out
You should get the sign in form again

## Screenshots

Current:

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/c28adb19-ca9b-4802-983a-19b02007f81d

Previously:

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/2dd9c9f6-6693-494d-9297-62011e7d5a5b

### DIFF
--- a/spec/system/claims/sign_out_as_claims_user_spec.rb
+++ b/spec/system/claims/sign_out_as_claims_user_spec.rb
@@ -3,6 +3,12 @@ require "rails_helper"
 RSpec.describe "Sign out as a Claims User", type: :system, service: :claims do
   let(:colin) { create(:claims_support_user, :colin) }
 
+  before do
+    # rubocop:disable RSpec/AnyInstance
+    allow_any_instance_of(DfESignInUser).to receive(:logout_url).and_return("/auth/dfe/sign-out")
+    # rubocop:enable RSpec/AnyInstance
+  end
+
   scenario "I sign out" do
     given_there_is_an_existing_claims_user_with_a_school_for(colin)
     when_i_visit_the_sign_in_path
@@ -10,6 +16,8 @@ RSpec.describe "Sign out as a Claims User", type: :system, service: :claims do
     i_should_see_a_sign_out_button
     when_i_click_sign_out
     i_expect_to_be_on_sign_in_page
+    when_i_visit_claims_schools_details_path
+    then_i_am_unable_to_access_the_page
   end
 
   private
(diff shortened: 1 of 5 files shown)


spec/system/claims/sign_out_as_claims_user_spec.rb
```diff
@@ -3,13 +3,21 @@
 RSpec.describe "Sign out as a Claims User", type: :system, service: :claims do
   let(:colin) { create(:claims_support_user, :colin) }
 
+  before do
+    # rubocop:disable RSpec/AnyInstance
+    allow_any_instance_of(DfESignInUser).to receive(:logout_url).and_return("/auth/dfe/sign-out")
+    # rubocop:enable RSpec/AnyInstance
```
Nitemaeric, line 9: Are we able to mock this directly onto something like `colin.dfe_sign_in_user`?

CatalinVoineag: Not exactly. We need to mock DFESignInUser and it's a bit of a pain

spec/system/claims/sign_out_as_claims_user_spec.rb
```diff
@@ -3,13 +3,21 @@
 RSpec.describe "Sign out as a Claims User", type: :system, service: :claims do
   let(:colin) { create(:claims_support_user, :colin) }
 
+  before do
+    # rubocop:disable RSpec/AnyInstance
```
JamieCleare2525, line 7: Do these comment need to be removed?

spec/system/claims/sign_out_as_claims_user_spec.rb
CatalinVoineag: Mainly because the `DFESignInUser` changes throughout this spec. If I mock it with an email address the spec will fail because it finds the user in DB and because of this he is logged in. If I mock it without an email, the user is basically logged out. 

The instance of `DFESignInUser` changes state and I don't think mocking the changed state is a good idea. So it feels much cleaner to just mock any instance of `DFESignInUser` to return the desired logout url

Nitemaeric (approved):

PR #497, opened 2024-04-19
Use the FormObject directly when adding a provider to a claim

Nitemaeric:
## Context

The `Claims::Claim::ProviderForm` can be simplified.

- Use `validates` directly. No need for custom validators.
- No need to forward the errors from the underlying model to the form object.

## Guidance to review

- It still works as expected.
- Tests pass.


CatalinVoineag (approved):
Yes! This is what I wanted from the beginning. Well done! ⭐

PR #503, opened 2024-04-22
Update CSV download headers and data

gms-gs:
## Context

We need to update the header format and add submitted_at field

The output file now has a new field and has the headers renamed

## Guidance to review

- Sign in as a support user 
- Visit claims page
- Download claims

## Link to Trello card

https://trello.com/c/wgHiVJ4H/402-update-the-csv-export

## Things to check

### DIFF
--- a/app/services/claims/generate_claims_csv.rb
+++ b/app/services/claims/generate_claims_csv.rb
@@ -3,7 +3,7 @@ require "csv"
 class Claims::GenerateClaimsCsv
   include ServicePattern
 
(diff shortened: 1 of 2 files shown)


Nitemaeric (approved):
There's a bit of potential for this to break _if_ `claim.school == nil` _somehow_.

Also, if we were to add `draft` claims, `submitted_at` would also be `nil`. Maybe `claim.submitted_at&.iso8601` and add a test with drafts? Not a blocker though, we can address later.

PR #508, opened 2024-04-22
Implement claim revisions on claim form

CatalinVoineag:
## Context

In testing the claim form we concluded that we cannot edit the same claim record on the check page when creating a claim.Issues were found when the user would click the back buttons to go back to the check page after completing half their changes. The changes they did were being saved. Even if they backed off from doing them.

To solve this issue and many others, we are implementing a revision system on the claim.

Creating a revision every time a user wants to edit a claim from the check page. If they back off half way through their changes we revert back to the last valid claim revision.

This implementation will increase the audit records and possibly pollute some of them with `system changes` i.e. changes to `previous_revision_id` which the user doesn't care about.

This should fix the issue where a support user edits a draft claim and with every edit, the draft claim is updated for everyone.

The way this works is by using the `internal_draft` status. A generated revision starts with `internal_draft` as its status and is changed to draft when the support user finishes editing the draft claim. The previous revision is then set to `internal_draft`, to not show more than 1 claim per reference.

Claim model
Claim form
Claim support form

## Guidance to review
Normal User: 
- Create a claim as a normal user and try to edit it on the check page
- Click all back buttons
- Add mentors without training hours and click back button until you get back to the check page.

Support User:
Do the same things as the normal user
Edit a draft claim
Edit a draft claim and click back buttons
Your changes to the draft claim should only be saved after you click `Update claim`

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/f7543371-f786-4e7a-8381-dfe1f5577154

### DIFF
--- a/app/controllers/claims/schools/claims/mentor_trainings_controller.rb
+++ b/app/controllers/claims/schools/claims/mentor_trainings_controller.rb
@@ -3,6 +3,17 @@ class Claims::Schools::Claims::MentorTrainingsController < Claims::ApplicationCo
   before_action :authorize_claim
   helper_method :mentor_training_form
 
+  def create_revision
+    revision = Claims::Claim::CreateRevision.call(claim: @claim)
+    mentor_training = revision.mentor_trainings.find_by(mentor_id: params.require(:id))
+
+    redirect_to edit_claims_school_claim_mentor_training_path(
+      @school,
+      revision,
+      mentor_training,
+    )
+  end
+
   def edit; end
 
   def update
--- a/app/services/claims/claim/create_revision.rb
+++ b/app/services/claims/claim/create_revision.rb
@@ -1,0 +1,26 @@
+class Claims::Claim::CreateRevision
+  include ServicePattern
+
+  def initialize(claim:)
+    @claim = claim
+  end
+
+  def call
+    revision_record = deep_dup
+    revision_record.save!
+
+    revision_record
+  end
+
+  private
+
+  attr_reader :claim
+
+  def deep_dup
+    dup_record = claim.dup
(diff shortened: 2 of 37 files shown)


Nitemaeric:
After discussion, we've decided to try a different approach.

app/controllers/claims/schools/claims/mentor_trainings_controller.rb
```diff
@@ -3,6 +3,18 @@ class Claims::Schools::Claims::MentorTrainingsController < Claims::ApplicationCo
   before_action :authorize_claim
   helper_method :mentor_training_form
 
+  def create_revision
```
Nitemaeric, line 6: Should we rename this to `revise` to shorten it?

Nitemaeric: Minor issue for now. Let's leave as is. :)

app/services/claims/claim/create_revision.rb
```diff
@@ -0,0 +1,20 @@
+class Claims::Claim::CreateRevision < Claims::Claim::CreateDraft
```
Nitemaeric, line 1:
```suggestion
class Claims::Claim::CreateRevision
```

CatalinVoineag: Nope 🙃

app/services/claims/claim/update_draft.rb
```diff
@@ -0,0 +1,28 @@
+class Claims::Claim::UpdateDraft < Claims::Claim::CreateDraft
```
Nitemaeric, line 1: This service does the same btw 😛

PR #521, opened 2024-04-24
Add dfe-analytics

CatalinVoineag:
## Context

This will send data on every web request and whenever a table in config/analytics is updated/created. Regardless of domain. We don't have a way to distinguish between them at the moment, though the web requests do contain the domain.

A scheduled job is also going to run every day after midnight to keep in sync the BigQuery data with our own data.

We will send data to dfe-analytics in Production, Sandbox, Staging and QA.

I have included every table that we have apart from the ones that our gems generate. Some contain personal information, they have been included in `config/analytics_pii.yml` to handle the sensitive information. 

Dfe analytics setup

## Guidance to review

Review code.
This will only work when the envs have the secret keys

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/afbe1544-0ed2-452b-8e87-107b15785ee2

### DIFF
--- a/spec/services/concerns/service_pattern_spec.rb
+++ b/spec/services/concerns/service_pattern_spec.rb
@@ -1,4 +1,5 @@
 require "rails_helper"
+require "./app/services/concerns/service_pattern" # needed until DFE-Digital/dfe-analytics#136 is fixed
 
 RSpec.describe ServicePattern do
   describe "#call" do
--- a/.env.example
+++ b/.env.example
@@ -25,3 +25,8 @@ DFE_SIGN_IN_ISSUER_URL=https://dev-oidc.signin.education.gov.uk
 TEACHING_RECORD_BASE_URL=https://preprod.teacher-qualifications-api.education.gov.uk
 TEACHING_RECORD_API_KEY=get-this-from-a-developer
 TEACHING_RECORD_API_MINOR_VERSION=20240101
+
+BIGQUERY_TABLE_NAME=events
+BIGQUERY_PROJECT_ID=rugged-abacus-218110
+BIGQUERY_DATASET=itt_mentor_events_test
+BIGQUERY_API_JSON_KEY=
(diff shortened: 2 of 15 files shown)


spec/services/concerns/service_pattern_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "rails_helper"
+require "./app/services/concerns/service_pattern" # don't get this from dfe-analytics
```
CatalinVoineag, line 2: The spec will include the service pattern from the dfe analytics gem without this 🙃

CatalinVoineag:
[Set up data pipeline (DfE analytics)](https://trello.com/c/u9W4PKnG/17-set-up-data-pipeline-dfe-analytics)

spec/services/concerns/service_pattern_spec.rb
Nitemaeric: 🙃

ollietreend: @CatalinVoineag @Nitemaeric FYI I've reported the issue so the dfe-analytics team are aware: https://github.com/DFE-Digital/dfe-analytics/issues/136

It's on their radar now. But we'll need this workaround for the time being.

Perhaps it'd be useful to add a comment to this line to reference that GitHub issue?

```suggestion
require "./app/services/concerns/service_pattern" # needed until DFE-Digital/dfe-analytics#136 is fixed
```

ollietreend (approved):
This looks good to me. Thanks for setting it up @CatalinVoineag ⭐

But before merging this PR, could you please add a reference to the dfe-analytics issue in the code comment? As per https://github.com/DFE-Digital/itt-mentor-services/pull/521#discussion_r1579296024

Thanks 🙇🏻

Nitemaeric (approved):

PR #522, opened 2024-04-24
Only show mentors with remaining claimable hours in claim flow

Nitemaeric:
## Context

When a mentor has no remaining claimable hours, they cannot be added to a claim. To avoid confusion and limit paths of error, we want to remove them from a School's list of selectable mentors when they have claimed all of their claimable hours across all schools.

- Refactored the `Claims::Claim::MentorsForm` to avoid the need to propagate errors.
  - We need to send in the blanks to trigger validations.
- Show only mentors with remaining hours to claim in the claim flow.
- Changed the `Claims::Mentor::CalculateRemainingMentorTrainingHoursForProvider` service to accept a `claim` argument and find the relevant `mentor_training` record from `mentor` and `provider`.

## Guidance to review

- Add a claim with a mentor and claim all 20 hours for that mentor.
- Try to make a claim for the same provider. The mentor should be missing and an inset text should show.

## Screenshots

CleanShot 2024-04-25 at 12 48 20

### DIFF
--- a/spec/system/claims/schools/claims/change_claim_on_check_page_spec.rb
+++ b/spec/system/claims/schools/claims/change_claim_on_check_page_spec.rb
@@ -55,11 +55,7 @@ RSpec.describe "Change claim on check page", type: :system, service: :claims do
     then_i_see_the_error("Select a mentor")
     when_i_check_the_mentor(mentor2)
     when_i_click("Continue")
-    when_i_click("Continue")
-    then_i_see_the_error("Select the number of hours")
-    when_i_add_training_hours("20 hours")
-    when_i_click("Continue")
-    then_i_check_my_answers(provider1, [mentor2], [20])
+    then_i_check_my_answers(provider1, [mentor2], [12])
     when_i_click("Submit claim")
     then_i_get_a_claim_reference(claim)
   end
--- a/app/components/claims/claim/mentors_form/disclaimer_component.rb
+++ b/app/components/claims/claim/mentors_form/disclaimer_component.rb
@@ -1,0 +1,30 @@
+class Claims::Claim::MentorsForm::DisclaimerComponent < ApplicationComponent
+  attr_reader :mentors_form
+
+  def initialize(mentors_form:, classes: [], html_attributes: {})
+    super(classes:, html_attributes:)
+
+    @mentors_form = mentors_form
+  end
+
+  def call
+    govuk_inset_text do
+      tag.p(sanitize(t(".disclaimer", provider_name:))) +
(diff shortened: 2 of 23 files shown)


Nitemaeric:
[As a User or Support User, I should not be able to see or select a mentor without any remaining claimable hours in the claim flow](https://trello.com/c/Qpy0I7VA/416-as-a-user-or-support-user-i-should-not-be-able-to-see-or-select-a-mentor-without-any-remaining-claimable-hours-in-the-claim-flow)

CatalinVoineag (commented):
The code looks good again 👍🏻 

There is an issue when changing the provider on the check page 😆 
The form doesn't edit the `mentor_training` record to the correct provider. It updates the claim.provider_id but the mentor_training record still has the old provider_id.

So you end up with an issue like this


https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/6f69f477-036f-4f95-a698-f35391b2b682

In my claim revision PR I fixed this issue by doing this https://github.com/DFE-Digital/itt-mentor-services/blob/c9b9315e99118e0eb3d66d07080f0d67a9ea1b34/app/forms/claims/claim/provider_form.rb#L17-L21

Don't know if you have a better way to handle it.

Nitemaeric:
> There is an issue when changing the provider on the check page 😆
The form doesn't edit the mentor_training record to the correct provider. It updates the claim.provider_id but the mentor_training record still has the old provider_id.

Nice callout! I didn't even realise.

spec/system/claims/schools/claims/change_claim_on_check_page_spec.rb
```diff
@@ -55,11 +55,7 @@
     then_i_see_the_error("Select a mentor")
     when_i_check_the_mentor(mentor2)
     when_i_click("Continue")
-    when_i_click("Continue")
-    then_i_see_the_error("Select the number of hours")
```
CatalinVoineag, line 59: Why did we remove these specs where we check for errors?

CatalinVoineag (approved):
Just a small comment but overall looks good to me 👍🏻

spec/system/claims/schools/claims/change_claim_on_check_page_spec.rb
Nitemaeric: The spec reads:

```ruby
then_i_expect_the_mentors_to_be_checked([mentor1, mentor2])
when_i_uncheck_the_mentors([mentor1, mentor2])
when_i_click("Continue")
then_i_see_the_error("Select a mentor")
when_i_check_the_mentor(mentor2)
```

Which means `mentor2`'s hours are already persisted and we move onto the next mentor **without** hours.

PR #527, opened 2024-04-25
Add Proof of Concept for responsive tables

Nitemaeric:
## Context

Wider tables are overflowing on mobile devices.

- Add a `utils` scss directory.
- ~Add a `responsive-table` utility class. Inspired from Bootstrap.~
- Add a `overflow-auto` utility class. Inspired from TailwindCSS. This allows the utility to be atomic and used in any component that requires this utility.

## Guidance to review

- Visit the `/schools/:school_id/claims` path with some claims on mobile. Only the table should be scrollable.

If well received, this may be a contender for an option (or default) of the `govuk_table` helper/component.

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/4231530/68744b69-ed7e-48a9-a425-0acbe4ecdfba

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -5,6 +5,7 @@ $govuk-assets-path: "";
 
 @import "govuk-frontend/dist/govuk/all";
 @import "components/all";
+@import "utils/all";
 @import "accessible-autocomplete/dist/accessible-autocomplete.min";
 @import "search-form";
 @import "filter-form";
--- a/app/assets/stylesheets/utils/_all.scss
+++ b/app/assets/stylesheets/utils/_all.scss
@@ -1,0 +1,1 @@
+@import "overflow";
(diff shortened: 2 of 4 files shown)


ollietreend (approved):
This approach seems reasonable to me.

GOV.UK already does the same thing. For example, try looking at this page with a narrow viewport: [Royal Courts of Justice daily cause list](https://www.gov.uk/government/publications/royal-courts-of-justice-cause-list/royal-courts-of-justice-daily-cause-list#london-administrative-court-cause-list)

That was implemented on GOV.UK a few years ago in alphagov/govuk_publishing_components#1493. And as pointed out in that PR, there's a wider discussion around how the GOV.UK Design System handles tables in small viewports here: alphagov/govuk-design-system-backlog#61.

If accessibility issues are raised with this approach, we may want to refer to the article mentioned here: https://github.com/alphagov/govuk-design-system-backlog/issues/61#issuecomment-735776376

..but I'd suggest this is 'good enough' for now, until we've had it confirmed by an accessibility audit 🙂

PR #535, opened 2024-04-26
Add CSP headers

Nitemaeric:
## Context

We should only allow assets to be loaded in from our own domain.

- Add CSP headers.

## Guidance to review

- As we don't have any code pulling in assets from 3rd parties, our app should work as intended.

## Link to Trello card

https://trello.com/c/gTL5sMZj/418-setup-our-security-headers

## Screenshots

CleanShot 2024-04-26 at 11 52 04

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -4,22 +4,23 @@
 # See the Securing Rails Applications Guide for more information:
 # https://guides.rubyonrails.org/security.html#content-security-policy-header
 
-# Rails.application.configure do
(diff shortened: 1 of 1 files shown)


ollietreend (approved):
Looks good 👍🏻 

```
$ curl -sI "https://manage-school-placements-535.test.teacherservices.cloud/" | grep -i "Content-Security-Policy"
content-security-policy: default-src 'self' https:; font-src 'self' https: data:; img-src 'self' https: data:; object-src 'none'; script-src 'self' https: 'nonce-'; style-src 'self' https:
```

PR #539, opened 2024-04-26
[RFC] Create app_table helper to extend govuk_table helper

Nitemaeric:
## Context

We want to make sure tables are _always_ accessible on mobile and don't cause a page overflow.

- Add a `app_table` helper that wraps a `govuk_table` in a `div` with the `overflow-auto` class.

  This makes the table responsive and viewable across all screen sizes.

## Guidance to review

- Try navigating throughout the services on mobile. Tables should be scrollable when they overflow and not scrollable when they do not overflow.


ollietreend (approved):
Looks good to me. Yay for responsive tables.

PR #541, opened 2024-04-26
Add show page for provider placements

JamieCleare2525:
## Context

- Add show page for Providers to view the details of a placement

- Added routes/controllers/views/etc. to allow Provider users the ability to view a placement.

## Guidance to review

- Sign in as Provider User Patricia
- Navigate to "Placements"
- Select a placement

## Link to Trello card

https://trello.com/c/IYG33bcm/286-build-placement-details-page-for-providers-from-search-results-page

## Screenshots
screencapture-placements-localhost-3000-providers-001898be-80c9-486a-ac42-58a947343419-placements-0bb1a480-366b-4442-bef3-fe4e824fc55c-2024-04-26-15_45_08

### DIFF
--- a/app/components/placement/summary_component.html.erb
+++ b/app/components/placement/summary_component.html.erb
@@ -1,7 +1,6 @@
 <li class="app-search-results__item">
   <h2 class="app-search-result__item-title">
-    <!-- TODO: link to placement details -->
-    <%= govuk_link_to("", no_visited_state: true) do %>
+    <%= govuk_link_to(placements_provider_placement_path(@provider, placement), no_visited_state: true) do %>
       <span><%= school.name %></span>
       <br>
       <span><%= placement.subject_names %></span>
--- a/app/controllers/placements/providers/placements_controller.rb
+++ b/app/controllers/placements/providers/placements_controller.rb
@@ -15,6 +15,11 @@ class Placements::Providers::PlacementsController < ApplicationController
     )
   end
 
+  def show
+    @placement = Placement.find(params[:id]).decorate
+    @school = @placement.school
+  end
+
   private
 
   def set_provider
--- a/app/components/placement/summary_component.rb
+++ b/app/components/placement/summary_component.rb
@@ -1,11 +1,12 @@
 class Placement::SummaryComponent < ApplicationComponent
   with_collection_parameter :placement
-  attr_reader :placement, :school
+  attr_reader :placement, :school, :provider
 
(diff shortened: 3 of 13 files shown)


app/controllers/placements/providers/placements_controller.rb
```diff
@@ -15,6 +15,11 @@ def index
     )
   end
 
+  def show
+    @placement = Placement.find(params[:id]).decorate
+    @school = @placement.school.decorate
```
ollietreend, line 20: I wonder if it'd be handy to use Draper's [decorates_association](https://github.com/drapergem/draper?tab=readme-ov-file#decorating-associated-objects) for this?

```ruby
class SchoolDecorator < OrganisationDecorator
  decorates_association :school
end
```

I think this would mean `@placement.school` is automatically decorated.

app/components/placement/summary_component.html.erb
```diff
@@ -1,7 +1,7 @@
 <li class="app-search-results__item">
   <h2 class="app-search-result__item-title">
     <!-- TODO: link to placement details -->
-    <%= govuk_link_to("", no_visited_state: true) do %>
+    <%= govuk_link_to(placements_provider_placement_path(@provider, placement), no_visited_state: true) do %>
```
ollietreend, line 3: The fact that we have to pass a `@provider` through, just to make this link work... 🤔 it's making me wonder if our routing is suboptimal.

Technically these placements don't "belong to" any specific provider. I wonder if we should be using top-level routes at this point, rather than nesting them inside `/providers/{a-specific-provider}/placements`. The placements themselves don't change based on the context of which provider we're logged in as.

What do you think?

Either way I think it's out of scope for this PR.

ollietreend (approved):
I'm approving because neither of my comments are blockers. This PR is good as is ✅

I'm interested in your thoughts about routing 👀

Nitemaeric (approved):
Nothing really effects Claims here.

Ideally, you guys can move the `Placement` model into `Placements::Placement` to namespace it under the Placements service.

The component can then accompany it at `Placement::Placement::SummaryComponent` and our CODEOWNER won't include @DFE-Digital/track-pay for review :)

app/components/placement/summary_component.html.erb
JamieCleare2525: As we've discussed, moving this routing to a top-level might be premature at this stage, as we don't know the further extend to how this page will be used.
Currently as a provider is the only user who can see this component, it might be worth revisiting in the future.

PR #542, opened 2024-04-26
Ask for DOB when adding a claim mentor

CatalinVoineag:
## Context

We want the user to input a date of birth when adding a mentor.
Otherwise its easy to guess your way to a trn and revealing personal
information about teachers.

The teacher record api can do this validation for us we just need to
update the version that we use and pass the DOB to the teacher record
api.

New Claims Mentor builder
New DOB field
New env variable and edits to specs

## Guidance to review

Add a mentor as a normal and support user
Try to add mentors you already added.

Dob with trn examples
```
# 1986-11-12 1234568
# 1987-09-14 1234566
```

`TEACHING_RECORD_API_MINOR_VERSION` needs to be at least `20240416`

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/e209ae63-95ff-4ab9-b9d6-b998cee9c5cc

### DIFF
--- a/app/controllers/claims/support/schools/mentors_controller.rb
+++ b/app/controllers/claims/support/schools/mentors_controller.rb
@@ -51,7 +51,7 @@ class Claims::Support::Schools::MentorsController < Claims::Support::Application
 
   def mentor_params
     params.require(:claims_mentor_form)
-          .permit(:first_name, :last_name, :trn)
+          .permit(:first_name, :last_name, :trn, :date_of_birth)
           .merge(default_params)
   end
 
--- a/app/forms/claims/mentor_form.rb
+++ b/app/forms/claims/mentor_form.rb
@@ -1,10 +1,18 @@
 class Claims::MentorForm < ApplicationForm
-  FORM_PARAMS = %i[trn].freeze
+  include ActiveModel::Attributes
+  FORM_PARAMS = [:trn, "date_of_birth(1i)", "date_of_birth(2i)", "date_of_birth(3i)"].freeze
 
-  attr_accessor :school, :trn, :first_name, :last_name
+  attribute :school
+  attribute :trn
+  attribute :first_name
+  attribute :last_name
+  attribute "date_of_birth(1i)", :integer
+  attribute "date_of_birth(2i)", :integer
+  attribute "date_of_birth(3i)", :integer
 
   validate :validate_mentor
   validate :validate_membership
+  validates :date_of_birth, presence: true
 
   def persist
     ActiveRecord::Base.transaction do
@@ -19,13 +27,28 @@ class Claims::MentorForm < ApplicationForm
   end
 
   def mentor
-    @mentor ||= MentorBuilder.call(trn:, first_name:, last_name:).becomes(Claims::Mentor)
+    @mentor ||= Claims::MentorBuilder.call(
+      trn:,
+      first_name:,
+      last_name:,
+      date_of_birth:,
+    )
   end
 
   def validate_mentor
     if mentor.errors.any?
-      mentor.errors.each { |err| errors.add(:trn, err.message) }
+      mentor.errors.each { |err| errors.add(err.attribute, err.message) }
     end
+  end
+
+  def date_of_birth
+    Date.new(
+      attributes["date_of_birth(1i)"], # year
+      attributes["date_of_birth(2i)"], # month
+      attributes["date_of_birth(3i)"], # day
+    )
+  rescue Date::Error, TypeError
+    nil
   end
 
   private
--- a/app/services/claims/mentor_builder.rb
+++ b/app/services/claims/mentor_builder.rb
@@ -1,0 +1,30 @@
+class Claims::MentorBuilder
+  include ServicePattern
+
+  def initialize(trn:, first_name: nil, last_name: nil, date_of_birth: nil)
+    @trn = trn
+    @first_name = first_name
+    @last_name = last_name
+    @date_of_birth = date_of_birth
+  end
+
+  def call
+    mentor = Mentor.find_or_initialize_by(trn:).becomes(Claims::Mentor)
+    return mentor if mentor.invalid_attributes?(:trn) # Return mentor if basic validation fails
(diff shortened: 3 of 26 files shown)


app/forms/claims/mentor_form.rb
```diff
@@ -1,10 +1,18 @@
...
+  FORM_PARAMS = [:trn, "date_of_birth(1i)", "date_of_birth(2i)", "date_of_birth(3i)"].freeze
 
-  attr_accessor :school, :trn, :first_name, :last_name
+  attribute :school
+  attribute :trn
+  attribute :first_name
+  attribute :last_name
+  attribute "date_of_birth(1i)"
+  attribute "date_of_birth(2i)"
+  attribute "date_of_birth(3i)"
```
Nitemaeric, line 11: Could these be `attr_accessor`s instead? `ActiveModel::Attributes` adds type-safety/casting.

app/forms/claims/mentor_form.rb
```diff
@@ -19,15 +27,30 @@ def as_form_params
   end
 
   def mentor
-    @mentor ||= MentorBuilder.call(trn:, first_name:, last_name:).becomes(Claims::Mentor)
+    @mentor ||= Claims::MentorBuilder.call(
+      trn:,
+      first_name:,
+      last_name:,
+      date_of_birth:,
+    ).becomes(Claims::Mentor)
```
Nitemaeric, line 35: A `Claims::MentorBuilder` should be able to handle the type-casting within the service, as it's already namespaced.

app/forms/claims/mentor_form.rb
```diff
@@ -19,15 +27,30 @@ def as_form_params
...
 
+  def date_of_birth
+    Date.new(
+      attributes["date_of_birth(1i)"].to_i, # year
+      attributes["date_of_birth(2i)"].to_i, # month
+      attributes["date_of_birth(3i)"].to_i, # day
+    )
+  rescue Date::Error
+    nil
+  end
```
Nitemaeric, line 52: We use this enough to warrant a helper method.

```ruby
date_attribute :date_of_birth

def self.date_attribute(attr_name)
  define_method(attr_name) do
    Date.new(
      attributes["#{attr_name}(1i)"].to_i, # year
      attributes["#{attr_name}(2i)"].to_i, # month
      attributes["#{attr_name}(3i)"].to_i, # day
    )
  rescue Date::Error
    nil
  end
end
```

\* I see where we use `attributes` now 🤔. Maybe we do `#send`?

app/models/claims/mentor.rb
```diff
@@ -2,12 +2,13 @@
...
 # Table name: mentors
 #
-#  id         :uuid             not null, primary key
-#  first_name :string           not null
-#  last_name  :string           not null
-#  trn        :string           not null
-#  created_at :datetime         not null
-#  updated_at :datetime         not null
+#  id            :uuid             not null, primary key
+#  date_of_birth :date
```
Nitemaeric, line 6: We mustn't store `date_of_birth` in our DB.

app/forms/claims/mentor_form.rb
Nitemaeric: \* Not a blocker

CatalinVoineag: This is quite different now. It will return nil if the values are not integers. This form cannot send dates like `12-12-0000` but the filter form that has a similar method to this can

app/views/claims/schools/mentors/check.html.erb
```diff
@@ -12,6 +12,7 @@
         <%= f.hidden_field :first_name, value: mentor_form.mentor.first_name %>
         <%= f.hidden_field :last_name, value: mentor_form.mentor.last_name %>
         <%= f.hidden_field :trn, value: mentor_form.trn %>
+        <%= f.govuk_date_field :date_of_birth, hidden: true, date_of_birth: true %>
```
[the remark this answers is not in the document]
CatalinVoineag: Yes, the form needs to remember the date of birth when you click `back` or `Change your search`

Screenshot from 2024-04-29 14-05-46

app/controllers/claims/support/schools/mentors_controller.rb
```diff
@@ -19,8 +19,13 @@ def create
   end
 
   def check
+    exceptions = [
+      TeachingRecord::RestClient::TeacherNotFoundError,
+      Claims::MentorBuilder::TeacherNotFoundError,
+    ]
+
```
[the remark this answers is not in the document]
CatalinVoineag: Yeah but I'm not sure what to inherit as it's just an empty class that inherits from Standard error. Just adds dependency I think

https://github.com/DFE-Digital/itt-mentor-services/blob/7d20a51b0c79f6583c8cf44a4cb65b939b7adef4/app/services/teaching_record/rest_client.rb#L12

app/services/claims/mentor_builder.rb
```diff
@@ -0,0 +1,36 @@
...
+    mentor
+  end
+
+  private
+
+  attr_accessor :trn, :first_name, :last_name, :date_of_birth
+
+  def dob_matches?
+    date_of_birth == teacher["dateOfBirth"].to_date
+  end
```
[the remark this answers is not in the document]
CatalinVoineag: Added a bit more clarity to this now https://github.com/DFE-Digital/itt-mentor-services/blob/7aa43c5cb2f5aa166a7038f93f9f6113ce5673a2/app/services/claims/mentor_builder.rb#L17-L21

Nitemaeric (approved):
We should call out in the PR description that the `TEACHING_RECORD_API_MINOR_VERSION` env var needs to be at least `20240416`. If not, it will give false positives.

The "Not found" message needs to be updated to "Check that you typed in the teacher reference number (TRN) and date of birth correctly.", but code looks good to go. :D

app/services/claims/mentor_builder.rb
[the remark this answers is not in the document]
CatalinVoineag: This has been refactored and addressed

PR #550, opened 2024-04-29
Add Provider to Placement

Kizr:
## Context

School users should be able to select a provider for their placements then they know which provider will be assigned. This is an optional choice.

- [x] Adds an optional `method` argument for the auto select form
- [x] Adds a new `name` delegation to `provider` for a Placement
- [x] Updates the Placement show page to display the provider and change link
- [x] Adds an edit page for editing providers on placements

## Guidance to review

- Create a new placement
- Click on the placement on the index page
- Click on Change in the provider row
- Select a provider
- Click on continue

## Link to Trello card

[Add Provider to Placement](https://trello.com/c/BSK7vNUk/269-provider-placements)

## Screenshots

image
image
image

### DIFF
--- a/app/controllers/placements/schools/placements_controller.rb
+++ b/app/controllers/placements/schools/placements_controller.rb
@@ -1,6 +1,6 @@
 class Placements::Schools::PlacementsController < ApplicationController
   before_action :set_school
-  before_action :set_placement, only: %i[show remove destroy]
+  before_action :set_placement, only: %i[show edit_provider update remove destroy]
   before_action :set_decorated_placement, only: %i[show remove]
 
   def index
@@ -11,6 +11,20 @@ class Placements::Schools::PlacementsController < ApplicationController
   def show; end
 
   def remove; end
+
+  def edit_provider
+    @providers = Provider.all
+  end
+
+  def update
+    @placement.provider = Provider.find(provider_params[:provider_id]) if provider_params[:provider_id].present?
+
+    if @placement.save!
+      redirect_to placements_school_placement_path(@school, @placement), flash: { success: t(".success") }
+    else
+      render :edit_provider
+    end
+  end
 
   def destroy
     @placement.destroy!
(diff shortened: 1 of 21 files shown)


app/views/placements/schools/placements/show.html.erb
```diff
@@ -50,11 +50,16 @@
             <%= render Placement::StatusTagComponent.new(@placement.status) %>
           <% end %>
         <% end %>
+        <% summary_list.with_row do |row| %>
+          <% row.with_key(text: t(".attributes.placements.provider")) %>
+          <% row.with_value(text: @placement.provider_name.presence || t("placements.schools.placements.not_known_yet")) %>
```
JamieCleare2525, line 55: Might be worth moving this logic to the decorator.

app/controllers/placements/schools/placements_controller.rb
```diff
@@ -12,6 +12,17 @@ def show; end
 
   def remove; end
 
+  def edit_provider
+    @providers = Provider.all
+  end
+
+  def update
+    @placement.provider = provider_params[:provider_name].present? ? Provider.find(provider_params[:provider_id]) : nil
```
JamieCleare2525, line 20:
```suggestion
    @placement.provider = provider_params[:provider_name].present? && provider_params[:provider_id].present? ? Provider.find(provider_params[:provider_id]) : nil
```

Kizr: The ID is funky here, the name is correct as it doesn't come through when deleting text. Looks wrong but it's correct.

I expect this to change/improve once designs come through

Kizr: I realise I could still check the presence here, it's always present with the component but that doesn't stop me testing, I have made a similar change in the second PR I have up as I moved this logic.

JamieCleare2525 (approved):

Nitemaeric (approved):

PR #552, opened 2024-04-29
Remove placement windows

ollietreend:
## Context

We added the concept of placement windows quite early on in the build, off the back of the design prototype that was in use at the time. However it turns out placement windows aren't as straightforward or well understood as initially thought. They have subsequently been hidden in the prototype and removed from flows such as "Add a placement".

We've now agreed that they're not required for 'day 1' of the service.

Placement windows will likely be re-introduced once the team has a better understanding of what they are, how they work, and where they'll be used within the service. There's no guarantee that our initial implementation will be compatible with that, though, so it seems sensible to just tidy it all up and create a blank slate for whenever they're re-introduced.

We can, of course, always refer back to this PR if ever we need to see how it was implemented originally.

Remove placement windows, start dates and end dates.

Also note that my IDE reformatted the dfe-analytics YAML configuration files. I just went with it rather than fighting VS Code 🙂 but I've kept that in a separate commit, so it shouldn't cloud up the other commits.

## Guidance to review

No specific guidance.

## Link to Trello card

https://trello.com/c/nKKn80UP/272-tidy-up-window-from-placements-for-day-1-this-is-not-needed

## Screenshots

| Before | After |
| --- | --- |
| Placements index page with a column for placement window | Placements index page without a placement window column |


db/migrate/20240429133246_remove_start_date_and_end_date_from_placements.rb
```diff
@@ -0,0 +1,7 @@
+class RemoveStartDateAndEndDateFromPlacements < ActiveRecord::Migration[7.1]
+  def change
+    change_table :placements, bulk: true do |t|
+      t.remove :start_date, :end_date, type: :date
+    end
```
Nitemaeric, line 5: I think you'll need to update the `analytics.yml` files and remove these columns from there.

Nitemaeric (commented):
Loving the formatting change 😄

db/migrate/20240429133246_remove_start_date_and_end_date_from_placements.rb
Nitemaeric: Oh, you did preemptively in the commit before this change.

Nitemaeric (approved):

PR #562, opened 2024-04-30
While a Mentor is associated with an "active" Claim, it cannot be "removed"

gms-gs:
## Context

We dont want to allow a user to remove a mentor who is part of an active claim

Add a blocking flow if the user is part of an active claim

## Guidance to review

- Add a mentor
- Create a claim with that mentor
- Try and remove that mentor

## Link to Trello card

https://trello.com/c/s8hqOMUv/420-while-a-mentor-is-associated-with-an-active-claim-it-cannot-be-removed

### DIFF
--- a/app/policies/claims/mentor_policy.rb
+++ b/app/policies/claims/mentor_policy.rb
@@ -1,5 +1,9 @@
 class Claims::MentorPolicy < Claims::ApplicationPolicy
   def destroy?
+    !Claims::Claim.active.joins(:mentor_trainings).where(mentor_trainings: { mentor_id: record }).exists?
+  end
+
+  def remove?
     true
   end
 end
--- a/app/views/claims/schools/mentors/_cannot_remove_mentor.erb
+++ b/app/views/claims/schools/mentors/_cannot_remove_mentor.erb
@@ -1,0 +1,13 @@
+<% content_for :page_title, t(".page_title") %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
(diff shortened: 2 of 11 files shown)


app/services/claims/mentor/can_remove_from_claim.rb
```diff
@@ -0,0 +1,19 @@
...
+    @mentor = mentor
+  end
+
+  def call
+    mentor.mentor_trainings.each do |mentor_training|
+      claim_status = Claims::Claim.find(mentor_training.claim_id).status
+      return false if ACTIVE_CLAIM_STATUSES.include?(claim_status)
+    end
+    true
+  end
```
Nitemaeric, line 16: This can potentially hit the database quite a few times. I'd advocate for minimal database queries as possible.

Maybe something like:

```ruby
Claims::Claim.active.joins(:mentor_trainings).where(mentor_trainings: { mentor_id: mentor }).exists?
```

Nitemaeric (commented):
Could this be moved into the mentor policy?

```ruby
class Claims::MentorPolicy
  def destroy?
    # Only if not related to active claims
  end

  # We need to override this, as the default is `remove? == destroy?`
  def remove?
    true
  end
end
```

In the `remove` view, we can then do:

```erb
<% if policy(mentor).destroy? %>
  # Are you sure?
<% else %>
  # You can't do this
<% end %>
```

This also prevents tech-savvy users from directly calling a `DELETE` request via a http client. Security :D

What do you think?

Nitemaeric (approved):

PR #563, opened 2024-04-30
Reject claims with invalid mentor training hours

CatalinVoineag:
## Context

We have a race condition where a user can create a claim for a mentor before someone else trying to do the same thing.

So you can end up with more mentor training hours per provider than the maximum 20 hours allowed, per provider.

This PR adds a validation step in the submit and draft controller actions and redirect the user to a reject page when this happens

This PR adds a validation step in the submit and draft controller actions and redirect the user to a reject page when this happens.

Validation method on the claim model. This should not be used as a `validates` method, we want to have control over when this is called.

New reject route and page

## Guidance to review

Try replicating the example below 👇🏻 

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/ba188aee-e8d6-4074-a0f4-dcc509a2bc74

### DIFF
--- a/app/controllers/claims/support/schools/claims_controller.rb
+++ b/app/controllers/claims/support/schools/claims_controller.rb
@@ -1,7 +1,7 @@
 class Claims::Support::Schools::ClaimsController < Claims::Support::ApplicationController
   include Claims::BelongsToSchool
 
-  before_action :set_claim, only: %i[check draft show edit update remove destroy]
+  before_action :set_claim, only: %i[check draft show edit update remove destroy rejected]
   before_action :authorize_claim
   helper_method :claim_provider_form
 
@@ -54,11 +54,15 @@ class Claims::Support::Schools::ClaimsController < Claims::Support::ApplicationC
   end
 
   def draft
+    return redirect_to rejected_claims_support_school_claim_path unless @claim.valid_mentor_training_hours?
+
     success_message = @claim.draft? ? t(".update_success") : t(".add_success")
     Claims::Claim::CreateDraft.call(claim: @claim)
 
     redirect_to claims_support_school_claims_path(@school), flash: { success: success_message }
   end
+
+  def rejected; end
 
   private
 
--- a/app/models/claims/claim.rb
+++ b/app/models/claims/claim.rb
@@ -29,6 +29,7 @@
 #  fk_rails_...  (school_id => schools.id)
 #
 class Claims::Claim < ApplicationRecord
+  MAXIMUM_CLAIMABLE_HOURS = 20
   audited
   has_associated_audits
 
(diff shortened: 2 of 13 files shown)


app/controllers/claims/support/schools/claims_controller.rb
```diff
@@ -54,12 +54,16 @@ def update
...
   def draft
+    return redirect_to reject_claims_support_school_claim_path if @claim.invalid_mentor_training_hours?
+
     success_message = @claim.draft? ? t(".update_success") : t(".add_success")
     Claims::Claim::CreateDraft.call(claim: @claim)
 
     redirect_to claims_support_school_claims_path(@school), flash: { success: success_message }
   end
 
+  def reject; end
```
Nitemaeric, line 65: This reads like a user-triggered action to me. Like, "I want to reject _thing_".

Can we rename this to `rejected` please? This way it reads like feedback to the user.

app/views/claims/schools/claims/reject.html.erb
```diff
@@ -0,0 +1,23 @@
...
+      </label>
+
+      <p class="govuk-body"><%= t(".reject_reason") %></p>
+
+      <p class="govuk-body">
+        <%= govuk_link_to t(".view_claims"), claims_school_claims_path(@school), no_visited_state: true %>
+      </p>
+
+      <p class="govuk-body">
+        <%= t(".guidance_html", support_email: mail_to(t("#{current_service}.support_email"), class: "govuk-link")) %>
```
Nitemaeric, line 19: This will never be rendered within the placements service, so we can do:

```suggestion
        <%= t(".guidance_html", support_email: mail_to(t("claims.support_email"), class: "govuk-link")) %>
```

app/models/claims/claim.rb
```diff
@@ -56,6 +56,22 @@ class Claims::Claim < ApplicationRecord
...
+        claim: self,
+      )
+
+      if mentor_training.hours_completed > remaining_training_hours
+        return true
+      end
+    end
+
+    false
+  end
```
Nitemaeric, line 73: Could this potentially be:

```ruby
class Claims::Claim
  validates_associated :mentor_trainings
end
```

And then we do a validation before the submit call?

```ruby
class Claims::Schools::ClaimsController
  def submit
    if @claim.valid?
      # perform Submit
    else
      redirect_to rejected...
    end
  end
end
```

Nitemaeric: Hmm, I think I can see a potential issue. When creating an `internal_draft`, it _might_ be invalid.

My worry is without a validation, we may be letting invalid data through.

Nitemaeric: I now recall that we don't actually validate `mentor_trainings` as records that effectively either. It's validated within the form object, but we need to validate the claim as a whole too.

app/models/claims/claim.rb
```diff
@@ -29,6 +29,7 @@
 #  fk_rails_...  (school_id => schools.id)
 #
 class Claims::Claim < ApplicationRecord
+  MAXIMUM_CLAIMABLE_HOURS = 20
```
Nitemaeric, line 32: This constant is repeated in 3 places now. We'll need to consolidate to a single place soon.

Not for this PR though.

Nitemaeric (approved):

app/models/claims/claim.rb
CatalinVoineag: Addressed this, kept it as a method but changed the query

PR #569, opened 2024-04-30
Skip placement mentors if none exist

Kizr:
## Context

If a school had no mentors we'd end up with an empty page, this change makes sure that the page is never shown if no mentors have been added.

Removes the add mentors page if none exist.

## Guidance to review

Remove all mentors and create a placement, make sure it works as expected.

## Link to Trello card

["Add a placement" – skip the "mentors" step if the org has no mentors](https://trello.com/c/r4KlVY8Z/303-add-a-placement-skip-the-mentors-step-if-the-org-has-no-mentors)

### DIFF
--- a/app/controllers/placements/schools/placements/build_controller.rb
+++ b/app/controllers/placements/schools/placements/build_controller.rb
@@ -3,6 +3,7 @@ class Placements::Schools::Placements::BuildController < ApplicationController
 
   def new
     reset_session
+    setup_skipped_steps
     @placement = build_placement
     school.primary_or_secondary_only? ? handle_primary_or_secondary : handle_other
   end
@@ -23,7 +24,6 @@ class Placements::Schools::Placements::BuildController < ApplicationController
   end
 
   def check_your_answers
-    session[:add_a_placement][:enable_phase_navigation] = true
     @placement = initialize_placement
     @phase = session.dig(:add_a_placement, "phase")
     @selected_mentor_text = if @placement.mentors.empty?
@@ -137,7 +137,13 @@ class Placements::Schools::Placements::BuildController < ApplicationController
   end
 
   def next_step(step)
-    "#{STEPS[STEPS.index(step.to_sym) + 1]}_placements_school_placement_build_index_path"
+    steps = if skipped_steps.present?
+              STEPS - skipped_steps.map(&:to_sym)
+            else
+              STEPS
+            end
+
+    "#{steps[steps.index(step.to_sym) + 1]}_placements_school_placement_build_index_path"
   end
 
   def school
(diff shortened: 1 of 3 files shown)


app/views/placements/schools/placements/build/check_your_answers.html.erb
```diff
@@ -24,20 +24,22 @@
             <% row.with_key { t(".phase") } %>
             <% row.with_value(text: @phase) %>
             <% unless @school.primary_or_secondary_only? %>
-                <% row.with_action(text: t(".change"), href: add_phase_placements_school_placement_build_index_path(@school, :add_phase, enable_quick_navigation: true), visually_hidden_text: t(".add_phase")) %>
+              <% row.with_action(text: t(".change"), href: add_phase_placements_school_placement_build_index_path(@school, :add_phase, enable_quick_navigation: true), visually_hidden_text: t(".add_phase")) %>
```
ollietreend, line 27: Is `enable_quick_navigation` a hangover from the Ghost of PRs Past? 👀

```suggestion
              <% row.with_action(text: t(".change"), href: add_phase_placements_school_placement_build_index_path(@school, :add_phase), visually_hidden_text: t(".add_phase")) %>
```

app/controllers/placements/schools/placements/build_controller.rb
```diff
@@ -153,12 +159,22 @@ def build_placement
...
 
+  def setup_skipped_steps
+    return if school.mentors.exists?
+
+    session[:add_a_placement][:skipped_steps] = %w[add_mentors]
+  end
+
+  def skipped_steps
+    session.dig(:add_a_placement, "skipped_steps") || []
+  end
```
ollietreend, line 170: There's absolutely nothing wrong with this approach, so no need to change it. But I do wonder whether we really need to store this in the session, or if it could be calculated dynamically.

Perhaps if/when we refactor this in the future, we could keep the 'conditional' bit closer to the step definition itself. For example (pseudo-code!):

```ruby
STEPS = [
  { name: :choose_phase, unless: -> { school.primary_or_secondary_only? } },
  { name: :choose_subject },
  { name: :choose_mentors, unless: -> { school.mentors.none? } },
  { name: :check_your_answers },
]
```

As we've said before, I think there's probably an opportunity to refine this a bit once the dust has settled a bit – or once we have more wizard-style flows to implement.

Kizr: This is kind of what I wanted to achieve but didn't know that you could do it this way, this is really interesting! Thanks for raising.

ollietreend: Added to the refactor card: https://trello.com/c/HhyaTg8c/98-tidy-up-the-add-a-placement-multi-step-form-implementation

ollietreend (approved):
Nice work 👍🏻

PR #570, opened 2024-04-30
Scope MentorMembershipPolicy destroy? to school

gms-gs:
We need to scope to the school. So to active claims of a school can't have that mentor. Not all active claims.

### DIFF
--- a/app/controllers/claims/schools/mentors_controller.rb
+++ b/app/controllers/claims/schools/mentors_controller.rb
@@ -3,7 +3,10 @@ class Claims::Schools::MentorsController < Claims::ApplicationController
 
   before_action :has_school_accepted_grant_conditions?
   before_action :set_mentor, only: %i[show remove destroy]
+  before_action :set_mentor_membership, only: %i[remove destroy]
+
   before_action :authorize_mentor
+  before_action :authorize_mentor_membership
 
   helper_method :mentor_form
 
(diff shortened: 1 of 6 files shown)


CatalinVoineag (approved):
Nice, works as expected

app/controllers/claims/schools/mentors_controller.rb
```diff
@@ -42,10 +45,18 @@ def set_mentor
     @mentor = @school.mentors.find(params.require(:id))
   end
 
+  def set_mentor_membership
+    @mentor_membership = @mentor.mentor_memberships.find_by!(school: @school)
+  end
```
Nitemaeric, line 50: This logic is duplicated on line 36. We should remove line 36 and use this reference instead.

app/controllers/claims/schools/mentors_controller.rb
```diff
@@ -42,10 +45,18 @@ def set_mentor
...
+
   def authorize_mentor
     authorize @mentor || Claims::Mentor
   end
 
+  def authorize_mentor_membership
+    authorize @mentor_membership || Claims::MentorMembership
```
Nitemaeric, line 57: The `|| Claims::MentorMembership` is only needed when we want to authorise collections of mentor memberships. We don't need to use this for this case.

```ruby
authorize @mentor_membership
```

is sufficient.

app/controllers/claims/support/schools/mentors_controller.rb
Nitemaeric, line 1: Same comments as previous file.

PR #576, opened 2024-05-01
Create user and add to school

gms-gs:
## Context

This allows us to:

- Update a school to be under the claims_service
- Create a user
- Associates that user to the school

Update the script to create a user and associate to a school

## Guidance to review

Run the script locally

## Link to Trello card

https://trello.com/c/ceLGrJHw/446-onboard-all-schools-and-their-first-users

### DIFF
--- a/app/services/claims/import_schools.rb
+++ b/app/services/claims/import_schools.rb
@@ -3,19 +3,32 @@ require "csv"
 class Claims::ImportSchools
   include ServicePattern
 
-  def initialize(csv_file_path:)
-    @csv_file_path = csv_file_path
+  def initialize(csv_string:)
+    @csv_string = csv_string
   end
 
   def call
-    urns = CSV.read(csv_file_path, headers: true).map { |row| row["urn"] }
+    ApplicationRecord.transaction do
+      CSV.parse(csv_string, headers: true) do |row|
+        school = School.find_by(urn: row["placement_school_urn"])
 
-    updated_count = School.where(urn: urns).update_all(claims_service: true)
-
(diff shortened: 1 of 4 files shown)


app/services/claims/import_schools.rb
```diff
@@ -8,11 +8,32 @@ def initialize(csv_file_path:)
...
+
+    users_data.each do |data|
+      school = schools.detect { |s| s.urn == data[:school_urn] }
+
+      begin
+        Claims::User.create!(first_name: data[:first_name], last_name: data[:last_name], email: data[:email], user_memberships: [UserMembership.new(organisation: school)])
+      rescue ActiveRecord::RecordInvalid => e
+        Rails.logger.error "Failed to create user: #{e.message}"
+      end
+    end
```
Nitemaeric, line 36: It's not letting me put in a suggestion, but how about:

```ruby
csv_string = # we'll be decrypting and reading from memory, so no need to base arguments on a path. If a path, it will be the encrypted file path and we'll need to decrypt here.

# Ensure all are added or none at all
ApplicationRecord.transaction do
  CSV.parse(csv_string, header: true) do |row|
    school = School.find_by!(urn: row["placement_school_urn"])

    school.update!(claims_service: true)
    school.users << Claims::User.create!(
      first_name: row["First name"],
      last_name: row["Last name"],
      email: row["Email"],
    )
  end
end
```

This updates each School slower, but this service should be infrequently run. For our dataset, it should be fine.

Nitemaeric (approved):

PR #579, opened 2024-05-01
Add mentors empty state page for claim flow

Nitemaeric:
## Context

When a User attempts to select mentors for a claim that have all fully been claimed for with the provider selected, we should show a message that stops users from proceeding further.

- Add a variant of the mentor select page that shows an error message when 
- Update `spec/system/claims/support/schools/claims/create_claim_spec.rb` to only create a single school. This removes the flakiness of this spec suite when it randomly creates 2 schools with the same name.

## Guidance to review

- Fully claim all hours for a school's mentors against a provider.
- Try to claim for a mentor against that provider.
- It should say that you have no mentors you can claim for.

## Screenshots

CleanShot 2024-05-01 at 15 25 15

https://github.com/DFE-Digital/itt-mentor-services/assets/4231530/13de0a64-6395-4472-a110-44646b26aea2

### DIFF
--- a/app/controllers/claims/schools/claims/mentors_controller.rb
+++ b/app/controllers/claims/schools/claims/mentors_controller.rb
@@ -8,13 +8,17 @@ class Claims::Schools::Claims::MentorsController < Claims::ApplicationController
 
   def create
     if claim_mentors_form.save
-      redirect_to(
-        edit_claims_school_claim_mentor_training_path(
-          @school,
-          claim,
-          claim.mentor_trainings.without_hours.first,
-        ),
-      )
+      if claim.mentor_trainings.without_hours.any?
+        redirect_to(
+          edit_claims_school_claim_mentor_training_path(
+            @school,
+            claim,
+            claim.mentor_trainings.without_hours.first,
+          ),
+        )
+      else
+        redirect_to check_claims_school_claim_path(@school, claim)
+      end
     else
       render :new
     end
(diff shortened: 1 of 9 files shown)


app/controllers/claims/schools/claims/mentors_controller.rb
```diff
@@ -8,13 +8,17 @@ def new; end
...
   def create
     if claim_mentors_form.save
-      redirect_to(
-        edit_claims_school_claim_mentor_training_path(
-          @school,
-          claim,
-          claim.mentor_trainings.without_hours.first,
-        ),
-      )
+      if claim.mentor_trainings.without_hours.any?
```
CatalinVoineag, line 11: This probably should be a method in the form object but considering the situation is fine.

app/views/claims/schools/claims/check.html.erb
```diff
@@ -62,6 +64,7 @@
                                  @claim,
                                  mentor_training,
                                ),
+                               visually_hidden_text: "Hours of training for #{mentor_training.mentor.full_name}",
```
CatalinVoineag, line 67: This should probably be a translation but I don't mind

CatalinVoineag (approved):
Played around and it looks good to me 👍🏻

app/controllers/claims/schools/claims/mentors_controller.rb
Nitemaeric: Agreed. I'll clean this up as soon as I can. Thanks!

PR #580, opened 2024-05-01
Update email content

gms-gs:
Update email content

### **Due to releasing today, values have been hardcoded for both services so we can get this over the line. This will be fixed on release and we can refactor the emailing content per service in order to not cause issues**

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -1,14 +1,14 @@
 class UserMailer < ApplicationMailer
   def user_membership_created_notification(user, organisation)
     notify_email to: user.email,
-                 subject: t(".subject", organisation_name: organisation.name),
-                 body: t(".body", user_name: user.full_name, organisation_name: organisation.name, service_name:, sign_in_url:)
+                 subject: t(".subject", service_name:),
+                 body: t(".body", user_name: user.first_name, organisation_name: organisation.name, service_name:, sign_in_url:)
   end
 
   def user_membership_destroyed_notification(user, organisation)
     notify_email to: user.email,
-                 subject: t(".subject", organisation_name: organisation.name),
-                 body: t(".body", user_name: user.full_name, organisation_name: organisation.name, service_name:)
+                 subject: t(".subject", service_name:),
+                 body: t(".body", user_name: user.first_name, organisation_name: organisation.name, service_name:)
   end
 
   def claim_submitted_notification(user, claim)
@@ -17,19 +17,24 @@ class UserMailer < ApplicationMailer
     notify_email to: user.email,
                  subject: t(".subject"),
                  body: t(".body",
+                         user_name: user.first_name,
+                         organisation_name: claim.school.name,
                          reference: claim.reference,
                          amount: claim.amount.format(symbol: true, decimal_mark: ".", no_cents: false),
                          link_to_claim:)
   end
 
-  def claim_created_support_notification(claim, email)
+  def claim_created_support_notification(claim, user)
     link_to_claim = claims_school_claim_url(id: claim.id, school_id: claim.school.id)
 
-    notify_email(to: email,
+    notify_email(to: user.email,
                  subject: t(".subject"),
                  body: t(".body",
+                         user_name: user.first_name,
+                         organisation_name: claim.school.name,
                          reference: claim.reference,
                          amount: claim.amount.format(symbol: true, decimal_mark: ".", no_cents: false),
+                         sign_in_url:,
                          link_to_claim:))
   end
 end
--- a/app/services/claims/claim/create_draft.rb
+++ b/app/services/claims/claim/create_draft.rb
@@ -18,11 +18,9 @@ class Claims::Claim::CreateDraft
   attr_reader :claim
 
   def send_claim_created_support_notification_email
-    user_emails = claim.school_users.pluck(:email)
-
-    user_emails.each do |email|
+    claim.school_users.each do |user|
       UserMailer.with(service: :claims)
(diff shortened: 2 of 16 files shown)


app/controllers/claims/schools/claims_controller.rb
```diff
@@ -49,7 +49,7 @@ def confirmation; end
   def submit
     return redirect_to rejected_claims_school_claim_path unless @claim.valid_mentor_training_hours?
 
-    Claims::Claim::Submit.call(claim: @claim, user: current_user)
+    Claims::Claim::Submit.call(claim: @claim, user: current_user, organisation: @school)
```
Nitemaeric, line 52: `school` is accessible from the `claim` 😅

app/mailers/user_mailer.rb
```diff
@@ -17,19 +17,25 @@ def claim_submitted_notification(user, claim)
...
+                         user_name: user.first_name,
+                         organisation_name: claim.school.name,
                          reference: claim.reference,
                          amount: claim.amount.format(symbol: true, decimal_mark: ".", no_cents: false),
+                         sign_in_url:,
                          link_to_claim:)
   end
 
-  def claim_created_support_notification(claim, email)
+  def claim_created_support_notification(claim, email, user)
```
Nitemaeric, line 28: Here you're accepting a `user`, but in the call below, you pass through `claim.school`.

app/services/claims/claim/create_draft.rb
```diff
@@ -22,7 +22,7 @@ def send_claim_created_support_notification_email
 
     user_emails.each do |email|
       UserMailer.with(service: :claims)
-        .claim_created_support_notification(claim, email).deliver_later
+        .claim_created_support_notification(claim, email, claim.school).deliver_later
```
Nitemaeric, line 25: Here.

app/services/claims/claim/submit.rb
```diff
@@ -20,7 +20,7 @@ def call
 
   def send_claim_submitted_notification_email
     UserMailer.with(service: :claims)
-      .claim_submitted_notification(user, claim).deliver_later
+      .claim_submitted_notification(user, claim, claim.school).deliver_later
```
Nitemaeric, line 23: You don't need to change this file.

config/locales/en/support_user_mailer.yml
```diff
@@ -5,13 +5,35 @@ en:
       body: |
         Dear %{user_name},
 
-        You have been invited to join %{service_name}.
+        You have been invited to join the %{service_name}.
```
Nitemaeric, line 8: Is this right? It reads very strangely.

"You have been invited to join the Claim funding for mentor training."

Maybe this should be?

```suggestion
        You have been invited to join the %{service_name} service.
```

spec/mailers/user_mailer_spec.rb
```diff
@@ -10,13 +10,29 @@
 
       it "sends the invitation" do
         expect(invite_email.to).to contain_exactly(user.email)
-        expect(invite_email.subject).to eq("You have been invited to #{organisation.name}")
+        expect(invite_email.subject).to eq("Invitation to join claims funding for mentor training")
```
Nitemaeric, line 13: This is incorrect. "claims funding for mentor training" is not our official service name.

app/mailers/user_mailer.rb
```diff
@@ -1,14 +1,14 @@
 class UserMailer < ApplicationMailer
   def user_membership_created_notification(user, organisation)
     notify_email to: user.email,
-                 subject: t(".subject", organisation_name: organisation.name),
-                 body: t(".body", user_name: user.full_name, organisation_name: organisation.name, service_name:, sign_in_url:)
+                 subject: t(".subject", service_name: user.service),
```
Nitemaeric, line 4: `service_name` is already available as a helper in `ApplicationMailer`. It uses looks up the official service name from our locales.

```suggestion
                 subject: t(".subject", service_name:),
```

app/mailers/user_mailer.rb
Nitemaeric, line 28: This change also makes me think, why do we need both `email` and `user`? Is `email` ever different from `user.email`?

app/mailers/user_mailer.rb
```diff
@@ -17,19 +17,25 @@ def claim_submitted_notification(user, claim)
     notify_email to: user.email,
                  subject: t(".subject"),
                  body: t(".body",
+                         user_name: user.first_name,
+                         organisation_name: claim.school.name,
                          reference: claim.reference,
                          amount: claim.amount.format(symbol: true, decimal_mark: ".", no_cents: false),
+                         sign_in_url:,
                          link_to_claim:)
```
Nitemaeric, line 25: This looks like it's no longer used.

Nitemaeric (approved):
I _think_ it's good to go now. @CatalinVoineag can you give it a final check please?

config/locales/en/user_mailer.yml
```diff
@@ -1,33 +1,82 @@
...
     claim_created_support_notification:
-      subject: A support user created your ITT mentor training claim
+      subject: New draft claim for mentor training
       body: |
-        Reference: %{reference}
-        Amount: %{amount}
+        Dear %{user_name},
+
+        We have added a draft claim for mentor training for %{organisation_name}.
+          Your claim reference is %{reference}.
```
CatalinVoineag, line 70: Is there an extra tab on line 70? Does it need to be there?

CatalinVoineag (approved):

PR #585, opened 2024-05-02
Fix back link when creating a claim

CatalinVoineag:
## Context

When creating a claim you can end up on the edit page of a claim, or mentors list.

These edit pages were programmed to always go to the check page, so when a user would click them it would try to redirect them to the check page.

This commit fixes this with a reviewed boolean field which is set to true when the user lands on the check page.

We use this boolean to know where to redirect the user, back to the check page or through the steps ending on the index page.

Once a user lands on the check page, the back buttons should always redirect him to the check page.

Reviewed boolean
Change back links
Change update success path(The continue buttons) 

## Guidance to review

Create a claim and right before getting to the check page, click the back links all the way to the index page

Create a claim and right before getting to the check page, click back link until the provider step then go forward.

Create a claim normally and edit it on the check page.

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/41b2b36c-62b5-4964-88ee-84c6d5c204e0

### DIFF
--- a/app/forms/claims/claim/mentors_form.rb
+++ b/app/forms/claims/claim/mentors_form.rb
@@ -17,6 +17,14 @@ class Claims::Claim::MentorsForm < ApplicationForm
     claim.save!
   end
 
+  def edit_back_path
+    if claim.reviewed?
+      check_claims_school_claim_path(claim.school, claim)
+    else
+      edit_claims_school_claim_path(claim.school, claim)
+    end
+  end
+
   def update_success_path
     if claim.mentor_trainings.without_hours.any?
       edit_claims_school_claim_mentor_training_path(
--- a/app/forms/claims/claim/provider_form.rb
+++ b/app/forms/claims/claim/provider_form.rb
@@ -10,7 +10,7 @@ class Claims::Claim::ProviderForm < ApplicationForm
   end
 
   def persist
-    return if claim.provider_id == provider_id
+    return true if claim.provider_id == provider_id
 
     ActiveRecord::Base.transaction do
       claim.provider_id = provider_id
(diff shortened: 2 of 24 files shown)


app/forms/claims/claim/provider_form.rb
```diff
@@ -21,6 +21,22 @@ def persist
     end
   end
 
+  def edit_back_link
```
Nitemaeric, line 24: Should this just be "back_path"?

```suggestion
  def back_path
```

app/services/claims/claim/reviewed.rb
```diff
@@ -0,0 +1,15 @@
+class Claims::Claim::Reviewed
```
Nitemaeric, line 1: The service should be a verb, in this case: `Review`.

app/forms/claims/claim/mentors_form.rb
```diff
@@ -17,6 +17,14 @@ def persist
     claim.save!
   end
 
+  def edit_back_path
```
Nitemaeric, line 20: I see we're still prefixing this with `edit_`.

Is the back path different for non-edit form flows?

Nitemaeric (approved):

app/forms/claims/claim/mentors_form.rb
```diff
@@ -17,6 +17,14 @@ def persist
...
+    if claim.reviewed
```
Nitemaeric, line 21: Actually, a nitpick, but could we add a `?` here?

```suggestion
    if claim.reviewed?
```

and the same for others

ollietreend (approved):

PR #587, opened 2024-05-02
Remove pii information from analytics.yml

CatalinVoineag:
## Context

BigQuery doesn't handle the pii information. Previously we were sending
this information and this caused the production pipeline to fail.

Adding them to the block list should fix this because we won't send pii
information

analitycs.yml files

## Guidance to review

Review code

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -24,9 +24,6 @@ shared:
     - created_at
     - updated_at
   :users:
-    - first_name
-    - last_name
-    - email
     - created_at
     - updated_at
     - id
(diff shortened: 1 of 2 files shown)


Nitemaeric (commented):
Don't we need to remove these from `analytics_pii.yml`?

CatalinVoineag:
> Don't we need to remove these from `analytics_pii.yml`?

If I do that, the specs break 🙃

Nitemaeric (approved):
Hmm, I'm unsure, but let's get it tested in an environment.

PR #589, opened 2024-05-02
Update email content from review

gms-gs:


config/locales/en/user_mailer.yml
```diff
@@ -11,7 +11,7 @@ en:
 
         If you have a DfE Sign-in account, you can use it to sign in:
 
-        [link](%{sign_in_url})
+        %{sign_in_url}
```
Nitemaeric, line 14: This will return plaintext. I think you'll want to do:

```suggestion
        [sign_in_url](%{sign_in_url})
```

CatalinVoineag (approved):

config/locales/en/user_mailer.yml
```diff
@@ -71,12 +71,12 @@ en:
 
         You can view the claim, edit and submit it on Claim funding for mentor training:
 
-        [link](%{sign_in_url})
+        %{link_to_claim}
```
Nitemaeric, line 74: Don't we want this to also be a link?

Nitemaeric (approved):
Happy once specs pass.

PR #597, opened 2024-05-02
Jc/static page markdown

JamieCleare2525:
## Context

- Add a start page for the Placements service.

- Replace sign-in page with start page for Placements service.
- Add gem to read MARKDOWN, and convert it into gov-uk styled HTML.

## Guidance to review

visit `http://placements.localhost:3000/sign-in`

## Link to Trello card

https://trello.com/c/F9sy9LTT/307-use-markdown-for-static-pages

## Things to check

## Screenshots

screencapture-placements-localhost-3000-sign-in-2024-05-02-17_04_23


app/controllers/sessions_controller.rb
```diff
@@ -2,7 +2,9 @@ class SessionsController < ApplicationController
   skip_before_action :authenticate_user!, only: %i[new callback failure]
   before_action :redirect_to_after_sign_in_path, only: %i[new], if: :user_signed_in?
 
-  def new; end
+  def new
+    render "placements/pages/start" if current_service == :placements
```
ollietreend, line 6: I've [asked about the difference](https://ukgovernmentdfe.slack.com/archives/C04MLBVP876/p1714750236261039) between Funding Mentors and School Placements, with regards to the start page vs. login page, and we're going to stick with the same approach as FM for now.

In other words: we'll still have that interstitial `/sign-in` page with the "Sign in using DfE Sign-in" button.

I think this should simplify the implementation of this start page, because we should be able to get rid of conditional logic like this.

I'd like to circle back round to this at a later date and see if we can remove that intermediary "sign in" page in a different way. Because once our service start page is hosted on GOV.UK, we won't be able to use a form POST for the "start" button. For now it's [on the backlog](https://trello.com/c/zEljg2Hv/141-remove-the-intermediary-sign-in-page-from-the-service).

config/locales/en/placements/pages.yml
```diff
@@ -1,5 +1,28 @@
...
+          If you are a school, use this service to record ITT placements at your organisation.
+
+          If you are an ITT provider, use this service to find ITT school placement opportunities.
+
+          ##Before you start##
+
+          If you are a school, you will be asked for:
+
+          - your mentors' Teacher reference number (TRN)
+          - your mentors' date of birth
```
ollietreend, line 26: I'd like to see if we can try putting this Markdown content into a `.md` file rather than storing it in the locale files.

Let's chat about this, but I'm hoping we'll be able to have a file structure like this:

```
.
└── app/
    └── views/
        └── placements/
            └── pages/
                ├── show.html.erb
                ├── _landing-page.md
                ├── _privacy-policy.md
                └── _terms-and-conditions.md
```

ollietreend (commented):
@JamieCleare2525 this is looking good – nice work!

I just have a couple of changes which I think might change the approach slightly. Happy to talk them through.

app/views/placements/pages/landing-page.md
```diff
@@ -0,0 +1,13 @@
+This service is for schools offering placements for trainee teachers and 
+Initial teacher training (ITT) providers in England 
+(accredited ITT providers and lead partners).
+
+If you are a school, use this service to record ITT placements at your organisation.
+If you are an ITT provider, use this service to find ITT school placement opportunities.
+
+##Before you start##
```
ollietreend, line 8: I'm kind of surprised that this works...? 🤔 Does it generate a H2?

I was expecting to see:

```suggestion
## Before you start
```

app/controllers/placements/pages_controller.rb
```diff
@@ -1,3 +1,9 @@
 class Placements::PagesController < ApplicationController
   skip_before_action :authenticate_user!
+
+  def start
```
ollietreend, line 4: Could you call this `landing_page` to keep it consistent with the name of the Markdown file?

```suggestion
  def landing_page
```

app/controllers/placements/pages_controller.rb
```diff
@@ -1,3 +1,9 @@
...
+    render locals: {
+      content: File.read(Rails.root.join("app/views/placements/pages/landing-page.md")),
```
ollietreend, line 6: Sorry. I realise I [asked for this](https://github.com/DFE-Digital/itt-mentor-services/pull/597#discussion_r1592056755) – but maybe it actually makes sense to use an underscore in the filename instead of a dash?

```suggestion
      content: File.read(Rails.root.join("app/views/placements/pages/landing_page.md")),
```

So as to be consistent with the action name and `.html.erb` view file.

spec/system/placements/start_page_spec.rb
```diff
@@ -0,0 +1,72 @@
+require "rails_helper"
+
+RSpec.describe "Start Page", type: :system, service: :placements do
```
ollietreend, line 3: Could you rename this to `landing_page_spec.rb`?

```suggestion
RSpec.describe "Landing Page", type: :system, service: :placements do
```

ollietreend (commented):
This looks great! 😍 Thanks for making those changes.

Just a couple of little nitpicks about naming consistency (i.e. `start` vs `landing_page`), and then I think it's good to go!

It'd also be good to tidy up the commit history a bit if possible, since you've added and then reverted some changes. I guess it could probably even just be one single commit?

app/controllers/placements/pages_controller.rb
JamieCleare2525: I called it start to keep it consistent with FM. Happy to change it though @ollietreend

app/views/placements/pages/landing-page.md
ollietreend: I've just checked locally and it _does_ generate a H2. Yay for loose standards.

ollietreend (approved):
Nice work ⭐  this looks good to me!

CatalinVoineag (approved):
Looks good

PR #601, opened 2024-05-03
Update email templates to respect service names

Nitemaeric:
## Context

When we updated the email templates for launch, we snuck in some tech debt by hardcoding the "Claims" service name.

We now need to update the templates to respect which service is calling them.

/* In the future, we may need separate templates/mailer classes for each service.

- Update templates to use `service` and `service_name` where suitable.

## Guidance to review

- Specs pass.
- @DFE-Digital/school-placements is happy? :D


ollietreend (approved):
Thanks for tidying this up @Nitemaeric 🎉

PR #608, opened 2024-05-07
Add breadcrumbs to our static pages

CatalinVoineag:
## Context

We want a way for the user to get back to the home page from our static pages. So the breadcrumbs will contain the current page and a home page.

The home page link will go to the school/claims index page if normal the user is logged in.

For support users it will redirect to organisations index page. This is because we redirect from the root path to these pages if a user is signed in.

Pages controller, views, locals

## Guidance to review

Go through the static pages and check the breadcrumbs

## Screenshots

https://github.com/DFE-Digital/itt-mentor-services/assets/11318084/74c7e7b8-78e6-42f3-b0f0-7bcce573bb6e

### DIFF
--- a/app/views/claims/pages/accessibility.en.html.erb
+++ b/app/views/claims/pages/accessibility.en.html.erb
@@ -1,4 +1,5 @@
-<% content_for :page_title, "Accessibility statement for Claim funding for mentor training" %>
+<% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: { t("home") => claims_root_path, t(".page_title") => claims_accessibility_path }, collapse_on_mobile: true) %>
+<% content_for :page_title, t(".page_title") %>
 
 <div class="govuk-width-container">
   <div class="govuk-grid-row">
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -44,6 +44,10 @@
 
     <%= yield :primary_navigation_content %>
 
+    <% if content_for?(:breadcrumbs) %>
+      <%= tag.div yield(:breadcrumbs), class: "govuk-width-container" %>
+    <% end %>
+
     <div class="govuk-width-container">
       <%= yield :before_content %>
     </div>
--- a/app/views/claims/pages/cookies.en.html.erb
+++ b/app/views/claims/pages/cookies.en.html.erb
@@ -1,4 +1,5 @@
-<% content_for :page_title, "Cookies on Claim funding for mentor training" %>
+<% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: { t("home") => claims_root_path, t(".page_title") => claims_cookies_path }, collapse_on_mobile: true) %>
+<% content_for :page_title, t(".page_title") %>
 
 <div class="govuk-width-container">
   <div class="govuk-grid-row">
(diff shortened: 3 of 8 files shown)


CatalinVoineag:
[The static pages should have breadcrumbs with a "Home" link to the "Start" page](https://trello.com/c/EaAQOWN6/461-the-static-pages-should-have-breadcrumbs-with-a-home-link-to-the-start-page)

app/views/claims/pages/accessibility.en.html.erb
```diff
@@ -1,4 +1,5 @@
-<% content_for :page_title, "Accessibility statement for Claim funding for mentor training" %>
+<% content_for :breadcrumbs, govuk_breadcrumbs(breadcrumbs: @breadcrumbs, collapse_on_mobile: true) %>
```
Nitemaeric, line 1: I wonder if the "partial" approach, similar to our `_primary_navigation` files might clean this up a tiny bit?

```suggestion
<%= render "breadcrumbs", breadcrumbs: @breadcrumbs %>
```

app/controllers/claims/pages_controller.rb
```diff
@@ -6,4 +6,39 @@ class Claims::PagesController < Claims::ApplicationController
   before_action :redirect_to_after_sign_in_path, only: :start, if: :user_signed_in?
 
   def start; end
+
+  def grant_conditions
+    @breadcrumbs = {
+      t("home") => claims_root_path,
+      t(".grant_conditions.page_title") => claims_grant_conditions_path,
+    }
+  end
```
Nitemaeric, line 15: Personally, I'm not a fan of UI setup in controller code, I'd rather set it up in the views directly.

Not a blocker though.

```erb
<%= govuk_breadcrumbs(breadcrumbs: { t("home") => claims_root_path, t(".grant_conditions.page_title") => claims_grant_conditions_path }) %>
```

CatalinVoineag: Yeah, not sure. I would put stuff like this in a decorator but there's nothing to decorate and the views are loaded with content. If we go with a partial we can't put the paths in the view

app/views/claims/pages/accessibility.en.html.erb
CatalinVoineag: I think it's not necessary, the abstraction that makes it clean is the govuk_breadcrumbs component I think. I think it's just adding indirection

app/views/layouts/application.html.erb
```diff
@@ -44,6 +44,10 @@
 
     <%= yield :primary_navigation_content %>
 
+    <div class="govuk-width-container">
+      <%= yield :breadcrumbs %>
+    </div>
```
ollietreend, line 49: It might be nice to use [content_for?](https://api.rubyonrails.org/classes/ActionView/Helpers/CaptureHelper.html#method-i-content_for-3F) here so we don't bloat the page with unnecessary empty `<div>`s.

Bonus points if you do the same for the `before_content` wrapper too ⭐

```suggestion
    <% if content_for?(:breadcrumbs) %>
      <%= tag.div yield(:breadcrumbs), class: "govuk-width-container" %>
    <% end %>
```

ollietreend (commented):
This looks good to me. It'd be nice to make the container `<div>` conditional, but it's not a blocker.

ollietreend (approved):

PR #609, opened 2024-05-07
Remove status from placement

Kizr:
## Context

Due to a redesign of what a status is, we no longer need to store this data. A status is now represented by whether there is a relationship to a provider and as such all of our existing code is redundant. This work clears the path for us to add the new status functionality.

- [x] Remove the `status` column from `placements`
- [x] Updated all related models, factories and specs
- [x] Update views to remove status
- [x] Remove status specific component

## Guidance to review

- Log in as Anne, create a placement, see that there is no mention of status
- Do the the same for Colin.

## Link to Trello card

[Remove 'status' from placements](https://trello.com/c/fHuBXmIU/305-remove-status-from-placements)

### Support
image
image

### School
image
image

### Add a placement
image


ollietreend (approved):
Lovely stuff! 😍 💥

PR #612, opened 2024-05-07
Change Add buttons to Save buttons

CatalinVoineag:
## Context

We want our `Add` buttons that create records to be named `Save`

Views
Locals
Specs

## Guidance to review

Review changes and review app

### DIFF
--- a/config/locales/en/claims/schools/users.yml
+++ b/config/locales/en/claims/schools/users.yml
@@ -17,7 +17,7 @@ en:
           page_title: Check your answers
           title: Check your answers
           caption: Add user
-          add_user: Add user
+          submit: Save user
           cancel: Cancel
           warning: "%{user_name} will be sent an email to tell them you’ve added them to %{school_name}."
           change: Change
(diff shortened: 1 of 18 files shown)


config/locales/en/claims/support/schools.yml
```diff
@@ -28,6 +28,7 @@ en:
           title_with_error: "Error: Enter a school name, URN or postcode"
         check:
           add_organisation: Add organisation
+          save_organisation: Save organisation
```
Nitemaeric, line 31: Do we still use `add_organisation`?

config/locales/en/claims/schools/users.yml
```diff
@@ -17,7 +17,7 @@ en:
           page_title: Check your answers
           title: Check your answers
           caption: Add user
-          add_user: Add user
+          add_user: Save user
```
Nitemaeric, line 20: Should we rename the key here?

I'd also push for context based keys rather than content based keys. In this case, it would probably be `submit` or along those lines.

config/locales/en/claims/support/schools.yml
CatalinVoineag: Yes, it's used in the context of the page

Screenshot from 2024-05-07 16-19-52

Nitemaeric (approved):

CatalinVoineag:
["Save" buttons should all show "Save ___" instead of "Add ___"](https://trello.com/c/Oecx1kZ3/464-save-buttons-should-all-show-save-instead-of-add)

PR #623, opened 2024-05-09
Add encrypted schools_with_emails.csv.enc file

Nitemaeric:
## Things to check


app/services/claims/update_school_emails.rb
```diff
@@ -0,0 +1,33 @@
...
+      CSV.parse(csv_string, headers: true) do |row|
+        school = School.find_by(urn: row["urn"])
+
+        if school.nil?
+          Rails.logger.info "School not found with URN: #{row["urn"]} (#{row["name"]})"
+          next
+        end
+
+        begin
+          school.update!(email_address: row["email"].downcase)
```
Nitemaeric, line 21: Let's use `normalizes` in the model, similar to the `User` model here.

app/services/claims/update_school_emails.rb
```diff
@@ -0,0 +1,35 @@
...
+        school = School.find_by(urn: row["urn"])
+
+        if school.nil?
+          Rails.logger.info "School not found with URN: #{row["urn"]} (#{row["name"]})"
+          next
+        end
+
+        email = row["email"]
+
+        unless email =~ /\A[\w+\-.]+@[a-z\d\-.]+\.[a-z]+\z/i
```
Nitemaeric, line 22: Is there a reason to not use `URI::MailTo::EMAIL_REGEXP`?

Nitemaeric: Also, it seems like we should add the email validation to the `School` model, no?

lib/tasks/update_school_emails.rake
```diff
@@ -0,0 +1,8 @@
+namespace :update_school_emails do
+  desc "Update schools emails"
+  task update_emails: :environment do
+    csv_string = Rails.application.encrypted("lib/data/schools_with_emails.csv.enc", env_key: "CSV_ENCRYPTION_KEY").read
+
+    Claims::UpdateSchoolEmails.call(csv_string:)
+  end
+end
```
Nitemaeric, line 8: This isn't `Claims` specific, should we just name this task `update_school_emails`?

`rails update_school_emails` which calls `UpdateSchoolEmails.call`.

PR #626, opened 2024-05-10
Add partner school toggle

Kizr:
## Context

Schools need to be able to filter by their partner providers to help them find placements

Adds a new partner providers filter to the find a placement page

## Guidance to review

Log in as Patricia and navigate to placements. Selecting the filter will only show partner schools. Both the clear all filters and the partner schools pill will clear this filter.

## Link to Trello card

[Change 'partner school' filter to a toggle](https://trello.com/c/bOmsoFoY)

## Screenshots

image
image

### DIFF
--- a/app/controllers/placements/providers/placements_controller.rb
+++ b/app/controllers/placements/providers/placements_controller.rb
@@ -5,7 +5,7 @@ class Placements::Providers::PlacementsController < ApplicationController
   def index
     @subjects = Subject.order_by_name.select(:id, :name)
     @school_types = compact_school_attribute_values(:type_of_establishment)
-    @schools = schools_scope.select(:id, :name)
+    @schools = schools_scope.order_by_name.select(:id, :name)
 
     @pagy, @placements = pagy(
       Placements::PlacementsQuery.call(
@@ -22,7 +22,7 @@ class Placements::Providers::PlacementsController < ApplicationController
   private
 
   def schools_scope
-    if filter_params[:only_partner_schools] && filter_params[:only_partner_schools].present?
+    if filter_params[:only_partner_schools].present?
       @provider.partner_schools
     else
       Placements::School.all
--- a/app/controllers/placements/providers/placements_controller.rb
+++ b/app/controllers/placements/providers/placements_controller.rb
@@ -22,7 +22,7 @@ class Placements::Providers::PlacementsController < ApplicationController
   private
 
   def schools_scope
-    if filter_params[:only_partner_schools]&.compact_blank&.empty?
+    if filter_params[:only_partner_schools] && filter_params[:only_partner_schools].empty?
       @provider.partner_schools
     else
       Placements::School.all
--- a/app/controllers/placements/providers/placements_controller.rb
+++ b/app/controllers/placements/providers/placements_controller.rb
@@ -54,13 +54,6 @@ class Placements::Providers::PlacementsController < ApplicationController
     }
   end
 
-  def query_params
-    {
-      filters: filter_form.query_params,
-      current_provider: @provider,
-    }
-  end
-
   def all_schools
     @all_schools ||= School.all
   end
--- a/app/controllers/placements/providers/placements_controller.rb
+++ b/app/controllers/placements/providers/placements_controller.rb
@@ -22,7 +22,7 @@ class Placements::Providers::PlacementsController < ApplicationController
   private
 
   def schools_scope
-    if filter_params[:only_partner_schools] && filter_params[:only_partner_schools].empty?
+    if filter_params[:only_partner_schools] && filter_params[:only_partner_schools].present?
       @provider.partner_schools
     else
       Placements::School.all
--- a/app/controllers/placements/providers/placements_controller.rb
+++ b/app/controllers/placements/providers/placements_controller.rb
@@ -5,7 +5,7 @@ class Placements::Providers::PlacementsController < ApplicationController
   def index
     @subjects = Subject.order_by_name.select(:id, :name)
     @school_types = compact_school_attribute_values(:type_of_establishment)
-    set_schools
+    @schools = schools_scope.select(:id, :name)
 
     @pagy, @placements = pagy(
       Placements::PlacementsQuery.call(
@@ -21,20 +21,12 @@ class Placements::Providers::PlacementsController < ApplicationController
 
   private
 
-  def set_schools
-    @schools = partner_schools_present? ? partner_schools : all_placements_schools
-  end
-
-  def partner_schools_present?
-    filter_params[:partner_schools]&.compact_blank.present?
-  end
-
-  def partner_schools
-    @provider.partner_schools.select(:id, :name)
-  end
-
-  def all_placements_schools
-    Placements::School.select(:id, :name)
+  def schools_scope
+    if filter_params[:only_partner_schools]&.compact_blank&.empty?
+      @provider.partner_schools
+    else
+      Placements::School.all
+    end
   end
 
   def set_provider
@@ -48,11 +40,18 @@ class Placements::Providers::PlacementsController < ApplicationController
   def filter_params
     params.fetch(:filters, {}).permit(
       :only_available_placements,
-      partner_school_ids: [],
+      :only_partner_schools,
       school_ids: [],
       subject_ids: [],
       school_types: [],
     )
+  end
+
+  def query_params
+    {
+      filters: filter_form.query_params,
+      current_provider: @provider,
+    }
   end
 
   def query_params
@@ -68,6 +67,6 @@ class Placements::Providers::PlacementsController < ApplicationController
 
   def compact_school_attribute_values(attribute)
     all_schools.where.not(attribute => nil)
-      .distinct(attribute).order(attribute).pluck(attribute)
+               .distinct(attribute).order(attribute).pluck(attribute)
   end
 end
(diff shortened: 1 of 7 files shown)


app/forms/placements/placements/filter_form.rb
```diff
@@ -4,7 +4,7 @@ class Placements::Placements::FilterForm < ApplicationForm
   attribute :school_ids, default: []
   attribute :subject_ids, default: []
   attribute :school_types, default: []
-  attribute :partner_school_ids, default: []
+  attribute :partner_schools, default: []
```
ollietreend, line 7: This could become a boolean and be given a more descriptive name.

```suggestion
  attribute :only_partner_schools, :boolean, default: false
```

app/queries/placements/placements_query.rb
```diff
@@ -1,17 +1,32 @@
 class Placements::PlacementsQuery < ApplicationQuery
   def call
-    scope = Placement.includes(:school, :subjects)
-
+    scope = set_initial_scope
```
ollietreend, line 3: Revert this back to:

```ruby
scope = Placement.includes(:school, :subjects)
```

and remove the methods you created:

- `set_initial_scope`
- `scope_with_partner_schools`
- `partner_schools_present?`

app/queries/placements/placements_query.rb
```diff
@@ -30,12 +45,6 @@ def school_type_condition(scope)
     scope.where(schools: { type_of_establishment: params[:school_types] })
   end
 
-  def partner_school_condition(scope)
-    return scope if params[:partner_school_ids].blank?
-
-    scope.where(school_id: params[:partner_school_ids])
-  end
-
```
ollietreend, line 38: Add this back in and make it more like this:

```ruby
def partner_school_condition(scope)
  return scope unless params[:only_partner_schools]

  provider = params[:context][:current_provider]
  scope.where(school_id: provider.partner_schools.select(:id))
end
```

Changes made:

1. Param name is now `only_partner_schools` and it's a `boolean` so we don't need to check it for 'emptiness' as if it were an array. We only care about whether it's 'truthy' or 'falsey'.
2. `params[:context][:current_provider]` – this is a `Placements::Provider` object, passed in from the controller so we know who the current provider is
3. Using `.select(:id)` instead of `.ids` means this becomes a subquery rather than an additional round-trip to the database. This should be more performant and mean we don't end up holding school IDs in memory.

app/controllers/placements/providers/placements_controller.rb
```diff
@@ -5,8 +5,7 @@ class Placements::Providers::PlacementsController < ApplicationController
   def index
     @subjects = Subject.order_by_name.select(:id, :name)
     @school_types = compact_school_attribute_values(:type_of_establishment)
-    @schools = Placements::School.select(:id, :name)
-    @partner_schools = @provider.partner_schools.select(:id, :name)
+    set_schools
 
     @pagy, @placements = pagy(
       Placements::PlacementsQuery.call(
```
ollietreend, line 11: We talked about changing this:

```ruby
Placements::PlacementsQuery.call(
  params: filter_form.query_params,
)
```

to:

```ruby
Placements::PlacementsQuery.call(
  params: query_params,
)
```

and introducing a `query_params` method in the controller, so we can include extra context and params to the query object beyond those which exist in the `FilterForm`.

```ruby
def query_params
  {
    filters: filter_form.query_params,
    current_provider: @provider,
  }
end
```

That way, the current provider becomes accessible to the query object.

app/controllers/placements/providers/placements_controller.rb
```diff
@@ -22,6 +21,22 @@ def show
...
+    filter_params[:partner_schools]&.compact_blank.present?
+  end
+
+  def partner_schools
+    @provider.partner_schools.select(:id, :name)
+  end
+
+  def all_placements_schools
+    Placements::School.select(:id, :name)
+  end
```
ollietreend, line 38: I feel like this could probably be collapsed a bit.

```suggestion
  def schools_scope
    if filter_params[:only_partner_schools]
      @provider.partner_schools
    else
      Placements::School.all
    end
  end
```

app/controllers/placements/providers/placements_controller.rb
```diff
@@ -5,8 +5,7 @@ class Placements::Providers::PlacementsController < ApplicationController
   def index
     @subjects = Subject.order_by_name.select(:id, :name)
     @school_types = compact_school_attribute_values(:type_of_establishment)
-    @schools = Placements::School.select(:id, :name)
-    @partner_schools = @provider.partner_schools.select(:id, :name)
+    set_schools
```
ollietreend, line 8: ...and then this becomes

```suggestion
    @schools = schools_scope.select(:id, :name)
```

ollietreend (commented):
As discussed, these are the things we talked through.

app/views/placements/providers/placements/_filter.html.erb
```diff
@@ -116,6 +114,22 @@
...
+            <%= form.govuk_check_boxes_fieldset(
+              :only_partner_schools,
+              legend: { text: "Partner schools", size: "s" },
+              small: true,
+              multiple: false,
+            ) do %>
+
+              <%= form.govuk_check_box :only_partner_schools,
+                                       true,
+                                       label: { text: "Only show placements offered by my partner schools" },
```
ollietreend, line 127:
```suggestion
              legend: { text: t(".partner_schools"), size: "s" },
              small: true,
              multiple: false,
            ) do %>

              <%= form.govuk_check_box :only_partner_schools,
                                       true,
                                       label: { text: t(".only_show_partner_schools") },
```

spec/forms/placements/placements/filter_form_spec.rb
```diff
@@ -110,19 +110,19 @@
...
             provider:,
-            filter: "partner_school_ids",
-            value: "school_id_1",
+            filter: "only_partner_schools",
+            value: false,
           ),
         ).to eq(
           placements_provider_placements_path(provider, filters: {
-            partner_school_ids: %w[school_id_2],
+            partner_schools: [],
```
ollietreend, line 125: Shouldn't this be:

```suggestion
            only_partner_schools: false,
```

? 🤔

I think we're getting a false-positive because `partner_schools: []` is no longer a defined filter.

spec/forms/placements/placements/filter_form_spec.rb
```diff
@@ -110,19 +110,19 @@
 
     context "when removing partner school id params" do
```
ollietreend, line 111: We might want to reword the tests and context blocks slightly here to reflect that we're no longer dealing with "partner school IDs":

```suggestion
    context "when removing the partner schools filter" do
```

ollietreend (commented):
I'm sorry for another heap of comments @Kizr 🙈 

These are all pretty small things now. Overall I think this PR is looking in really good shape 🙌🏻 I'm glad to see it's all working as we'd hoped!

app/controllers/placements/providers/placements_controller.rb
```diff
@@ -22,6 +21,14 @@ def show
 
   private
 
+  def schools_scope
+    if filter_params[:only_partner_schools] && filter_params[:only_partner_schools].present?
```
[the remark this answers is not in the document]
Kizr: Yeah that was a suggestion from the IDE, but now it's okay with it? I dunno lol.

app/forms/placements/placements/filter_form.rb
```diff
@@ -22,9 +22,11 @@ def clear_filters_path(provider)
   end
 
   def index_path_without_filter(provider:, filter:, value: nil)
-    without_filter = compacted_attributes.merge(
-      filter => compacted_attributes[filter].reject { |filter_value| filter_value == value },
-    )
+    without_filter = if BOOLEAN_ATTRIBUTES.include?(filter)
```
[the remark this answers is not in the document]
Kizr: I'll leave it as is, but it's cool!

app/queries/placements/placements_query.rb
```diff
@@ -3,42 +3,47 @@ def call
...
-    return scope if params[:school_types].blank?
+    return scope if filter_params[:school_types].blank?
 
-    scope.where(schools: { type_of_establishment: params[:school_types] })
+    scope.where(schools: { type_of_establishment: filter_params[:school_types] })
   end
 
   def partner_school_condition(scope)
-    return scope if params[:partner_school_ids].blank?
+    return scope if filter_params[:only_partner_schools] == "false" || filter_params[:only_partner_schools].blank?
```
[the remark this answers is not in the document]
Kizr: I think this snuck in during my changes from a string to a boolean and flew under the radar

app/queries/placements/placements_query.rb
```diff
@@ -3,42 +3,47 @@ def call
...
   def available_condition(scope)
-    return scope if params[:available].blank?
+    return scope if filter_params[:available].blank?
 
     scope.where(provider_id: nil)
   end
+
+  def filter_params
+    @filter_params ||= params.fetch(:filters, {}).each_value { |v| v.compact_blank! if v.is_a?(Array) }
+  end
```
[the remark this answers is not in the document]
Kizr: Another one that slipped through due to the amount of changes, it isn't needed :)

JamieCleare2525 (approved):

PR #628, opened 2024-05-12
As a support user I can remove a schools grant conditions agreement

gms-gs:
## Context

As a support user I should be able to remove the accepted grant conditions

Add a link to allow a support user to remove a schools grant conditions

## Guidance to review

- Sign in as a school
- Accept the grant conditions
- Sign in as a support user
- Visit that school and follow the flow

## Link to Trello card

https://trello.com/c/kwsmnRfQ/486-as-a-support-user-i-can-remove-a-schools-grant-conditions-agreement

## Screenshots
Screenshot 2024-05-12 at 17 33 30

Screenshot 2024-05-12 at 17 33 37
Screenshot 2024-05-12 at 17 33 43

### DIFF
--- a/app/models/claims/school.rb
+++ b/app/models/claims/school.rb
@@ -66,6 +66,8 @@
 #  fk_rails_...  (trust_id => trusts.id)
 #
 class Claims::School < School
+  audited
+
   default_scope { claims_service }
 
   has_many :users, through: :user_memberships
--- a/app/views/claims/support/schools/remove_grant_conditions_acceptance_check.html.erb
+++ b/app/views/claims/support/schools/remove_grant_conditions_acceptance_check.html.erb
@@ -1,0 +1,18 @@
+<% content_for :page_title, t(".page_title", user_name: @school.claims_grant_conditions_accepted_by.full_name) %>
+
+<% content_for(:before_content) do %>
+  <%= govuk_back_link(href: claims_support_school_path(@school)) %>
+<% end %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-l"><%= t(".agreement") %></span>
+      <h1 class="govuk-heading-l"><%= t(".page_title", user_name: @school.claims_grant_conditions_accepted_by.full_name) %></h1>
+    </div>
+  </div>
+
+  <%= govuk_button_to(t(".remove_agreement"), remove_grant_conditions_acceptance_claims_support_school_path(@school), warning: true, method: :put, class: "govuk-button govuk-button--warning") %>
+
+  <%= govuk_link_to t(".cancel"), claims_support_school_path(@school), no_visited_state: true %>
+</div>
--- a/config/routes/claims.rb
+++ b/config/routes/claims.rb
@@ -61,6 +61,11 @@ scope module: :claims, as: :claims, constraints: {
     end
 
     resources :schools, except: %i[destroy update] do
+      member do
+        put :remove_grant_conditions_acceptance, path: "remove-grant-conditions-acceptance"
+        get :remove_grant_conditions_acceptance_check, path: "remove-grant-conditions-acceptance"
+      end
+
       collection do
         get :check
         get :check_school_option
--- a/spec/system/claims/support/schools/remove_grant_conditions_spec.rb
+++ b/spec/system/claims/support/schools/remove_grant_conditions_spec.rb
@@ -1,0 +1,45 @@
+require "rails_helper"
+
+RSpec.describe "Remove a users agreement to the grant conditions for a school", type: :system do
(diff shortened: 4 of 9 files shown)


app/models/claims/school.rb
```diff
@@ -78,4 +79,10 @@ class Claims::School < School
   def grant_conditions_accepted?
     claims_grant_conditions_accepted_at?
   end
+
+  def remove_grant_conditions_agreement
+    self.claims_grant_conditions_accepted_at = nil
+    self.claims_grant_conditions_accepted_by_id = nil
+    save!
+  end
```
Nitemaeric, line 87: Let's avoid adding mutative methods to models. I'd suggest `#update!(claims_grant_conditions_accepted_at: nil, claims_grant_conditions_accepted_by: nil)` in the controller directly for this.

The alternative would be to add a service named `Claims::School::RemoveGrantConditionsAcceptance`.

app/policies/claims/support_user_policy.rb
```diff
@@ -4,6 +4,6 @@ def destroy?
   end
 
   def remove?
-    destroy?
+    user.support_user?
   end
```
Nitemaeric, line 8: This is a good catch, but I think it should go into the `destroy?` method, no?

```ruby
def destroy?
  user != record && user.support_user?
end
```

We can remove `remove?`, as it's already defined in `Claims::ApplicationPolicy`.

app/controllers/claims/support/schools/grant_conditions_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    @school.remove_grant_conditions_agreement
+    redirect_to claims_support_school_path(@school), flash: { success: t(".success") }
+  end
+
+  private
+
+  def authorize_support_user
+    authorize @support_user || Claims::SupportUser
+  end
+end
```
Nitemaeric, line 18: This is a good way to structure the "grant conditions acceptance" as a resource, but in this case, especially as we're only mutating a School record, we can do:

```ruby
class Claims::Support::SchoolsController
  # PUT /support/schools/:school_id/remove_grant_conditions_acceptance
  def remove_grant_conditions_acceptance
    # ...
  end
end
```

`PUT` instead of `DELETE` as we're mutating the School row, not dropping any rows.

app/models/claims/school.rb
```diff
@@ -66,6 +66,7 @@
 #  fk_rails_...  (trust_id => trusts.id)
 #
 class Claims::School < School
+  audited
   default_scope { claims_service }
```
Nitemaeric, line 70: Can we add an empty line between these 2 unrelated lines please?

config/routes/claims.rb
```diff
@@ -61,6 +61,11 @@
     end
 
     resources :schools, except: %i[destroy update] do
+      member do
+        put :remove_grant_conditions_acceptance
+        get :remove_grant_conditions_acceptance_check
+      end
```
Nitemaeric, line 67: To make these seamless to the user, can we use the same path please? Also, Simon has nudged us to use dashes for word breaks.

```suggestion
      member do
        put :remove_grant_conditions_acceptance, path: "remove-grant-conditions-acceptance"
        get :remove_grant_conditions_acceptance_check, path: "remove-grant-conditions-acceptance"
      end
```

app/views/claims/support/schools/remove_grant_conditions_acceptance_check.html.erb
```diff
@@ -0,0 +1,18 @@
+<% content_for(:before_content) do %>
+  <%= govuk_back_link(href: claims_support_school_path(@school)) %>
+<% end %>
+
+<% content_for :page_title, t(".page_title", user_name: @school.claims_grant_conditions_accepted_by.full_name) %>
```
Nitemaeric, line 5: To associate this with how things are rendered on the page, can we bring this up to line 1 and then call the `before_content` block please?

spec/system/claims/support/schools/remove_grant_conditions_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "rails_helper"
+
+RSpec.describe "Remove a users agreement to the grant conditions for a school", type: :system do
+  let!(:support_user) { create(:claims_support_user, :colin) }
+  let!(:claims_user) { create(:claims_user, first_name: "George", last_name: "Barlow") }
+
+  let!(:school) { create(:school, :claims, claims_grant_conditions_accepted_at: Time.zone.now, claims_grant_conditions_accepted_by_id: claims_user.id) }
```
Nitemaeric, line 7:
```suggestion
  let!(:school) { create(:claims_school, claims_grant_conditions_accepted_at: Time.zone.now, claims_grant_conditions_accepted_by_id: claims_user.id) }
```

Nitemaeric (approved):
1 minor thing and good to 🚀

PR #637, opened 2024-05-14
Replace 'school type' with 'establishment group'

ollietreend:
## Context

The product & design team have decided to stop using the 'school type' field in favour of the 'establishment group' field that we import from GIAS. This is intended to be a service-wide change.

- Remove the 'school type' field from all views in the service
- Replace it with 'establishment group'
- Change the placements search filter to use 'establishment group' instead of 'school type'

Note that I have not removed 'school type' from the GIAS import. It shouldn't do any harm to keep this field in the database in case it's needed again in future.

## Guidance to review

- Do the tests pass?
- Does the review app look okay?

## Link to Trello card

https://trello.com/c/XKwyTqcA/314-change-school-type-to-establishment-group-service-wide


JamieCleare2525 (approved):
Tested locally. All working as expected.

PR #639, opened 2024-05-15
Move find placements to top level

Kizr:
## Context

We will need to add a "My placements" page in the future that should live at the existing "Find placements" page, this refactors "Find placements" in preparation for the "My placements" work.

- [x] Moves all of the "Find placements" logic out of the Provider namespace and up one level.
- [x] Sets the provider ID when we know who the provider is, or when the user selects their organisation. 

## Guidance to review

Sign in as Patricia and verify that everything continues to work as expecetd. The only visible change is that the new URL will just be `/placements`

## Link to Trello card

[Move placements search results to the top-level namespace](https://trello.com/c/hEGTjNTy)


app/controllers/placements/providers/partner_schools_controller.rb
```diff
@@ -80,6 +80,7 @@ def destroy
   private
 
   def set_provider
+    session[:placements_provider_id] = params[:provider_id]
```
JamieCleare2525, line 83: Isn't the session already set at this point?

app/controllers/placements/providers/users_controller.rb
```diff
@@ -8,4 +10,8 @@ def set_organisation
   def redirect_to_index
     redirect_to placements_provider_users_path(@organisation)
   end
+
+  def set_provider
```
JamieCleare2525, line 14: Isn't this session already set at this point? And should this change the `set_organisation` method?

app/helpers/placements/routes/organisations_helper.rb
```diff
@@ -37,6 +37,7 @@ def placements_organisation_path(organisation)
     when School
       placements_school_placements_path(organisation)
     when Provider
+      session[:placements_provider_id] = organisation.id
```
JamieCleare2525, line 40: Not sure this is necessary

CatalinVoineag (approved):

PR #644, opened 2024-05-15
Convert to browser back button behaviour to retain filters

Kizr:
https://trello.com/c/VDSgNUdW

## Context

When a user sets filters when finding a placement they are lost if the user views the placement and presses the back button.

- [x] Use rails `:back` link to emulate browser back button behaviour which retains the filters in the URL.
- [x] Adds a presence check for a filter that was missed in a spec

## Guidance to review

- Log in as Patricia
- Navigate to Placements
- Set some flters
- View a placement
- Press the back button
- Confirm that filters are still present

## Link to Trello card

[The 'back' link on the placement details page should remember the user's search parameters](https://trello.com/c/VDSgNUdW)

### DIFF
--- a/spec/system/placements/providers/placements/view_placements_list_spec.rb
+++ b/spec/system/placements/providers/placements/view_placements_list_spec.rb
@@ -176,6 +176,7 @@ RSpec.describe "Placements / Providers / Placements / View placements list",
         )
         then_i_can_see_a_placement_for_school_and_subject("Primary School", "Primary with mathematics")
         and_i_can_not_see_a_placement_for_school_and_subject("Secondary School", "Chemistry")
+        and_i_can_see_a_preset_filter("Placements Available", "Placements Available")
         and_i_can_see_a_preset_filter("Partner schools", "Partner schools")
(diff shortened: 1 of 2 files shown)


spec/system/placements/providers/placements/view_placements_list_spec.rb
```diff
@@ -225,6 +226,17 @@
...
+    context "when a user views a placement and returns to the index page" do
+      scenario "User filters are preserved when using the back button" do
+        when_i_visit_the_placements_index_page
+        when_i_check_filter_option("only-available-placements", true)
+        and_i_click_on("Apply filters")
+        and_i_navigate_to("Primary with mathematics")
+        and_i_click_on("Back")
+        then_i_can_see_a_preset_filter("Placements Available", "Placements Available")
+      end
+    end
```
ollietreend, line 239: Does this change also preserve the search location and page number?

It might be nice to cover those here too, just to be really explicit about it.

ollietreend (approved):
Nice! 😍 Thanks for bolstering that system test.

PR #656, opened 2024-05-19
Add Claims service Slack notification when a Claim has been submitted

Nitemaeric:
## Context

We want to know when a Claim has been made.

- Introduce a `SlackNotifier` library class.

  The `SlackNotifier` is app-agnostic and is therefore placed in the `lib` directory.

  The API signature mirrors `ActionMailer` and therefore should provide a very familiar developer experience.

  We could update the class to use a `defaults` method, as `ActionMailer` does, to set default values on the `message` call.

  Where `to` needs to be an array of email addresses for `ActionMailer`, `SlackNotifier` should be able to configure a `to` of a Slack channel ID. Alternatives, we could name the parameter `channel`. We also provide a `endpoint_url` value that acts as the _delivery_method_ (or SMTP configuration) comparable.

- To provide a clear separation of concerns between app-specific code vs app-agnostic code, add the "service" logic to an `ApplicationSlackNotifier`, as we do with `ActionMailer`.

- Add a `Claims::ApplicationSlackNotifier` class that sets `service` and `endpoint_url` for the "Claims" service usage.

- Add a `Claims::ClaimSlackNotifier` to send claim-subject notifications.

- Call `Claims::ClaimSlackNotifier` in the `Claims::Claim::Submit` service.

## Guidance to review

- Review the code and code architecture.
- Ping me for a user-based `webhook_url` and I'll provide you with one for local testing.

## Screenshots

CleanShot 2024-05-19 at 23 50 17

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -20,3 +20,5 @@ TEACHING_RECORD_BASE_URL=https://preprod.teacher-qualifications-api.education.go
 TEACHING_RECORD_API_MINOR_VERSION=20240416
 TEACHING_RECORD_API_KEY=secret
 GIAS_CSV_BASE_URL=https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public
+
+CLAIMS_SLACK_WEBHOOK_URL=https://hooks.slack.com/services/secret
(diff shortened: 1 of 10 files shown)


JamieCleare2525 (approved):
Tested locally. All appears to work.

PR #658, opened 2024-05-20
[BUGFIX] Prevent ApplicationSlackNotifier from failing when PORT is missing

Nitemaeric:
## Context

The `ApplicationSlackNotifier` is failing when `ENV["PORT"]` is missing.

The `ApplicationSlackNotifier` should default to no specific port, i.e. be port-less (port 80 by default on browsers).

- Use `ENV["PORT"]` instead of `ENV.fetch("PORT")` and allow `nil` for missing `PORT` env var.

## Guidance to review

- Remove `PORT` from your local env vars, it should now work fine.
  - The host for urls built in the slack notifier will be `claims.localhost` instead of `claims.localhost:3000` (if your PORT is set to 3000). In Production, this is the intended behaviour.

### DIFF
--- a/app/slack_notifiers/application_slack_notifier.rb
+++ b/app/slack_notifiers/application_slack_notifier.rb
@@ -6,7 +6,7 @@ class ApplicationSlackNotifier < SlackNotifier
   private
 
   def default_url_options
-    { host:, port: ENV.fetch("PORT") }
+    { host:, port: ENV["PORT"] }
   end
 
   def host


Kizr (commented):
~Won't this still fail, albeit more quietly now? Should we be adding the port number and keeping the fetch, so that we can spot this error in the future?~

I reviewed before seeing the Guidance section, in which case this is not an issue.

app/slack_notifiers/application_slack_notifier.rb
```diff
@@ -6,7 +6,7 @@ class ApplicationSlackNotifier < SlackNotifier
   private
 
   def default_url_options
-    { host:, port: ENV.fetch("PORT") }
+    { host:, port: ENV["PORT"] }
```
JamieCleare2525, line 9: If we can use this without `port` do we even need it as an arguments?

Nitemaeric:
> Won't this still fail, albeit more quietly now? Should we be adding the port number and keeping the fetch, so that we can spot this error in the future?

No, it shouldn't break anything. The change is that `development` will build incorrect urls, production does not typically expect any `PORT` value.

app/slack_notifiers/application_slack_notifier.rb
Nitemaeric: We need it for development, but I can clean it up post fix.

JamieCleare2525 (approved):
Happy to approve this, as a bug fix.

PR #660, opened 2024-05-20
Update subject and placement data model for additional subjects

JamieCleare2525:
## Context

- With the redesign to include "Additional Subjects", the data model has been changed to allow the following:
  - Subjects belongs to a Parent Subject (optional)
  - A Subject has many Child Subjects
  - A Placement belongs to a Subject (optional for now, but filled by `after_save`)
  - A Placement has many Placement Additional Subjects
  - A Placement has many Additional Subjects, through Placement Additional Subjects

- Changes to Subject and Placement models to allow for Placements to have Additional Subjects
- Changes to Placement so that it belongs to a Subject

## Guidance to review

in the rails console:

```ruby
placement = Placement.new(school: Placements::School.last, subjects: [Subject.last])
placement.save!
placement.subject # This should return the same record as Subject.last
```

## Link to Trello card

https://trello.com/c/N1R0wbsY/369-update-the-data-model-to-support-parent-child-associations-on-subjects

## Things to check

### DIFF
--- a/app/models/placement.rb
+++ b/app/models/placement.rb
@@ -22,7 +22,8 @@
 #  fk_rails_...  (subject_id => subjects.id)
 #
 class Placement < ApplicationRecord
-  after_save :assign_subject # TODO: Remove when data migrated
+  # TODO: Remove when data migrated
+  after_save :assign_subject, unless: proc { |placement| placement.subject.present? }
 
   has_many :placement_mentor_joins, dependent: :destroy
   has_many :mentors, through: :placement_mentor_joins, class_name: "Placements::Mentor"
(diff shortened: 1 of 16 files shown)


app/models/placement.rb
```diff
@@ -47,4 +55,13 @@ def self.order_by_school_ids(school_ids)
     end
     order(condition)
   end
+
+  # TODO: Remove after data migrated
+  def assign_subject
+    return if subjects.blank? ||
+      subject.present? ||
+      subject == subjects.last
```
Kizr, line 63: `subjects.exists?` might be more performant here?

```suggestion
    return if subjects.exists? ||
      subject.present? ||
      subject == subjects.last
```

We could also predicate this `after_save` on a Proc instead of returning inside the method? Not a huge deal though :)

Kizr: I was thinking along the lines this: [9.2 Using :if and :unless with a Proc](https://guides.rubyonrails.org/active_record_callbacks.html#using-if-and-unless-with-a-proc)

Kizr (approved):

CatalinVoineag (approved):

PR #673, opened 2024-05-23
Add data migration to associate modern language child subjects

JamieCleare2525:
## Context

- Changes to update Subjects, related to the "Modern Languages", to associated them with the parent subject "Modern Languages" 

- `PublishTeacherTraining::Subject::Import` considers `ModernLanguagesSubject` its own subject area.
- `PublishTeacherTraining::Subject::Import` assigns subjects within the `ModernLanguagesSubject` subject area with the parent subject "Modern Languages"
- Added Data Migration to backfill database with child subject associations for "Modern Languages" associated subjects.

## Guidance to review

- run `bundle exec rake data:migrate`
- This will backfill and language subjects with the parent subject "Modern Languages"
- in the rails console run: `Subject.find_by(name: "Modern Languages").child_subjects`
- This should return [French, German...]

- run `bundle exec rake db:drop db:create db:setup`
- in the rails console run: `Subject.find_by(name: "Modern Languages").child_subjects`
- This should return [French, German...]

 in the rails console run:
```ruby
# Remove all parent subjects
Subject.update_all(parent_subject_id: nil)

# Run the subject importer
PublishTeacherTraining::Subject::Import.call

Subject.find_by(name: "Modern Languages").child_subjects
# This should return [French, German...]
```

## Link to Trello card

https://trello.com/c/BQwACPHr/372-update-subjects-to-use-parent-child-associations

### DIFF
--- a/app/services/publish_teacher_training/subject/import.rb
+++ b/app/services/publish_teacher_training/subject/import.rb
@@ -3,8 +3,9 @@ module PublishTeacherTraining
     class Import
       include ServicePattern
 
-      PRIMARY_IDS = %w[PrimarySubject].freeze
-      SECONDARY_IDS = %w[SecondarySubject ModernLanguagesSubject].freeze
+      PRIMARY_ID = "PrimarySubject".freeze
+      SECONDARY_ID = "SecondarySubject".freeze
+      MODERN_LANGUAGES_ID = "ModernLanguagesSubject".freeze
 
       def call
         fetch_subject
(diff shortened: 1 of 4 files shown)


db/data/20240522123948_update_modern_language_child_subjects.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    subjects_data = api_response.fetch("included")
+
+    assign_parent_subject(parent_subject: modern_languages, child_subject_ids:, subjects_data:)
+  end
+
+  def down
+    return if modern_languages.blank?
+
+    Subject.where(parent_subject: modern_languages).update_all(parent_subject_id: nil)
```
Nitemaeric, line 14: Strong migrations will prevent us from using our app models in migrations, as when the code is eager-loaded, previous definitions will break.

A recommendation is to create a migration-scoped AR class.

```ruby
class SomeMigration
  class SubjectForMigration
    self.table_name = "subjects"
  end

  def up
    SubjectForMigration.where(...).update_all(...)
  end
end
```

This way, app changes don't break migrations, they're self-contained.

JamieCleare2525: This is not a migration. This is a [data-migration](https://github.com/ilyakatz/data-migrate) and strong migration does not apply.
This is also a point in time migration of data, and will likely be removed after a certain amount of time has past.

app/services/publish_teacher_training/subject/import.rb
```diff
@@ -33,36 +37,52 @@ def fetch_subject
           subject_ids: secondary_subject_ids,
           subjects_data:,
         )
+
+        # Sync Modern Language Subjects
+        sync_subjects(
+          subject_area: :secondary,
+          subject_ids: modern_language_subject_ids,
+          subjects_data:,
+          parent_subject: ::Subject.find_by(name: "Modern Languages"),
```
ollietreend, line 46: Would this be safer as:

```suggestion
          parent_subject: ::Subject.find_by!(name: "Modern Languages"),
```

so it explodes if the MFL subject can't be found?

Otherwise I think this will run with `parent_subject: nil`, so the MFL subjects become top level.

ollietreend (approved):
Looks good to me 👍🏻

PR #680, opened 2024-05-24
Add additional subjects journey

Kizr:
## Context

We need to support additional subjects for modern foreign languages to match the new format that design have implemented, this allows users to select which languages for the top level subject "Modern Foreign Languages" the placement is for.

- Updates references to `subjects` to be `subject` after @JamieCleare2525's work refactoring the way we interact with subjects for placements
- Adds the additional subject page that allows us to select options for subjects which have children, e.g. modern foreign languages
- Updates the placement decorator to work with these changes 

## Guidance to review

- Log in as Anne
- Add a placement
- Select Modern Languages
- Select which languages the placement is for
- Continue through the flow, until you get to check your answers
- See your chosen languages displayed
- Click on change and change the languages
- Continue through the flow, until you get to check your answers
- See that your selection has updated
- Create the placement
- Verify that it has been created

## Link to Trello card

[Update the ‘Add a placement’ and ‘Change subject’ journeys](https://trello.com/c/nmNKh8kE/371-update-the-add-a-placement-and-change-subject-journeys)

## Screenshots

NB: Languages are set up locally and do not reflect the production data.

image
image
image


app/models/placements/schools/placements/build/placement.rb
```diff
@@ -11,7 +11,7 @@
 #
 # Indexes
 #
-#  index_placements_on_provider_id  (provider_id)
+#  index_placements_on_provider_id  (provider_id)a
```
JamieCleare2525, line 14: What's this "a" doing?

Kizr: Just hanging out 😅 I'll move them along.

app/views/placements/schools/placements/build/check_your_answers.html.erb
```diff
@@ -30,10 +30,18 @@
 
           <% summary_list.with_row do |row| %>
             <% row.with_key(text: t(".subject")) %>
-            <% row.with_value(text: @placement.subjects.map(&:name).to_sentence) %>
+            <% row.with_value(text: @placement.subject.name) %>
             <% row.with_action(text: t(".change"), href: add_subject_placements_school_placement_build_index_path(@school, :add_subject), visually_hidden_text: t(".add_subject")) %>
           <% end %>
 
+          <% if @placement.subject.has_child_subjects? %>
```
JamieCleare2525, line 37: I believe you've delegated this method in your `Placement` model. 
```suggestion
          <% if @placement.has_child_subjects? %>
```

app/views/placements/schools/placements/build/check_your_answers.html.erb
```diff
@@ -30,10 +30,18 @@
...
+            <% summary_list.with_row do |row| %>
+              <% row.with_key(text: t(".additional_subjects")) %>
+              <% row.with_value(text: @placement.additional_subjects.map(&:name).to_sentence) %>
```
JamieCleare2525, line 40: Can you use the decorator method here. That way should anything change you only need to change it in one place?

app/views/placements/schools/placements/show.html.erb
```diff
@@ -27,11 +27,7 @@
         <% summary_list.with_row do |row| %>
           <% row.with_key(text: t(".attributes.placements.subject")) %>
           <% row.with_value do %>
-            <ul class="govuk-list">
```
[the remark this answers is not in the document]
JamieCleare2525: It might be worth using a partial for the placement details and conditional change actions.

app/views/placements/support/schools/placements/show.html.erb
```diff
@@ -24,11 +24,7 @@
         <% summary_list.with_row do |row| %>
           <% row.with_key(text: t(".attributes.placements.subject")) %>
           <% row.with_value do %>
```
JamieCleare2525, line 26: See comment made above.

config/locales/en/placements/schools/placements.yml
```diff
@@ -32,6 +34,12 @@ en:
             subject: Subject
             continue: Continue
             cancel: Cancel
+          add_additional_subjects:
+            title: Add placement
+            title_with_error: "Error: Select an additional subject"
```
JamieCleare2525, line 39:
```suggestion
            title_with_error: "Error: Additional subject - Add placement"
```

config/locales/en/placements/schools/placements.yml
```diff
@@ -32,6 +34,12 @@ en:
             subject: Subject
             continue: Continue
             cancel: Cancel
+          add_additional_subjects:
+            title: Add placement
```
JamieCleare2525, line 38:
```suggestion
            title: Additional subject - Add placement
```

db/schema.rb
```diff
@@ -68,9 +68,6 @@
     t.index ["submitted_by_type", "submitted_by_id"], name: "index_claims_on_submitted_by"
   end
 
-  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
```
JamieCleare2525, line 71: This shouldn't have been removed https://github.com/ilyakatz/data-migrate#installation

spec/components/placement/summary_component_spec.rb
```diff
@@ -52,48 +52,75 @@
     end
   end
 
-  context "when given a single placements" do
+  context "when given a single placement" do
+    let(:subject) { subject_1 }
+    let(:subject_1) { create(:subject, name: "Biology") }
```
JamieCleare2525, line 57: This is duplicated at the top of this file

db/schema.rb
Kizr: I have no idea how I've manged to remove that 🤯

app/views/placements/schools/placements/build/check_your_answers.html.erb
```diff
@@ -30,10 +30,18 @@
...
             <% row.with_key(text: t(".subject")) %>
-            <% row.with_value(text: @placement.subjects.map(&:name).to_sentence) %>
+            <% row.with_value(text: @placement.subject.name) %>
             <% row.with_action(text: t(".change"), href: add_subject_placements_school_placement_build_index_path(@school, :add_subject), visually_hidden_text: t(".add_subject")) %>
           <% end %>
 
+          <% if @placement.subject_has_child_subjects? %>
+            <% summary_list.with_row do |row| %>
+              <% row.with_key(text: t(".additional_subjects")) %>
+              <% row.with_value(text: @placement.additional_subjects.map(&:name).to_sentence) %>
```
JamieCleare2525, line 40: Could we put this code in the `Placement` model as method? (Assuming we are using this in the decorator as well)

app/models/placements/schools/placements/build/placement.rb
```diff
@@ -87,4 +87,8 @@ def build_phase(phase)
     phase.presence ||
       (school.primary_or_secondary_only? ? school.phase : "Primary")
   end
+
+  def additional_subject_names
```
JamieCleare2525, line 91: Why is this only in the `build` version? Might as well be in the parent `Placement` model, and simplify the decorator

JamieCleare2525: Just a thought, you could try using this example in your controller to decorate the build version of a placement.
https://github.com/drapergem/draper#decorating-objects

JamieCleare2525 (approved):

app/views/placements/schools/placements/index.html.erb
```diff
@@ -20,7 +20,7 @@
               <% @placements.each do |placement| %>
                 <% body.with_row do |row| %>
                   <% row.with_cell(text: govuk_link_to(
-                    placement.subject_names,
+                    placement.subject_name,
```
ollietreend, line 23: In fact, maybe we actually _should_ do the refactor in this PR. Since we're already renaming `#subject_names` to `#subject_name`, maybe we should just go ahead and rename it to `#title`.

app/views/placements/schools/placements/build/check_your_answers.html.erb
```diff
@@ -30,10 +30,18 @@
 
           <% summary_list.with_row do |row| %>
             <% row.with_key(text: t(".subject")) %>
-            <% row.with_value(text: @placement.subjects.map(&:name).to_sentence) %>
+            <% row.with_value(text: @placement.subject.name) %>
             <% row.with_action(text: t(".change"), href: add_subject_placements_school_placement_build_index_path(@school, :add_subject), visually_hidden_text: t(".add_subject")) %>
           <% end %>
 
+          <% if @placement.subject_has_child_subjects? %>
```
ollietreend, line 37: Would this `if` statement be more appropriate as:

```suggestion
          <% if @placement.has_additional_subjects? %>
```

in other words

```suggestion
          <% if @placement.additional_subjects.present? %>
```

i.e. checking whether additional subjects are present on the placement, rather than if the subject has children?

It's not a strong opinion, but somehow it feels more semantically correct.

config/locales/en/placements/schools/placements.yml
```diff
@@ -8,8 +8,10 @@ en:
               invalid: Select a phase
             mentor_ids:
               invalid: Select a mentor or not yet known
-            subject_ids:
+            subject:
               invalid: Select a subject
+            additional_subject_ids:
+              invalid: Select an additional subject
```
ollietreend, line 14: I'm not _entirely sure_ if "additional subject" is the right term to use here. While we've modelled it internally as "additional subjects", the prototype & UI doesn't ever call them that.

But I think this should be fine for now. I'd rather not over-complicate things or pre-empt the findings of the upcoming content review.

app/models/placement.rb
```diff
@@ -63,4 +64,8 @@ def assign_subject
 
     update!(subject: subjects.last)
   end
+
+  def additional_subject_names
+    additional_subjects.map(&:name).sort
+  end
```
ollietreend, line 70: I'm not entirely convinced this needs to be on the model. It feels more appropriate for the decorator since it's primarily a view concern (and in fact, that's the only place this method is used currently.)

app/controllers/placements/schools/placements/build_controller.rb
```diff
@@ -25,8 +33,10 @@ def add_mentors
   end
 
   def check_your_answers
-    @placement = initialize_placement
+    @placement = initialize_placement.decorate
```
ollietreend, line 36: Hooray 🙌🏻  you got the placement decorator working on `Placements::Schools::Placements::Build::Placement` instances!

ollietreend (commented):
Nice work on this @Kizr! 🙌🏻  Looks like this was a mammoth one, not least because of all the spec files you needed to fix!

Just a couple of comments from me. The main one is about the placement decorator – let's talk it through.

app/decorators/placement_decorator.rb
```diff
@@ -10,16 +10,18 @@ def mentor_names
     end
   end
 
-  def subject_names
-    if subjects.any?
-      subjects.pluck(:name).sort.to_sentence
+  def subject_name
+    if additional_subjects.exists? || additional_subjects.present?
```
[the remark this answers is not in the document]
Kizr: Yes that's right, because we're using the decorator in both the built placement and the saved placement we can get a minor performance gain by using `.exists?` over `.present?` but unfortunately we can't use that if the placement hasn't been saved.

We could just swap it out for `.present?` but it does mean pulling stuff out of the database.

app/views/placements/schools/placements/build/check_your_answers.html.erb
```diff
@@ -30,10 +30,18 @@
...
+            <% summary_list.with_row do |row| %>
+              <% row.with_key(text: t(".additional_subjects")) %>
+              <% row.with_value(text: @placement.subject_name) %>
+              <% row.with_action(text: t(".change"), href: add_additional_subjects_placements_school_placement_build_index_path(@school, :add_additional_subjects), visually_hidden_text: t(".add_additional_subjects")) %>
+            <% end %>
+          <% end %>
```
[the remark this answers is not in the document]
ollietreend: We also talked about `#additional_subject_names` being implemented using `pluck` instead:

```ruby
additional_subjects.order(:name).pluck(:name).to_sentence
```

app/views/placements/schools/placements/build/check_your_answers.html.erb
ollietreend: We talked about how this saves us from doing more data massaging in the controller. Since we don't explicitly clear the `additional_subject_ids` from the session if the user changes the `subject_id` from MFL to something else.

So it seems fair enough to leave this as is for now. It's the simplest way to catch that edge case without adding extra stuff into the controller.

ollietreend (approved):
Nice work! ❤️

PR #694, opened 2024-05-30
Update summary lists throught the application

Kizr:
## Context

When `actions` is not specified on a `govuk_summary_list` component it reserves a space for them even if they're not used. This change means that when actions are not present the component will use the full width available for the data.

- [x] Canvasses the entire apps usage of `govuk_summary_list` and adds `actions: false` or sets `actions` conditionally based on wheter or not actions are used.

## Guidance to review

Check out the changes and have a quick run through of pages that use summary lists to confirm that they are still working as expected. Any instances of text being squashed when actons were not present should be resolved.

## Link to Trello card

[Make summary lists display full-width when they have no actions](https://trello.com/c/JFl42Qnf)

## Screenshots

|Before|After|
|------|------|
| image | image |

After:

### DIFF
--- a/app/views/claims/pages/index.html.erb
+++ b/app/views/claims/pages/index.html.erb
@@ -8,6 +8,7 @@
       </p>
 
       <%= govuk_summary_list(
+        actions: false,
         rows: [
           { key: { text: "Service" }, value: { text: "Claims" } },
           { key: { text: "Signed in as" }, value: { text: current_user&.first_name } },
--- a/app/views/claims/schools/_additional_details.html.erb
+++ b/app/views/claims/schools/_additional_details.html.erb
@@ -1,4 +1,4 @@
-<%= govuk_summary_list(html_attributes: { id: "additional-details" }) do |summary_list| %>
+<%= govuk_summary_list(html_attributes: { id: "additional-details" }, actions: false) do |summary_list| %>
   <% summary_list.with_row do |row| %>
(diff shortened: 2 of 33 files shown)


Gemfile
```diff
@@ -70,7 +70,7 @@ gem "omniauth_openid_connect"
 gem "omniauth-rails_csrf_protection"
 
 # Pagination
-gem "pagy", "~> 8.4"
+gem "pagy", "~> 7.0"
```
JamieCleare2525, line 73: I don't think this is correct.

Gemfile.lock
```diff
@@ -390,7 +390,7 @@ GEM
       validate_url
       webfinger (~> 2.0)
     os (1.1.4)
-    pagy (8.4.0)
+    pagy (7.0.11)
```
JamieCleare2525, line 393: See the comment above

ollietreend (approved):
This looks good to me – although could you please squash/fixup the commits so we don't have a "revert Gemfile changes" commit?

FYI @Nitemaeric this affects Claims too. It was a visual bug that was noticed during our end-to-end service run through. No functional changes, just making sure the summary lists use their full row width whenever possible. It seems sensible to apply the fix across the board.

JamieCleare2525 (approved):

Nitemaeric (approved):
Nice catch! Looks good to me.

CatalinVoineag (approved):

PR #703, opened 2024-06-03
Alter code to only assign only a single subject to placements

JamieCleare2525:
## Context

- Code has been changed to only assign 1 subject per placement. Replacing a join table with `subject_id` (aka `subject`)

- Seeds changed to assign a school directly to the placement (no join table)
- Associations to `placement_subject_join` table removed

## Guidance to review

- Test the placements creation flow.

## Link to Trello card

https://trello.com/c/5XCQQMJ2/370-re-model-placement-subjects-part-2


app/models/subject.rb
```diff
@@ -22,8 +22,7 @@ class Subject < ApplicationRecord
   belongs_to :parent_subject, class_name: "Subject", optional: true
 
   has_many :child_subjects, class_name: "Subject", foreign_key: :parent_subject_id, dependent: :destroy
-  has_many :placement_subject_joins, dependent: :restrict_with_exception
-  has_many :placements, through: :placement_subject_joins # TODO: Remove `through: :placement_subject_joins` after data migration
+  has_many :placements
```
ollietreend, line 25: Any reason for dropping the `dependent: :restrict_with_exception`?

It could be handy just to make sure we never accidentally delete a subject – for example, when using the Rails console.

ollietreend (commented):
This is looking good. I'd be keen to see if we can get the review app to work, though.

ollietreend (approved):
Looks good to me 👍🏻

PR #716, opened 2024-06-04
Placements to show filter

Kizr:
## Context

Designs have been updated to support a radio button selection of filters instead of just available placements

- [x] Removed references to available placements
- [x] Added the new Placements to show filter
- [x] Updated the filter form logic to exclude this filter from the selected filters list
- [x] Updated the placement query to support this filter
- [x] Updated the filter form specs to work with these changes

## Guidance to review

- Log in as Patricia
- Make a note of the organisation
- Log out
- Log in as Anne
- Create a placement, assign it to the same organisation that Patricia is in
- Create a second placement assigned to a different organisation
- Log out
- Log in as Patricia
- Navigate to Placements
- Verify that Available placements is selected by default, all placements should be available
- Change the filter to Assigned to me, there should be 1 unavailable placement
- Change the Placements to show filter to all placements, there should be 2 unavailable placements

## Link to Trello card

[Implement the 'Placements to show' search filter](https://trello.com/c/pwv8cvSD)

## Screenshots

image

### DIFF
--- a/app/forms/placements/placements/filter_form.rb
+++ b/app/forms/placements/placements/filter_form.rb
@@ -4,8 +4,8 @@ class Placements::Placements::FilterForm < ApplicationForm
   attribute :school_ids, default: []
   attribute :subject_ids, default: []
   attribute :year_groups, default: []
+  attribute :placements_to_show, default: "available_placements"
   attribute :only_partner_schools, :boolean, default: false
-  attribute :only_available_placements, :boolean, default: false
 
   def initialize(params = {})
     params.each_value { |v| v.compact_blank! if v.is_a?(Array) }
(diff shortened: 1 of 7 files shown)


app/queries/placements/placements_query.rb
```diff
@@ -8,7 +8,7 @@ def call
     scope = partner_school_condition(scope)
     scope = subject_condition(scope)
     scope = establishment_group_condition(scope)
-    scope = only_available_placements_condition(scope)
+    scope = placements_to_show(scope)
```
ollietreend, line 11: Should this method name be suffixed with `_condition` for consistency?

```suggestion
    scope = placements_to_show_condition(scope)
```

app/forms/placements/placements/filter_form.rb
```diff
@@ -14,15 +14,15 @@ def initialize(params = {})
   end
 
   def filters_selected?
-    attributes.values.compact.flatten.any?
+    attributes.except("placements_to_show").values.compact.flatten.any?
   end
 
   def clear_filters_path(search_location: nil)
-    placements_placements_path(search_location:)
+    placements_placements_path(search_location:, filters: { placements_to_show: })
```
ollietreend, line 21: Would the expected user behaviour be that clicking "Clear filters" resets the 'Placements to show' filter to its default value?

Right now it looks like the selected option will be persisted. I'm wondering if it should actually revert to the default.

```suggestion
    placements_placements_path(search_location:, filters: { placements_to_show: "available_placements" })
```

ollietreend (commented):
Nice work implementing this. Just a couple of questions.

app/forms/placements/placements/filter_form.rb
Kizr: This was deliberate, if we don't show the pills or option to clear filters regardless of the state of this selection, I don't think it makes sense to have it change that filter if something else has been added.

ollietreend (approved):

PR #737, opened 2024-06-07
Add normalize to mentor and school_contact

Kizr:
## Context

Whitespace fails validation rules when we could be removing it before it becomes an issue.

- [x] Normalise `Mentor` TRN
- [x] Normalise `Placements::SchoolContact` name and email address
- [x] delete whitespace from teacher API path
- [x] Add specs

## Guidance to review

- Add a mentor
- Add space to TRN
- Verify that TRN is still valid
- Add ITT placement contact
- Add whitespice to both fields
- Verify that it is still valid

## Link to Trello card

[Strip whitespace from user input before validation](https://trello.com/c/i5E3VJ9j/463-strip-whitespace-from-user-input-before-validation)

### DIFF
--- a/app/services/teaching_record/rest_client.rb
+++ b/app/services/teaching_record/rest_client.rb
@@ -13,7 +13,7 @@ module TeachingRecord
     class HttpError < StandardError; end
 
     def self.get(path)
-      response = Request.get("/v3/#{path}")
+      response = Request.get("/v3/#{path.delete(" ")}")
 
       if response.ok?
         JSON.parse(response.body || "{}")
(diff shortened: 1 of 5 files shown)


app/services/teaching_record/rest_client.rb
```diff
@@ -13,7 +13,7 @@ class TeacherNotFoundError < StandardError; end
     class HttpError < StandardError; end
 
     def self.get(path)
-      response = Request.get("/v3/#{path}")
+      response = Request.get("/v3/#{path.delete(" ")}")
```
JamieCleare2525, line 16: Can we use `strip` instead of `delete(" ")`? https://apidock.com/ruby/String/strip§

Kizr: I believe the space was actually in the middle of the path in this instance, given there should be no spaces in a path I opted to remove them all.

JamieCleare2525 (approved):

PR #740, opened 2024-06-07
Update add a placement back button

Kizr:
## Context

The back button was not working as expected during the bug party.

- [x] Added logic to work the previous step

## Guidance to review

- Log in as Anne
- Add a placement
- Fill in the form until you get to check your answers
- Click on the back button until you return to the placements page

## Link to Trello card

[Confirm interaction for the 'back' link during the 'Add placement' journey – and implement it](https://trello.com/c/6ofqSDIW/464-confirm-interaction-for-the-back-link-during-the-add-placement-journey-and-implement-it)

### DIFF
--- a/app/controllers/placements/schools/placements/build_controller.rb
+++ b/app/controllers/placements/schools/placements/build_controller.rb
@@ -142,7 +142,7 @@ class Placements::Schools::Placements::BuildController < ApplicationController
                   flash: { alert: t("errors.internal_server_error.page_title") } and return
     end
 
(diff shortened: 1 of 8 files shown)


ollietreend (commented):
Your implementation looks good. But I think it could be wise to add a system test for this – i.e. to cover the 'backwards' journey – so we don't accidentally regress and send users into an infinite loop again.

What do you think?

ollietreend (approved):

PR #760, opened 2024-06-11
Refactor support user authorisation

ollietreend:
## Context

We noticed a bug where some support console controllers were open to non-support users. This happened because there was an assumption baked in that all support controllers would extend from the parent `Placements::Support::ApplicationController`, which is where the authorisation check was happening.

However the 'partner schools' and 'partner providers' controllers didn't extend from that parent controller. Instead, they extended the respective controllers from the non-support area. Therefore the authorisation check in `Placements::Support::ApplicationController` was not happening.

Given the similarity of functionality between the partnerships controllers for both non-support and support interfaces, it seems reasonable to share the controllers here. But we still need support users to be authorised, too.

I've refactored how & where support user authorisation happens. Authorisation will now happen automatically on every controller within the `Placements::Support` module namespace, making it far less likely for us to 'forget' to add it.

- Introduce `Placements::ApplicationController` which is used by every controller within the Placements namespace.
- Ensure only support users can access support console controllers.
- Remove `Placements::Support::ApplicationController` since it's no longer needed.
- Add test coverage to make sure the bug cannot return.

## Guidance to review

1. Sign in as a support user.
2. Navigate to an organisation's partner schools/providers page. Copy the URL.
3. Sign out, and sign in again as a non-support user.
4. Try going to the support console URL you copied in Step 2.
5. You should not be able to see the page.

## Link to Trello card

https://trello.com/c/M8gi1QEn/479-fix-a-bug-where-non-support-users-can-access-the-partnerships-controllers-in-support-if-they-know-the-url


JamieCleare2525 (approved):
Tested on the review app. All working as expected.

PR #773, opened 2024-06-17
Decompose add mentors

Kizr:
## Context

- [x] Add mentors step
- [x] Use the step rather than the placement for adding mentors
- [x] Remove redundant logic 

## Guidance to review

- Log in as Anne
- Add a placement
- Confirm the mentors are saved as expected

## Link to Trello card

### DIFF
--- a/config/locales/en/activemodel.yml
+++ b/config/locales/en/activemodel.yml
@@ -5,6 +5,11 @@ en:
         record_invalid: "Validation failed: %{errors}"
         between: "must be between %{min} and %{max}"
       models:
+        placements/add_placement/steps/mentors:
+          attributes:
+            mentor_ids:
+              blank: Select a mentor or not yet known
+              invalid: Select a mentor or not yet known
         claims/claim/mentor_training_form:
           attributes:
             custom_hours_completed:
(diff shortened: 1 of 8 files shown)


spec/wizards/placements/add_placement/steps/mentors_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        school = create(:placements_school)
+        mentor = create(:placements_mentor_membership, school:).mentor
+        step = described_class.new(school:)
+
+        step.mentor_ids = [mentor.id]
+
+        expect(step.mentor_ids).to eq([mentor.id])
+      end
+    end
+  end
```
ollietreend, line 58: Do we also want a scenario to cover when the array contains random strings and/or mentor IDs which belong to _a different school_?

The point being, those will be filtered out because we do:

```ruby
school.mentors.where(id: mentor_ids).ids
```

This seems like a valuable behaviour to cover with a test.

config/locales/en/activemodel.yml
```diff
@@ -5,6 +5,11 @@ en:
         record_invalid: "Validation failed: %{errors}"
         between: "must be between %{min} and %{max}"
       models:
+        placements/add_placement/steps/mentors:
+          attributes:
+            mentor_ids:
+              blank: Select a mentor or not yet known
+              invalid: Select a mentor or not yet known
```
ollietreend, line 12: Is `invalid` now redundant? I think we removed that error type in the end.

```suggestion
              blank: Select a mentor or not yet known
```

ollietreend (commented):
I love it! 😍 This is a really good step in the right direction.

Just a few minor points from me.

ollietreend (approved):
😍

PR #774, opened 2024-06-17
Decompose add year group

Kizr:
## Context

- [x] Add year group step
- [x] Use the step rather than the placement for selecting a year group
- [x] Remove redundant logic 

## Guidance to review

- Log in as Anne
- Add a placement
- Confirm the year group has saved as expected

## Link to Trello card

### DIFF
--- a/app/models/placements/school.rb
+++ b/app/models/placements/school.rb
@@ -81,4 +81,8 @@ class Placements::School < School
   has_many :partner_providers,
            through: :partnerships,
            source: :provider
+
+  def primary?
+    phase == "Primary"
+  end
 end
--- a/app/models/placements/school.rb
+++ b/app/models/placements/school.rb
@@ -81,8 +81,4 @@ class Placements::School < School
   has_many :partner_providers,
            through: :partnerships,
            source: :provider
-
-  def primary?
-    phase == "Primary"
-  end
 end
--- a/spec/wizards/placements/add_placement/steps/year_group_spec.rb
+++ b/spec/wizards/placements/add_placement/steps/year_group_spec.rb
@@ -1,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe Placements::AddPlacement::Steps::YearGroup, type: :model do
+  describe "attributes" do
+    it { is_expected.to have_attributes(school: nil, year_group: nil) }
+  end
+
+  describe "delegations" do
+    it { is_expected.to delegate_method(:primary?).to(:school).with_prefix }
+  end
+
+  describe "validations" do
+    it { is_expected.to validate_presence_of(:school) }
+
+    context "when the school is primary" do
+      subject { described_class.new(school:) }
+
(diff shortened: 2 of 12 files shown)


spec/wizards/placements/add_placement/steps/year_group_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  describe "validations" do
+    it { is_expected.to validate_presence_of(:school) }
+
+    context "when the school is primary" do
+      subject { described_class.new(school:) }
+
+      let(:school) { instance_double(Placements::School) }
+
+      before { allow(school).to receive(:primary?).and_return(true) }
```
ollietreend, line 20: Could we use a factory rather than a double here, please? 🙏🏻

I always feel uncomfortable about using doubles because they can end up subtly breaking things if the underlying object changes, and the test double doesn't get updated to match.

```suggestion
      let(:school) { build(:placements_school, phase: "Primary") }
```

It might also be a nice opportunity to add a `:primary` trait to the [schools factory](https://github.com/DFE-Digital/itt-mentor-services/blob/9102ee3579eec8dd7a2184e646a51d76f4f38518/spec/factories/schools.rb).

app/models/placements/school.rb
```diff
@@ -81,4 +81,8 @@ class Placements::School < School
   has_many :partner_providers,
            through: :partnerships,
            source: :provider
+
+  def primary?
+    phase == "Primary"
+  end
```
ollietreend, line 87: I think this would be better off in the parent `School` class rather than the namespaced `Placements::School`. We already have [`primary_or_secondary_only?`](https://github.com/DFE-Digital/itt-mentor-services/blob/9102ee3579eec8dd7a2184e646a51d76f4f38518/app/models/school.rb#L120-L122) defined in the parent, and this `primary?` method feels suitably similar.

We should also use the constant `PRIMARY_PHASE`:

```suggestion
  def primary?
    phase == PRIMARY_PHASE
  end
```

ollietreend (commented):
This is looking good 😍  just a couple of things 🙏🏻

spec/wizards/placements/add_placement/steps/year_group_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    it { is_expected.to delegate_method(:primary?).to(:school).with_prefix }
+  end
+
+  describe "validations" do
+    it { is_expected.to validate_presence_of(:school) }
+
+    context "when the school is primary" do
+      subject { described_class.new(school:) }
+
+      let(:school) { create(:placements_school, :primary) }
```
ollietreend, line 18: Micro-optimisation: it may be possible to use `build` here rather than `create`, to avoid needing to write to the database.

ollietreend (approved):
This looks great! 😍 Thanks for making those changes.

I've left one final minor comment, but it's not a blocker. So I'm approving ✅

PR #780, opened 2024-06-18
Decompose add phase

Kizr:
## Context

- [x] Add phase step
- [x] Use the step rather than the placement for adding phase
- [x] Remove redundant logic 

## Guidance to review

- Log into a school that is not secondary or primary
- Add a placement
- Confirm the phase selection is visible

## Link to Trello card

### DIFF
--- a/app/wizards/placements/add_placement/steps/phase.rb
+++ b/app/wizards/placements/add_placement/steps/phase.rb
@@ -5,9 +5,10 @@ class Placements::AddPlacement::Steps::Phase
   attribute :school
   attribute :phase
 
+  VALID_PHASES = [Placements::School::PRIMARY_PHASE, Placements::School::SECONDARY_PHASE].freeze
+
   validates :school, presence: true
-  validates :phase, presence: true
-  validate :phase_is_valid
+  validates :phase, presence: true, inclusion: { in: VALID_PHASES }
 
   def phases_for_selection
     { primary: "Primary", secondary: "Secondary" }
@@ -15,13 +16,5 @@ class Placements::AddPlacement::Steps::Phase
 
   def wizard_attributes
     { phase: }
-  end
-
-  private
-
-  def phase_is_valid
-    return if [Placements::School::PRIMARY_PHASE, Placements::School::SECONDARY_PHASE].include?(phase)
-
-    errors.add(:phase, :invalid)
   end
 end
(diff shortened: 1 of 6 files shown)


app/wizards/placements/add_placement/steps/phase.rb
```diff
@@ -0,0 +1,27 @@
...
+    { phase: }
+  end
+
+  private
+
+  def phase_is_valid
+    return if [Placements::School::PRIMARY_PHASE, Placements::School::SECONDARY_PHASE].include?(phase)
+
+    errors.add(:phase, :invalid)
+  end
```
Kizr, line 26: Should we move this into a setter similarly to Mentors step?

spec/wizards/placements/add_placement/steps/phase_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+
+      context "and the phase is not primary or secondary" do
+        it "is invalid" do
+          step = described_class.new(school:, phase: "Nursery")
+
+          expect(step).not_to be_valid
+        end
+      end
+    end
```
ollietreend, line 38: Looking at the commit history on this branch, I can see you've already refactored these tests. If you wanted to, you could remove even more 'noise' and repetition from each context by doing this:

```ruby
describe "phase" do
  subject { described_class.new(school: build(:placements_school), phase:) }

  context "when the phase is primary" do
    let(:phase) { "Primary" }

    it { is_expected.to be_valid }
  end

  context "when the phase is secondary" do
    let(:phase) { "Secondary" }

    it { is_expected.to be_valid }
  end

  context "when the phase is neither primary nor secondary" do
    let(:phase) { "Nursery" }

    it { is_expected.to be_invalid }
  end
end
```

But I don't feel strongly about it – what you have already is good!

spec/wizards/placements/add_placement/steps/phase_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  describe "#phases_for_selection" do
+    it "returns primary and secondary phases" do
+      step = described_class.new(school: build(:placements_school))
```
ollietreend, line 43: Nice work for switching from `create` to `build` 😉

ollietreend (commented):
This is looking good 👍🏻 nice work!

Just a couple of opinions, but no blockers. I'm happy to approve if you don't think these are valuable changes.

ollietreend (approved):

PR #781, opened 2024-06-18
Store sessions in the database

ollietreend:
> [!IMPORTANT]  
> **Care should be taken when deploying this change. To reduce the risk of disrupting live user sessions, it should be deployed out of hours.**
>
> Users will be signed out because their existing sessions will be discarded. However the app is currently low-traffic enough that an out of hours deployment seems like a reasonable mitigation.

## Context

Currently we use the cookie session store. This is the default session store which comes out of the box with Rails. The cookie store mostly works fine and is a good default, but there is one major drawback: storage space is limited to ~4 KB.

From experience, it's surprisingly easy to use up that space. Features such as multi-step forms (a.k.a. wizards) will often depend on using the session as temporary storage – and since such forms accept user input, the amount of session storage required will consequently be unpredictable and vary wildly.

If you over-fill session cookies, the application will raise `CookieOverflow` errors. As abundantly tasty as these errors sound, they don't make for a good user experience. Users will randomly start seeing the "something went wrong" page, and the server will give `500 Internal Server Error` responses.

Thankfully [Rails provides other session stores](https://guides.rubyonrails.org/action_controller_overview.html#session) which don't have a hard limit on storage.

I've set up the Active Record session storage adapter. This means user sessions will be stored in the database, and we'll be able to read & write to sessions without worrying about the ~4 KB storage limit.

As [recommended in the documentation](https://github.com/rails/activerecord-session_store), I have also set up a scheduled job which will prune the database of old & expired sessions. However rather than using the `db:sessions:trim` Rake task, it ended up being simpler to re-implement the trim functionality as an Active Job so it can be scheduled via Good Job.

## Guidance to review

- This PR should not result in any user-facing changes.
- Does everything work the same as it did before?

## Link to Trello card

https://trello.com/c/AVwbE2ax/504-use-a-database-backed-session-store

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -61,6 +61,9 @@ gem "redcarpet", "~> 3.6"
 # GoodJob backend for Active Job
 gem "good_job", "~> 3.29"
 
+# Store user sessions in the database
+gem "activerecord-session_store"
+
 gem "govuk-components", "~> 5.4.0"
 gem "govuk_design_system_formbuilder", "~> 5.4"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -91,6 +91,13 @@ GEM
       timeout (>= 0.4.0)
     activerecord-nulldb-adapter (1.0.1)
       activerecord (>= 5.2.0, < 7.2)
+    activerecord-session_store (2.1.0)
+      actionpack (>= 6.1)
+      activerecord (>= 6.1)
+      cgi (>= 0.3.6)
+      multi_json (~> 1.11, >= 1.11.2)
+      rack (>= 2.0.8, < 4)
+      railties (>= 6.1)
     activestorage (7.1.3.4)
       actionpack (= 7.1.3.4)
       activejob (= 7.1.3.4)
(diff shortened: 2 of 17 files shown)


Kizr (approved):

Nitemaeric (approved):
Looks good to me.

1 thing to note with the TrimExpiredSessionsJob is that as we schedule this hourly, the session expiration may last 2 hours and 59 minutes if a user signs in at _just_ the right time.

Not an issue really though.

ollietreend:
Thanks for the quick reviews @Kizr & @Nitemaeric! ⭐ 

> 1 thing to note with the TrimExpiredSessionsJob is that as we schedule this hourly, the session expiration may last 2 hours and 59 minutes if a user signs in at just the right time.

That's technically true, although in practice it won't make a difference because the user authentication [always expires after 2 hours regardless][1]. That is also the case with the current cookie-backed sessions which could technically last for _any_ length of time (it's up to the browser when they get deleted).

This cleanup job is primarily there just to keep the database from filling up with dead sessions. But the frequency of the schedule isn't dictated by the length of the session timeout itself.

[1]: https://github.com/DFE-Digital/itt-mentor-services/blob/60d1fb398098f58f53c3849e9f41b0343714f9e0/app/models/dfe_sign_in_user.rb#L37

PR #784, opened 2024-06-20
Decompose additional subjects

Kizr:
## Context

- [x] Add additional subjects step
- [x] Use the step rather than the placement for adding additional subjects
- [x] Remove redundant logic 

## Guidance to review

- Log into a school that is not secondary or primary
- Add a secondary placement
- Select Modern foreign languages from subject
- Select additional languages

## Link to Trello card


app/wizards/placements/add_placement/steps/additional_subjects.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def parent_subject
+    @parent_subject ||= Subject.find(parent_subject_id)
+  end
+
+  private
+
+  def additional_subjects_valid
+    return if additional_subject_ids.all? { |id| Subject.exists?(id:) }
```
ollietreend, line 35: ### Check that it's a child of the parent subject

I think we could tighten this up a little. We should probably be checking that all `additional_subject_ids` are children of `parent_subject`, rather than just checking they exist.

### Performance tweak

Also, it _might be_ more performant? 🤔 I think it should cut out N queries to the database by avoiding the loop.

```suggestion
return if (additional_subject_ids - parent_subject.child_subjects.ids).blank?
```

If every additional subject is a child of the parent, then we should end up with an empty array. Any outliers will remain. I appreciate the array mathematics might hurt readability, though.

### The `inclusion` validator

In fact, now I think about it, I think the `inclusion` validator should be able to handle it. And then we can drop the custom validation method altogether.

It turns out `inclusion` [accepts a lambda](https://edgeapi.rubyonrails.org/classes/ActiveModel/Validations/HelperMethods.html#method-i-validates_inclusion_of) 🙌🏻 which means it should be possible to do something like this:

```ruby
validates :additional_subject_ids, presence: true,
          inclusion: { in: ->(step) { step.additional_subjects_valid.ids } }
```

spec/wizards/placements/add_placement/steps/additional_subjects_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      let(:school) { create(:placements_school) }
+      let(:parent_subject) { create(:subject) }
+
+      it "is invalid of the subject does not exist" do
+        step = described_class.new(parent_subject_id: parent_subject.id, additional_subject_ids: ["not_a_subject"])
+
+        expect(step).to be_invalid
+      end
+
+      it "is valid if all additional subjects exist" do
```
ollietreend, line 23: > is valid if all additional subjects exist

**_...and are children of the parent subject_**

spec/wizards/placements/add_placement/steps/additional_subjects_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  describe "validations" do
+    it { is_expected.to validate_presence_of(:school) }
+    it { is_expected.to validate_presence_of(:parent_subject_id) }
+    it { is_expected.to validate_presence_of(:additional_subject_ids) }
+
+    describe "#additional_subjects_valid" do
+      let(:school) { create(:placements_school) }
+      let(:parent_subject) { create(:subject) }
+
+      it "is invalid of the subject does not exist" do
```
ollietreend, line 17: Typo:

```suggestion
      it "is invalid if the subject does not exist" do
```

spec/wizards/placements/add_placement/steps/additional_subjects_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+        expect(step).to be_valid
+      end
+    end
+  end
+
+  describe "#additional_subjects_for_selection" do
+    let(:parent_subject) { create(:subject) }
+    let(:child_subject) { create(:subject, parent_subject:) }
+
+    it "returns additional subjects for the subject" do
```
ollietreend, line 36: This might be more descriptive:

```suggestion
    it "returns child subjects of the parent subject" do
```

spec/wizards/placements/add_placement/steps/additional_subjects_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    let(:parent_subject) { create(:subject) }
+
+    it "normalises the value to an array" do
+      step = described_class.new
+
+      step.additional_subject_ids = parent_subject.id
+
+      expect(step.additional_subject_ids).to eq([parent_subject.id])
+    end
+  end
```
ollietreend, line 53: Is this behaviour actually needed? 🤔

I think we should always receive values as an array because the form inputs are checkboxes.

I'm not convinced we need a custom setter method for this attribute.

ollietreend (commented):
Nice work! ⭐ This is generally looking good, but I think one or two things might need to be tweaked.

PR #785, opened 2024-06-20
Decompose add subject

Kizr:
## Context

- [x] Add subject step
- [x] Use the step rather than the placement for adding subject
- [x] Remove redundant logic 

## Guidance to review

- Log in as Anne
- Add a placement
- Select a subject

## Link to Trello card

### DIFF
--- a/app/wizards/placements/add_placement/steps/subject.rb
+++ b/app/wizards/placements/add_placement/steps/subject.rb
@@ -8,10 +8,13 @@ class Placements::AddPlacement::Steps::Subject
 
   validates :school, presence: true
   validates :phase, presence: true
-  validates :subject_id, presence: true
+  validates :subject_id, presence: true, inclusion: { in: ->(step) { step.subjects_for_selection.ids } }, if: ->(step) { step.phase.present? }
 
   def subjects_for_selection
-    phase == "Primary" ? Subject.parent_subjects.primary : Subject.parent_subjects.secondary
+    {
+      "Primary" => Subject.parent_subjects.primary,
+      "Secondary" => Subject.parent_subjects.secondary,
+    }.fetch phase
   end
 
   def wizard_attributes
--- a/app/wizards/placements/add_placement/steps/subject.rb
+++ b/app/wizards/placements/add_placement/steps/subject.rb
@@ -1,0 +1,24 @@
+class Placements::AddPlacement::Steps::Subject
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attribute :school
+  attribute :phase
+  attribute :subject_id
+
+  validates :school, presence: true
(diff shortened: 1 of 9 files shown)


app/wizards/placements/add_placement/steps/subject.rb
```diff
@@ -0,0 +1,24 @@
...
+  attribute :phase
+  attribute :subject_id
+
+  validates :school, presence: true
+  validates :phase, presence: true
+  validates :subject_id, presence: true
+
+  def subjects_for_selection
+    phase == "Primary" ? Subject.parent_subjects.primary : Subject.parent_subjects.secondary
+  end
```
ollietreend, line 15: I'm mostly just thinking out loud with this one. I know we've avoided using `case` statements because of the inevitable code coverage issues it causes (i.e. there's always a `nil` branch in the logic).

But instead, what about this?

```suggestion
  def subjects_for_selection
    {
      "Primary" => Subject.parent_subjects.primary,
      "Secondary" => Subject.parent_subjects.secondary,
    }.fetch phase
  end
```

I think I prefer readability of this compared to using a [ternary operator](https://www.rubyguides.com/2019/10/ruby-ternary-operator/), but I'm wondering if it causes the same coverage issues that we get with `case`? 🤔

That said, we do use this approach elsewhere – for example here:

https://github.com/DFE-Digital/itt-mentor-services/blob/4836a5ff5645357884e31f4072207e251f073fc6/app/helpers/routes_helper.rb#L7-L12

So maybe there isn't a code coverage issue. In a way it feels 'safer' because it would crash if an unrecognised phase was passed in.

What do you think? Not a strong opinion, but I'm wondering if it's worth switching to this?

app/wizards/placements/add_placement/steps/subject.rb
```diff
@@ -0,0 +1,24 @@
...
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attribute :school
+  attribute :phase
+  attribute :subject_id
+
+  validates :school, presence: true
+  validates :phase, presence: true
+  validates :subject_id, presence: true
```
ollietreend, line 11: Interestingly we don't validate that `subject_id` is valid, even though we do validate that `additional_subject_ids` are valid over in #784.

Maybe we should?

```suggestion
  validates :subject_id, presence: true,
            inclusion: { in: ->(step) { step.subjects_for_selection.ids } }
```

ollietreend (commented):
Nice! 🤓 🎉

ollietreend (commented):
Oops – I forgot something. Looking at the commits, it looks like a rebase might have borked the commit history? 🤔 There are commits from `main` in this branch, and I don't think there should be.

PR #805, opened 2024-06-27
Consolidate organisation user controllers

JamieCleare2525:
## Context

- Consolidate Organisation User Controllers

- Delete `app/controllers/placements/support/organisations/users_controller.rb`.
- Consolidate `app/controllers/placements/support/providers/users_controller.rb`, `app/controllers/placements/support/schools/users_controller.rb` to inherit from `app/controllers/placements/organisations/users_controller.rb`/
- Introduce and improve upon policies used to restrict user members being deleted.

## Guidance to review

- Test the onboarding process for Support Users being able to onboard users onto schools and providers.

## Link to Trello card

https://trello.com/c/qSimG74T/525-consolidate-organisation-user-controllers


app/policies/placements/user_membership_policy.rb
```diff
@@ -0,0 +1,13 @@
+class Placements::UserMembershipPolicy < ApplicationPolicy
+  def destroy?
+    return false unless user != record.user
```
ollietreend, line 3: Would this logic be clearer the other way around?

```suggestion
    # Users cannot delete themselves
    return false if user == record.user
```

To me, this feels easier to parse. "Return false if the current user is the `record.user`."

app/policies/placements/user_membership_policy.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    user.support_user? ||
+      user.schools.ids.include?(record.organisation.id) ||
+      user.providers.ids.include?(record.organisation.id)
```
ollietreend, line 7: I think this could be simplified to:

```suggestion
    user.support_user? ||
      user.user_memberships.exists?(organisation_id: record.organisation_id)
```

Rather than trying schools and then falling back to providers, it directly queries the current user's memberships as a whole. That should cut out an unnecessary database query in the case of destroying a provider.

Likewise using `record.organisation_id` in favour of `record.organisation.id` could potentially cut out another query to the database.

### For bonus points: check the organisation type, too

Technically we should really also be checking against the `organisation_type`.

```suggestion
    user.support_user? ||
      user.user_memberships.exists?(
        organisation_id: record.organisation_id,
        organisation_type: record.organisation_type,
      )
```

If our database used numeric IDs, we'd inevitably end up with users in School `1234` being able to remove users from Provider `1234` because the IDs match.

Thankfully that's not a problem for us since we're using UUIDs, so the organisation IDs will in theory always be unique. But it wouldn't hurt to tighten it up anyway – just in case.

spec/policies/placements/user_membership_policy_spec.rb
```diff
@@ -0,0 +1,67 @@
+require "rails_helper"
+
+RSpec.describe Placements::UserMembershipPolicy do
+  subject(:user_membership_policy) { described_class }
+
+  let(:user_membership) { create(:user_membership, user:, organisation:) }
+
+  permissions :destroy?, :remove? do
+    context "when current user and user are the the same user" do
```
ollietreend, line 9:
```suggestion
    context "when current user and user are the same" do
```

Or, potentially easier to understand:

```suggestion
    context "when the current user tries to delete themselves" do
```

spec/policies/placements/user_membership_policy_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+          let(:organisation) { create(:provider, :placements) }
+          let(:current_user_membership) { create(:user_membership, user: current_user, organisation:) }
+
+          before { current_user_membership }
+
+          it "permit access" do
+            expect(user_membership_policy).to permit(current_user, user_membership)
+          end
+        end
+      end
```
ollietreend, line 64: I think the code nesting is a bit odd here. Is the Provider context supposed to be nested inside the School context, or are they supposed to be siblings?

ollietreend (commented):
Really nice work @JamieCleare2525! ⭐ 

This is much tidier, and I like the use of `authorize` and the new Pundit policy.

I've added a few comments which are mostly code-style opinions. I'm happy for you to disagree with any of them. And happy to talk any of them through.

ollietreend (approved):
Thanks for making those tweaks 👍🏻

PR #807, opened 2024-06-28
Redirect users after sign in

Kizr:
## Context

When using a deep link into the application the user would always been redirected to the root path after sign in, we want to send them to their requested page instead.

- [x] Make use of the requested path session value to send the user to the correct place
- [x] Redirect the user to the organisatons controller if a provider cannot be found.

## Guidance to review

- Log in as any user
- Navigate to a page in the application
- Copy the URL
- Log out
- Paste the URL
- Log in
- You should be on the same URL you requested

## Link to Trello card

[After signing in, redirect users back to the page they originally requested](https://trello.com/c/CA5fcgyX/513-after-signing-in-redirect-users-back-to-the-page-they-originally-requested)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -28,11 +28,19 @@ class ApplicationController < ActionController::Base
   end
 
   def after_sign_in_path
-    if current_user.support_user?
+    if requested_path.present?
+      requested_path
+    elsif current_user.support_user?
       support_root_path
     else
       organisations_path
     end
+  end
+
+  def requested_path
+    return if [sign_in_path, sign_out_path, nil].include?(session["requested_path"])
+
+    session[:requested_path]
   end
 
   def after_sign_out_path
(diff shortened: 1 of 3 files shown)


JamieCleare2525 (approved):

ollietreend (approved):
Looks good to me 👍🏻

app/controllers/application_controller.rb
```diff
@@ -28,13 +28,21 @@ def user_signed_in?
...
     else
       organisations_path
     end
   end
 
+  def requested_path
+    return if [sign_in_path, sign_out_path, nil].include?(session["requested_path"])
+
+    session[:requested_path]
+  end
```
ollietreend, line 44: We should be careful to make sure we [always use strings for session keys][1]. Symbols will probably work just fine, but it's a bit weird because different session adapters behave differently. AFAIK every adapter works as expected with strings.

[1]: https://github.com/DFE-Digital/itt-mentor-services/commit/3f37281c7560bdb031b5baf5db73353dd7c3292a

```suggestion
  def requested_path
    return if [sign_in_path, sign_out_path, nil].include?(session["requested_path"])

    session["requested_path"]
  end
```

PR #818, opened 2024-07-02
[Support] Add a placement

Kizr:
## Context

Adding this to the support console better helps support our users

- [x] Add a placement is now available in the support console
- [x] Split the system spec into a shared example to reduce duplicate code
- [x] Added specific exceptions for the shared example 

## Guidance to review

- Log in as Colin
- Select a school
- Add a placement

## Link to Trello card

[Build 'Add a placement' flow for support users](https://trello.com/c/oWIBpTh2/261-build-add-a-placement-flow-for-support-users)


.rubocop.yml
```diff
@@ -49,3 +49,12 @@ RSpec/SpecFilePathSuffix:
 
 RSpec/VerifiedDoubleReference:
   Enabled: false
+
+RSpec/LetSetup:
```
JamieCleare2525, line 53: I'm guessing you need this for the `let!` in spec where you are performing actions before the shared example.

app/controllers/placements/support/schools/placements/add_placement_controller.rb
```diff
@@ -0,0 +1,61 @@
+class Placements::Support::Schools::Placements::AddPlacementController < Placements::ApplicationController
```
JamieCleare2525, line 1:
```suggestion
class Placements::Support::Schools::Placements::AddPlacementController < Placements::Schools::Placements::AddPlacementController
```
Most of this code is identical to what's in the `Placements::Schools::Placements::AddPlacementController`

app/views/placements/support/schools/placements/add_placement/_check_your_answers_step.html.erb
```diff
@@ -0,0 +1,62 @@
...
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-l">
+        <%= t(".add_placement") %>
+      </span>
+
+      <span class="govuk-heading-l">
+        <%= t(".title") %>
+      </span>
+
+      <%= govuk_summary_list do |summary_list| %>
```
JamieCleare2525, line 16: It's probably worth making this a partial. Then any changes to the "check your answers" page of a placements are reflected for both support and school users.

.rubocop.yml
```diff
@@ -49,3 +49,12 @@ RSpec/SpecFilePathSuffix:
...
+  Exclude:
+    - spec/system/placements/support/placements/add_a_placement_spec.rb
+    - spec/system/placements/schools/placements/add_a_placement_spec.rb
+
+RSpec/NoExpectationExample:
+  Exclude:
+    - spec/support/shared_examples/add_a_placement_wizard.rb
```
ollietreend, line 60: I think this is going to be a common problem with shared examples for system specs.

It looks like this rule is already configured to exclude system specs as part of rubocop-govuk:

https://github.com/alphagov/rubocop-govuk/blob/87565131e1e6b15569e311e78ef72260e86f475b/config/rspec.yml#L34-L39

What do you think about us creating a directory specifically for shared examples for system specs? That way we could exclude the entire directory, similar to how it's done for `spec/system` and avoid needing to update the Rubocop config every time we add shared examples for system specs.

For example, we could go with:

```suggestion
RSpec/NoExpectationExample:
  Exclude:
    - spec/support/shared_examples/system/**/*.rb
```

or if that feels too deeply nested, what about:

```suggestion
RSpec/NoExpectationExample:
  Exclude:
    - spec/support/shared_system_examples/**/*.rb
```

What do you think?

.rubocop.yml
ollietreend: You can avoid this by moving the `let!` declarations into the shared examples file. Which should be fine, because they're actually identical between both sets of system specs.

Then we shouldn't need this rule exclusion.

spec/support/shared_examples/add_a_placement_wizard.rb
```diff
@@ -0,0 +1,574 @@
+RSpec.shared_examples "an add a placement wizard" do
```
ollietreend, line 1:
```suggestion
RSpec.shared_examples "an add a placement wizard" do
  let!(:subject_1) { create(:subject, name: "Primary subject", subject_area: :primary) }
  let!(:subject_2) { create(:subject, name: "Secondary subject", subject_area: :secondary) }
  let!(:subject_3) { create(:subject, name: "Secondary subject 2", subject_area: :secondary) }
  let!(:mentor_1) { create(:placements_mentor) }
  let!(:mentor_2) { create(:placements_mentor) }
```

spec/system/placements/support/placements/add_a_placement_spec.rb
```diff
@@ -0,0 +1,50 @@
+require "rails_helper"
+
+RSpec.describe "Placements / Support / Schools / Placements / Add a placement",
+               service: :placements, type: :system do
+  let!(:subject_1) { create(:subject, name: "Primary subject", subject_area: :primary) }
+  let!(:subject_2) { create(:subject, name: "Secondary subject", subject_area: :secondary) }
+  let!(:subject_3) { create(:subject, name: "Secondary subject 2", subject_area: :secondary) }
+  let!(:mentor_1) { create(:placements_mentor) }
+  let!(:mentor_2) { create(:placements_mentor) }
+
```
ollietreend, line 10:
```suggestion
```

app/controllers/placements/support/schools/placements/add_placement_controller.rb
ollietreend: +1 to Jamie's suggestion. I feel like the main (only?) things which differs are:

- `step_path` method
- `back_link_path` method
- and the redirect path in the `update` method, which could be extracted out to another overridable method

app/views/placements/support/schools/placements/add_placement/_additional_subjects_step.html.erb
ollietreend, line 1: It's a shame we've had to duplicate all these step partials 😕

I wonder if we could/should relocate the step partials – for example:

```
views/placements/schools/placements/add_placement/_subject_step.html.erb → 
views/wizards/placements/add_placement_wizard/_subject_step.html.erb
```

That would require a change to the `BaseStep#to_partial_path` method – effectively coupling the step partials with their step classes, rather than coupling them to the specific controller that implements them. Perhaps that makes more sense for our needs though.

What do you think?

ollietreend: Or maybe we're just being too 'clever' by trying to render those partials using `#to_partial_path`?

We could always update the relevant `edit.html.erb` with the correct path to the partial instead. 🤔

app/views/placements/support/schools/placements/add_placement/_check_your_answers_step.html.erb
ollietreend: A partial specifically for this `govuk_summary_list` may not be necessary if we can reuse _all_ the step partials wholesale.

ollietreend (commented):
Yay for code reuse! 🙌🏻

app/controllers/placements/schools/placements/add_placement_controller.rb
```diff
@@ -23,12 +23,16 @@ def update
...
-      redirect_to placements_school_placements_path(@school), flash: { success: t(".success") }
+      redirect_to update_path(@school), flash: { success: t(".success") }
     end
   end
 
   private
 
+  def update_path(school)
+    placements_school_placements_path(school)
+  end
```
ollietreend, line 34: Minor nitpick about the naming of this method. I'm not sure `update_path` makes sense in this context, because it's actually pointing to the placements index route rather than an 'update' route.

Would it be more intention revealing if it was called something like `after_create_path`?

That would follow the example of existing methods like `after_sign_in_path` and `after_sign_out_path`:

https://github.com/DFE-Digital/itt-mentor-services/blob/de9aa3de7e51481425368ccdcf999ba82126af51/app/controllers/application_controller.rb#L30-L38

https://github.com/DFE-Digital/itt-mentor-services/blob/de9aa3de7e51481425368ccdcf999ba82126af51/app/controllers/application_controller.rb#L46-L48

ollietreend (commented):
Lovely stuff! 😍

Just a teeny-tiny minor nitpick from me. But it's not a blocker – feel free to disagree.

app/controllers/placements/schools/placements/add_placement_controller.rb
```diff
@@ -58,4 +62,8 @@ def back_link_path
       placements_school_placements_path(@school)
     end
   end
+
+  def new_add_placement_mentor_path
```
ollietreend, line 66: Would this be a simpler name for it? Since it's local to this controller, it's already within the context of the "add placement" flow.

```suggestion
  def add_mentor_path
```

ollietreend (approved):
Lovely work! 🫶 ✅

I have left one more naming nitpick, but I'm approving regardless because I don't want to block this. I'm just a pedant but what you've done is already excellent.

JamieCleare2525 (approved):

PR #819, opened 2024-07-02
Jc/consolidate partnership controller code

JamieCleare2525:
## Context

- Move common/identical code in partnerships to a `Placements::PartnershipsController`

## Link to Trello card

https://trello.com/c/UHi9kwlA/533-consolidate-placements-partnerships-controllers

### DIFF
--- a/app/controllers/placements/schools/partner_providers_controller.rb
+++ b/app/controllers/placements/schools/partner_providers_controller.rb
@@ -59,11 +59,7 @@ class Placements::Schools::PartnerProvidersController < Placements::Partnerships
   end
 
   def partner_provider
-    @partner_provider ||= if @partnership.present?
-                            @partnership.provider
-                          else
-                            partnership_form.provider.decorate
-                          end
+    @partner_provider ||= partnership_form.provider.decorate
   end
   alias_method :partner_organisation, :partner_provider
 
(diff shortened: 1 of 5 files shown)


Kizr (approved):

app/controllers/placements/schools/partner_providers_controller.rb
```diff
@@ -1,88 +1,41 @@
...
 
-  def set_school
+  def set_organisation
     @school = current_user.schools.find(params.fetch(:school_id))
   end
 
-  def set_decorated_partner_provider
+  def source_organisation
+    @school
+  end
```
ollietreend, line 36: I like the distinction of "source organisation" and "partner organisation".

I'm just wondering if this particular bit of code would get any simpler if we named the instance variable `@organisation`, instead of calling it either `@school` or `@provider`? That way the source organisation will always be in a known variable name.

ollietreend (approved):
Overall I think this looks good! ✨🧹 

It definitely makes sense to consolidate this code because I expect 'partner schools' and 'partner providers' will more than likely always be 'mirror image' implementations of each other.

Literally only one small comment, but it's really no big deal. There's probably a reason why you've done it the way you have.

PR #820, opened 2024-07-03
De-duplicate wizard step partials

Kizr:
## Context

We are duplicating views for wizards when the journey remains the same, we feel that these views/steps are coupled to the logic of the wizard and should be treated as such

- [x] Move step views into the new wizards folder
- [x] Consolidate existing views and tranlsations into this new location
- [x] Update base step path to point to new location

## Guidance to review

- Add a placement as Colin & Anne, experience should not change.

## Link to Trello card

[Deduplicate wizard step partials](https://trello.com/c/uO73zQd4/542-deduplicate-wizard-step-partials)


app/wizards/placements/base_step.rb
```diff
@@ -10,6 +10,16 @@ def initialize(wizard:, attributes:)
...
+  end
+
+  private
+
+  def base_path
+    "placements/wizards/#{class_to_path(wizard.class)}/"
+  end
+
+  def class_to_path(klass)
+    klass.name.demodulize.underscore
```
ollietreend, line 23:
```
def to_partial_path
  "wizards/#{self.class.name.underscore}"
end
```

Instead of:
- app/views/placements/wizards/add_placement_wizard/mentors_step

How about:
- app/views/wizards/placements/add_placement_wizard/mentors_step

Would make the implementation simpler, and maybe more convention-y?

config/locales/en/placements/wizards/add_placement_wizard.yml
ollietreend, line 1: Nitpick: this file is missing a newline character at the end.

But we've changed the settings in your IDE now, so it'll probably be added automatically if you just re-save this file.

ollietreend (approved):

spec/support/shared_examples/system/add_a_placement_wizard.rb
Nitemaeric, line 1: This file isn't nested under a `placements` directory, so GitHub thinks it's shared for us too.

Nitemaeric (approved):

PR #837, opened 2024-07-10
Add settings page to reset claims data

Nitemaeric:
## Context

We regularly need to reset the claims database to perform UR and testing.

- Add a settings page that appears only in non-production environments to reset the Claims database.

## Guidance to review

- Sign in as Anne
- Create a claim
- Sign out and in as Colin (Support User)
- Navigate to the Settings page
- Click on "Reset Database"
- Confirm
- All claims should now have been deleted

### DIFF
--- a/app/controllers/claims/support/databases_controller.rb
+++ b/app/controllers/claims/support/databases_controller.rb
@@ -1,0 +1,9 @@
+class Claims::Support::DatabasesController < Claims::Support::ApplicationController
+  before_action :skip_authorization
+
+  def destroy
+    Claims::ResetDatabase.call
+
+    redirect_to claims_support_settings_path, flash: { success: t(".success") }
+  end
+end
--- a/app/views/claims/support/databases/reset.en.html.erb
+++ b/app/views/claims/support/databases/reset.en.html.erb
@@ -1,0 +1,32 @@
+<% content_for :page_title, "Reset database" %>
+<% render "claims/support/primary_navigation", current: :settings %>
+
+<%= content_for(:before_content) do %>
+  <%= govuk_back_link(href: claims_support_settings_path) %>
+<% end %>
+
+<div class="govuk-width-container">
+  <span class="govuk-caption-l">Reset database</span>
+  <h1 class="govuk-heading-l">Are you sure you want to reset the <%= govuk_tag text: hosting_environment_phase(current_service), colour: hosting_environment_color, classes: "govuk-tag--middle" %> database?</h1>
+
+  <p class="govuk-body">This action will truncate the following tables:</p>
+
+  <%= govuk_list type: :bullet do %>
+    <%= tag.li tag.code("claims", class: "inline-code") %>
+    <%= tag.li tag.code("mentor_memberships WHERE mentor_memberships.type = 'Claims::MentorMembership'", class: "inline-code") %>
+  <% end %>
+
+  <p class="govuk-body">It will also perform the following data updates:</p>
+
+  <%= govuk_list type: :bullet, spaced: true do %>
+    <%= tag.li do %>
+      <div>Reset all grant conditions acceptances for all schools</div>
+      <div><code class="inline-code">schools.claims_grant_conditions_accepted_at = null</code></div>
+      <div><code class="inline-code">schools.claims_grant_conditions_accepted_by_id = null</code></div>
+    <% end %>
+  <% end %>
+
+  <%= govuk_button_to "Reset database", claims_support_database_path, warning: true, method: :delete, class: "govuk-button--warning" %>
+
+  <%= govuk_link_to "Cancel", claims_support_settings_path, no_visited_state: true %>
+</div>
--- a/app/assets/stylesheets/components/_all.scss
+++ b/app/assets/stylesheets/components/_all.scss
@@ -3,6 +3,7 @@
 @import "content_header";
 @import "header";
 @import "heading";
+@import "inline_code";
 @import "link";
 @import "organisation_list_item";
 @import "phase_banner";
@@ -10,3 +11,4 @@
 @import "secondary_navigation";
 @import "search_results";
 @import "typography";
+@import "tag";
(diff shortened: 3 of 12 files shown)


RMcVelia:
Hi, I deleted the worker for this PR as it was using the broken goodjob and logging excessively. Could you rebase to pick up the fixes when possible?

ollietreend (commented):
@Nitemaeric I appreciate I've just reviewed this while it's still in draft. However since you're proposing adding a new View Component to the shared namespace, I thought I'd give some feedback.

I'm happy to talk it through if you'd like.

app/components/inline_code_component.html.erb
```diff
@@ -0,0 +1 @@
+<span class="inline-code"><%= @snippet %></span>
```
ollietreend, line 1: Would this be more semantic as a [`<code>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code) tag? Given the CSS styling and class name, that seems to be the most appropriate HTML tag to use.

Whereas [`<span>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span) has no semantic value.

spec/components/inline_code_component_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe InlineCodeComponent, type: :component do
+  pending "add some examples to (or delete) #{__FILE__}"
```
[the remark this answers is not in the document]
Nitemaeric: I agree. The View Component isn't required.

Nitemaeric:
@ollietreend Great feedback. Thanks! I'll remove the view component. There's no need for it indeed.

app/views/claims/support/databases/remove.en.html.erb
```diff
@@ -0,0 +1,27 @@
...
+
+  <p class="govuk-body">It will also perform the following data updates:</p>
+
+  <%= govuk_list type: :bullet, spaced: true do %>
+    <%= tag.li do %>
+      <div>Reset all grant conditions acceptances for all schools</div>
+      <div><%= inline_code_tag "schools.claims_grant_conditions_accepted_at = null" %></div>
+      <div><%= inline_code_tag "schools.claims_grant_conditions_accepted_by_id = null" %></div>
+    <% end %>
+  <% end %>
```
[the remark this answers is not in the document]
ollietreend: That makes sense. I guess I was interpreting the `<div>` tags as if they were substitutes for `<li>` tags, but without semantic value.

app/controllers/claims/support/databases_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Claims::Support::DatabasesController < Claims::Support::ApplicationController
+  before_action :skip_authorization
+
+  def destroy
+    Claims::ResetDatabase.call
+
+    redirect_to claims_support_settings_path, flash: { success: "Database has been reset" }
```
JamieCleare2525, line 7: Text should be in locales

app/views/claims/support/databases/reset.en.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      <div>Reset all grant conditions acceptances for all schools</div>
+      <div><code class="inline-code">schools.claims_grant_conditions_accepted_at = null"</code></div>
+      <div><code class="inline-code">schools.claims_grant_conditions_accepted_by_id = null"</code></div>
+    <% end %>
+  <% end %>
+
+  <%= govuk_button_to "Reset database", claims_support_database_path, warning: true, method: :delete, class: "govuk-button govuk-button--warning" %>
+
+  <%= govuk_link_to "Cancel", claims_support_settings_path, no_visited_state: true %>
+</div>
```
JamieCleare2525, line 32: The text in this file should be in locales

app/views/claims/support/databases/reset.en.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    <%= tag.li tag.code("claims", class: "inline-code") %>
+    <%= tag.li tag.code("mentor_memberships WHERE mentor_memberships.type = 'Claims::MentorMembership'", class: "inline-code") %>
+  <% end %>
+
+  <p class="govuk-body">It will also perform the following data updates:</p>
+
+  <%= govuk_list type: :bullet, spaced: true do %>
+    <%= tag.li do %>
+      <div>Reset all grant conditions acceptances for all schools</div>
+      <div><code class="inline-code">schools.claims_grant_conditions_accepted_at = null"</code></div>
```
JamieCleare2525, line 24:
```suggestion
      <div><code class="inline-code">schools.claims_grant_conditions_accepted_at = null</code></div>
```

app/views/claims/support/databases/reset.en.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      <div><code class="inline-code">schools.claims_grant_conditions_accepted_by_id = null"</code></div>
```
JamieCleare2525, line 25:
```suggestion
      <div><code class="inline-code">schools.claims_grant_conditions_accepted_by_id = null</code></div>
```

app/views/claims/support/databases/reset.en.html.erb
Nitemaeric: This is a localised view - namely because it's a static behind-the-scenes page that wouldn't be (support) user-facing, ever. It's for development and pre-production environments only.

Nitemaeric: I've avoided moving the text of this file into a locale file, as it is a localised view, signified with the file extension - `reset.en.html.erb`. If we need to add in a non-`en` locale for this page (Which we shouldn't - as this page is for DfE development/product team only - not Support, not end-users).

Side note, I think the only possibility of adding an additional language falls on the end-user's side. That possibly being Welsh - if the system expands to Wales.

ollietreend (approved):

JamieCleare2525 (approved):

PR #840, opened 2024-07-11
Prepare the database for Good Job v4

ollietreend:
## Context

We need to perform some database migrations to prepare for the upgrade from Good Job v3 to v4. This will allow us to re-apply the upgrade in #828, which was reverted by #838.

> [!NOTE]  
> **This PR does not upgrade Good Job to v4.** It prepares the database for v4, but these changes are still compatible with v3.99.1 which is currently in use.

I've followed the instructions in [the upgrade guide][1]. The database migrations were automatically generated by the gem.

I've also updated the DfE Analytics blocklist to exclude the newly added fields.

## Guidance to review

Run the migrations in this pull request, then open a Rails console and run:

```
GoodJob.v4_ready?
```

It should return `true`.

## Link to Trello card

https://trello.com/c/vneRTcpO/554-review-and-merge-dependabot-prs

[1]: https://github.com/bensheldon/good_job?tab=readme-ov-file#upgrading-v3-to-v4

### DIFF
--- a/config/analytics_blocklist.yml
+++ b/config/analytics_blocklist.yml
@@ -33,6 +33,7 @@ shared:
     - created_at
     - updated_at
     - state
+    - lock_type
   :good_job_executions:
     - id
     - created_at
@@ -45,6 +46,9 @@ shared:
     - finished_at
     - error
     - error_event
+    - error_backtrace
+    - process_id
+    - duration
   :good_job_batches:
     - id
     - created_at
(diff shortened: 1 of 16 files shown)


Nitemaeric (approved):
🚀

Nitemaeric:
Note:

```ruby
GoodJob.v4_ready?
```

is a bit sneaky. If there are no enqueued jobs, it will always return `true`.

ollietreend:
> If there are no enqueued jobs, it will always return `true`.

🤦🏻‍♂️

PR #854, opened 2024-07-15
Create claim windows and academic years

Nitemaeric:
## Context

1. We want to be able to open and close claim windows from the support console.
2. We want to be able to schedule open and close dates ahead of time and avoid support users from needing to manually toggle the windows. (Prevent someone from needed to switch the window on/off at midnight)
3. We want to be able to associate a `Claim` with a specific `ClaimWindow`.
4. Furthermore, we want to associate a `Claim` with a specific `AcademicYear`.

- Add `AcademicYear` model.
- Add `Claims::ClaimWindow` model.
- Add CRUD operations for `Claims::ClaimWindow` within the support console.

## Guidance to review

- [Understanding academic years](https://www.gov.uk/guidance/understanding-academic-years)
- There is an alternative of storing academic years of claim windows as enum strings, such as `"2023 to 2024"` and `"2024 to 2025"` and so on, but this _feels_ prone to potential error. Happy to be persuaded otherwise.
- This PR is based on developer-driven design. There is no prototype to reference.
  - Design feedback has been provided.
- Many unit tests are missing. This will be addressed next week with this Trello card: https://trello.com/c/ZBYQ1tNG/575-add-unit-tests-for-the-claim-window-code
  - This is due to deadlines. We need this code to go live today (18 July 2024), as it will take effect tomorrow.

### ER Diagram

CleanShot 2024-07-15 at 23 49 37

## Link to Trello card

https://trello.com/c/xMBtq6Q5/563-add-claim-windows

## Screenshots

\* Some screenshots are slightly outdated or missing from this list altogether, such as the "check" pages.

| Path | Screenshot |
| ---- | ---------- |
| `/support/claim_windows` | CleanShot 2024-07-15 at 23 32 30 |
| `/support/claim_windows/new` | CleanShot 2024-07-15 at 23 33 11 |
| `/support/claim_windows/:id` | CleanShot 2024-07-15 at 23 33 45 |
| `/support/claim_windows/:id/edit` | CleanShot 2024-07-15 at 23 34 13 |
| `/support/claim_windows/:id/remove` | CleanShot 2024-07-15 at 23 34 44 |
| `/support/claim_windows/:id/remove` (current window) | CleanShot 2024-07-15 at 23 35 18 |

### DIFF
--- a/app/controllers/claims/support/claim_windows_controller.rb
+++ b/app/controllers/claims/support/claim_windows_controller.rb
@@ -1,0 +1,72 @@
+class Claims::Support::ClaimWindowsController < Claims::Support::ApplicationController
+  before_action :set_claim_window, only: %i[show edit edit_check update remove destroy]
+  before_action :authorize_claim_window
+
+  def index
+    @claim_windows = Claims::ClaimWindow.order(starts_on: :desc)
+  end
+
+  def new
+    @claim_window = if params[:claims_claim_window]
+                      Claims::ClaimWindow.new(claim_window_params)
+                    else
+                      Claims::ClaimWindow.new
+                    end
+  end
+
+  def new_check
+    @claim_window = Claims::ClaimWindow::Build.call(claim_window_params:)
+
+    render :new unless @claim_window.valid?
+  end
+
+  def create
+    @claim_window = Claims::ClaimWindow::Build.call(claim_window_params:)
+    @claim_window.save!
+
+    redirect_to claims_support_claim_windows_path, flash: { success: "Claim window created" }
+  end
+
+  def show; end
+
+  def edit
+    if params[:claims_claim_window]
+      @claim_window.assign_attributes(claim_window_params)
+    end
+  end
+
+  def edit_check
+    @claim_window = Claims::ClaimWindow::Build.call(claim_window: @claim_window, claim_window_params:)
+
+    render :edit unless @claim_window.valid?
+  end
+
+  def update
+    @claim_window = Claims::ClaimWindow::Build.call(claim_window: @claim_window, claim_window_params:)
+    @claim_window.save!
+
+    redirect_to claims_support_claim_window_path(@claim_window), flash: { success: "Claim window updated" }
+  end
+
+  def remove; end
+
+  def destroy
+    @claim_window.discard!
+
+    redirect_to claims_support_claim_windows_path, flash: { success: "Claim window removed" }
+  end
+
+  private
+
+  def claim_window_params
+    params.require(:claims_claim_window).permit(:starts_on, :ends_on)
+  end
+
+  def set_claim_window
+    @claim_window = Claims::ClaimWindow.find(params.require(:id))
+  end
+
+  def authorize_claim_window
+    authorize @claim_window || Claims::ClaimWindow
+  end
+end
--- a/app/models/academic_year.rb
+++ b/app/models/academic_year.rb
@@ -1,0 +1,13 @@
+# == Schema Information
+#
+# Table name: academic_years
+#
+#  id         :uuid             not null, primary key
+#  starts_on  :date
+#  ends_on    :date
+#  name       :string
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+class AcademicYear < ApplicationRecord
+end
--- a/app/models/claims/claim_window.rb
+++ b/app/models/claims/claim_window.rb
@@ -1,0 +1,53 @@
+# == Schema Information
+#
+# Table name: claim_windows
+#
+#  id               :uuid             not null, primary key
+#  discarded_at     :date
+#  ends_on          :date
+#  starts_on        :date
+#  created_at       :datetime         not null
+#  updated_at       :datetime         not null
+#  academic_year_id :uuid             not null
+#
+# Indexes
+#
+#  index_claim_windows_on_academic_year_id  (academic_year_id)
+#  index_claim_windows_on_discarded_at      (discarded_at)
+#
+# Foreign Keys
+#
+#  fk_rails_...  (academic_year_id => academic_years.id)
+#
+class Claims::ClaimWindow < ApplicationRecord
+  include Discard::Model
+
+  default_scope -> { kept }
+
+  belongs_to :academic_year
+
+  validates :starts_on, presence: true
+  validates :ends_on, presence: true, comparison: { greater_than_or_equal_to: :starts_on }
+
+  validate :does_not_overlap_with_another_claim_window
+
+  delegate :name, to: :academic_year, prefix: true
+
+  def self.find_by_date(date)
+    find_by(starts_on: ...date, ends_on: date..)
+  end
+
+  def self.current
+    find_by_date(Date.current)
+  end
+
+  private
+
+  def does_not_overlap_with_another_claim_window
+    overlapping_claim_window = Claims::ClaimWindow.where.not(id:).find_by("starts_on <= ? AND ends_on >= ?", ends_on, starts_on)
+
+    return if overlapping_claim_window.blank?
+
+    errors.add(:base, "Claim window overlaps with an existing claim window - #{I18n.l(overlapping_claim_window.starts_on, format: :long)} to #{I18n.l(overlapping_claim_window.ends_on, format: :long)}")
+  end
+end
--- a/app/models/claims/claim_window.rb
+++ b/app/models/claims/claim_window.rb
@@ -29,25 +29,47 @@ class Claims::ClaimWindow < ApplicationRecord
   validates :starts_on, presence: true
   validates :ends_on, presence: true, comparison: { greater_than_or_equal_to: :starts_on }
 
-  validate :does_not_overlap_with_another_claim_window
+  validate :is_not_longer_than_an_academic_year
+  validate :does_not_start_in_the_past, on: :create
+  validate :does_not_start_within_another_claim_window
+  validate :does_not_end_in_the_past, on: :create
+  validate :does_not_end_within_another_claim_window
 
   delegate :name, to: :academic_year, prefix: true
+  delegate :past?, to: :ends_on
+  delegate :future?, to: :starts_on
 
-  def self.find_by_date(date)
-    find_by(starts_on: ...date, ends_on: date..)
-  end
-
-  def self.current
-    find_by_date(Date.current)
+  def current?
+    (starts_on..ends_on).cover?(Date.current)
   end
 
   private
 
-  def does_not_overlap_with_another_claim_window
-    overlapping_claim_window = Claims::ClaimWindow.where.not(id:).find_by("starts_on <= ? AND ends_on >= ?", ends_on, starts_on)
+  def is_not_longer_than_an_academic_year
+    return if starts_on.blank? || ends_on.blank?
 
-    return if overlapping_claim_window.blank?
+    if (ends_on - starts_on).to_i > 365
+      errors.add(:base, :longer_than_academic_year)
+    end
+  end
 
-    errors.add(:base, "Claim window overlaps with an existing claim window - #{I18n.l(overlapping_claim_window.starts_on, format: :long)} to #{I18n.l(overlapping_claim_window.ends_on, format: :long)}")
+  def does_not_start_within_another_claim_window
+    return unless Claims::ClaimWindow.where.not(id:).where("starts_on <= :starts_on AND ends_on >= :starts_on", starts_on:).exists?
+
+    errors.add(:starts_on, :overlap)
+  end
+
+  def does_not_end_within_another_claim_window
+    return unless Claims::ClaimWindow.where.not(id:).where("starts_on <= :ends_on AND ends_on >= :ends_on", ends_on:).exists?
+
+    errors.add(:ends_on, :overlap)
+  end
+
+  def does_not_start_in_the_past
+    errors.add(:starts_on, :past) if starts_on&.past?
+  end
+
+  def does_not_end_in_the_past
+    errors.add(:ends_on, :past) if ends_on&.past?
   end
 end
--- a/app/services/claims/claim_window/build.rb
+++ b/app/services/claims/claim_window/build.rb
@@ -1,0 +1,30 @@
+class Claims::ClaimWindow::Build
+  include ServicePattern
+
+  def initialize(claim_window_params:, claim_window: nil)
+    @claim_window = claim_window
+    @claim_window_params = claim_window_params
+  end
+
+  def call
+    self.claim_window ||= Claims::ClaimWindow.new
+    self.claim_window.assign_attributes(claim_window_params)
+    self.claim_window.academic_year = find_academic_year
+    self.claim_window
+  end
+
+  private
+
+  attr_accessor :claim_window, :claim_window_params
+
+  def find_academic_year
+    return unless claim_window.starts_on
+
+    year = claim_window.starts_on.month > 1 ? claim_window.starts_on.year - 1 : claim_window.starts_on.year - 2
+
+    AcademicYear.create_with(
+      starts_on: Date.new(year, 9, 1),
+      ends_on: Date.new(year + 1, 8, 31),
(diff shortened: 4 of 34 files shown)


app/controllers/claims/support/claim_windows_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    @claim_window = Claims::ClaimWindow::Build.call(claim_window_params:)
+
+    render :new unless @claim_window.valid?
+  end
+
+  def create
+    @claim_window = Claims::ClaimWindow::Build.call(claim_window_params:)
+    @claim_window.save!
+
+    redirect_to claims_support_claim_windows_path, flash: { success: "Claim window created" }
```
JamieCleare2525, line 27: This is the first use of the word "created" in our app. Should this not be "Claim window added"?
I would also suggest this go into `config/locales` rather than hard coded text.

app/controllers/claims/support/claim_windows_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+    @claim_window = Claims::ClaimWindow::Build.call(claim_window: @claim_window, claim_window_params:)
+
+    render :edit unless @claim_window.valid?
+  end
+
+  def update
+    @claim_window = Claims::ClaimWindow::Build.call(claim_window: @claim_window, claim_window_params:)
+    @claim_window.save!
+
+    redirect_to claims_support_claim_window_path(@claim_window), flash: { success: "Claim window updated" }
```
JamieCleare2525, line 48: I would also suggest this go into `config/locales` rather than hard coded text.

app/controllers/claims/support/claim_windows_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def remove; end
+
+  def destroy
+    @claim_window.discard!
+
+    redirect_to claims_support_claim_windows_path, flash: { success: "Claim window removed" }
```
JamieCleare2525, line 56: I would also suggest this go into `config/locales` rather than hard coded text.

app/models/academic_year.rb
```diff
@@ -0,0 +1,13 @@
+# == Schema Information
+#
+# Table name: academic_years
+#
+#  id         :uuid             not null, primary key
+#  starts_on  :date
+#  ends_on    :date
+#  name       :string
```
JamieCleare2525, line 8: No validation for these fields being present. Or for `ends_on` being greater than `starts_on`.

Personal preference (feel free to ignore): `start_at` or `start_date` over `starts_on`

app/models/claims/claim_window.rb
```diff
@@ -0,0 +1,75 @@
...
+# Foreign Keys
+#
+#  fk_rails_...  (academic_year_id => academic_years.id)
+#
+class Claims::ClaimWindow < ApplicationRecord
+  include Discard::Model
+
+  default_scope -> { kept }
+
+  belongs_to :academic_year, optional: true
```
JamieCleare2525, line 27: This doesn't seem like it should be `optional: true`

app/models/claims/claim_window.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  validates :starts_on, presence: true
+  validates :ends_on, presence: true
```
JamieCleare2525, line 30: Should included comparison validation:
validates :ends_on, comparison: { less_than: :starts_on }

app/services/claims/claim_window/build.rb
```diff
@@ -0,0 +1,30 @@
...
+    self.claim_window.academic_year = find_academic_year
+    self.claim_window
+  end
+
+  private
+
+  attr_accessor :claim_window, :claim_window_params
+
+  def find_academic_year
+    return unless claim_window.starts_on
```
JamieCleare2525, line 21: Not sure this is necessary. As ClaimWindow validates that a `starts_on` is present.

app/controllers/claims/support/claim_windows_controller.rb
Nitemaeric: Good shout. This is a quick change. I'll address it now.

app/views/claims/support/claim_windows/index.html.erb
```diff
@@ -0,0 +1,35 @@
...
+
+<%= content_for(:before_content) do %>
+  <%= govuk_back_link(href: claims_support_settings_path) %>
+<% end %>
+
+<div class="govuk-width-container">
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-l"><%= t(".page_title") %></h1>
+      <%= govuk_button_to t(".add_claim_window"), new_claims_support_claim_window_path, method: :get %>
```
JamieCleare2525, line 12:
```suggestion
      <%= govuk_button_link_to t(".add_claim_window"), new_claims_support_claim_window_path %>
```

app/models/academic_year.rb
Nitemaeric: I've adopted a convention over time of `_at` being a `timestamp` attribute and `_on` being a `date` attribute. This is a common convention, e.g. https://github.com/thoughtbot/guides/tree/main/rails

app/services/claims/claim_window/build.rb
Nitemaeric: This is necessary, as the `Build#call` does not validate the `ClaimWindow` within it's method scope.

Alternatively, we _could_ do:

```ruby
claim_window = ...
claim_window.academic_year = find_academic_year if claim_window.valid?
```

but that's pulling in external dependencies to the `find_academic_year`. The academic year only depends on a `claim_window.starts_on` date being present.

db/migrate/20240712094833_create_academic_years.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    up_only do
+      [2023, 2024, 2025].each do |year|
+        AcademicYearForMigration.create!(
+          name: "#{year} to #{year + 1}",
+          starts_on: Date.new(year, 8, 1),
+          ends_on: Date.new(year + 1, 7, 31),
+        )
+      end
+    end
```
JamieCleare2525, line 23: This belongs in a data migration.

db/migrate/20240712103537_create_claims_claim_windows.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    up_only do
+      academic_year_date = Date.parse("12 July 2024")
+
+      ClaimWindowForMigration.create!(
+        academic_year_id: AcademicYearForMigration.find_by(starts_on: ...academic_year_date, ends_on: academic_year_date..).id,
+        starts_on: Date.parse("2 May 2024"),
+        ends_on: Date.parse("19 July 2024"),
+      )
+    end
```
JamieCleare2525, line 28: This belongs in a data migration.

db/migrate/20240712094833_create_academic_years.rb
Nitemaeric: Does it need to be? This is a safe migration.

db/migrate/20240712103537_create_claims_claim_windows.rb
Nitemaeric: This too. It is a safe migration.

db/migrate/20240712094833_create_academic_years.rb
```diff
@@ -0,0 +1,25 @@
+class CreateAcademicYears < ActiveRecord::Migration[7.1]
+  class AcademicYearForMigration < ApplicationRecord
+    self.table_name = :academic_years
+  end
+
+  def change
+    create_table :academic_years, id: :uuid do |t|
+      t.string :name
+      t.date :starts_on
+      t.date :ends_on
```
JamieCleare2525, line 10: These attributes should likely be `null: false`

db/migrate/20240712094833_create_academic_years.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      t.timestamps
+    end
+
+    up_only do
+      [2023, 2024, 2025].each do |year|
+        AcademicYearForMigration.create!(
+          name: "#{year} to #{year + 1}",
+          starts_on: Date.new(year, 8, 1),
+          ends_on: Date.new(year + 1, 7, 31),
```
ollietreend, line 20: Are these dates definitely correct?

- 1 August 2023 – 31 July 2024

Schools usually start around 3 September, not August. Is 1 August an 'official' date?

app/models/claims/claim_window.rb
Nitemaeric: This is a bit of an annoyance. Yes, it shouldn't be optional.

Although, if this is required, if the `starts_on` value is missing, the missing academic year error is exposed in the form.

CleanShot 2024-07-18 at 15 19 15

The proper solution is to extract these validations out into a FormObject, but for the sake of meeting delivery constraints, this is a worthwhile workaround.

db/migrate/20240712094833_create_academic_years.rb
Nitemaeric: I've used [this](https://www.gov.uk/guidance/understanding-academic-years) as a reference for these dates.

db/migrate/20240712094833_create_academic_years.rb
JamieCleare2525: The idea of using a data migration is to clearly separate changes to the database, with data entry.
This is so if need be, the data entry can be reverted without the need to roll back the changes to the database as well.

I would also argue that both these migrations run `up_only` with no way of reverting the data entry is these are incorrect. They will also run again if the migration is reverted, creating additional/duplicate AcademicYears and ClaimWindow records. 

So is it a safe migration?

Nitemaeric: If the migration is reverted, the tables are dropped, along with the data. There will be no duplicate entries.

JamieCleare2525:
On a side note regarding the flow. I believe that the DfE usually have a one question per page guideline. I would suggest breaking the process out into separate pages for:
- Select academic year
- Select claim window

app/models/claims/claim_window.rb
JamieCleare2525: screencapture-track-and-pay-854-test-teacherservices-cloud-support-claim-windows-9af23a63-bb79-45c8-bfeb-06aec385f4d4-check-2024-07-18-15_34_24

db/migrate/20240712094833_create_academic_years.rb
ollietreend: Thanks for sharing that link. Although [that page says](https://www.gov.uk/guidance/understanding-academic-years#academic-year-start-and-end-dates):

> For the majority of students, each academic year will run from 1 September to 31 August but there are 4 official start dates in each academic year with each lasting 12 months.
> 
> If your course starts between: | Your academic year will run from:
> -- | --
> 1 August and 31 December | 1 September to 31 August
> 1 January and 31 March | 1 January to 31 December
> 1 April and 30 June | 1 April to 31 March
> 1 July and 31 July | 1 July to 30 June

Which is pretty unclear.

But it sounds like "the majority of students" it should be 1 September – 31 August?

JamieCleare2525:
I can create a ClaimWindow far in the past. There should be a limitation on how far back you can set the window.
screencapture-track-and-pay-854-test-teacherservices-cloud-support-claim-windows-9af23a63-bb79-45c8-bfeb-06aec385f4d4-check-2024-07-18-15_35_25

db/migrate/20240712094833_create_academic_years.rb
JamieCleare2525: But the issue is that there is likely nothing wrong with the database part of the migration, only the data entry. So rather than reverting the whole thing, only revert the thing that's incorrect.

JamieCleare2525:
I can also change the dates of a ClaimWindow, even when a claim has been submit for that ClaimWindow. I imagine this should be prevented.

Nitemaeric:
> I can create a ClaimWindow far in the past. There should be a limitation on how far back you can set the window. screencapture-track-and-pay-854-test-teacherservices-cloud-support-claim-windows-9af23a63-bb79-45c8-bfeb-06aec385f4d4-check-2024-07-18-15_35_25

This causes no problems, just useless data. As a Support User interface, I think it's safe to rely on Support Users _not_ adding data that isn't desired.

Nitemaeric:
> I can also change the dates of a ClaimWindow, even when a claim has been submit for that ClaimWindow. I imagine this should be prevented.

No. If the end date is moved to be in the past, it gets closed and becomes uneditable.

If a window is extended, that's all good.

db/migrate/20240712094833_create_academic_years.rb
ollietreend: It's true that we do have a pattern for data migrations, which was established in https://github.com/DFE-Digital/itt-mentor-services/pull/659. Changes to the database schema should go in `db/migrate` whereas data migrations should go in `db/data`.

It would be cleaner not to create or update records in migrations. None of the existing schema migrations currently perform data changes.

For the sake of consistency with existing migrations, would you mind moving it over @Nitemaeric? 🙏🏻 

It should be relatively easy with this generator:

```
bundle exec rails generate data_migration NAME
```

spec/models/academic_year_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+#  name       :string
+#  starts_on  :date
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+require "rails_helper"
+
+RSpec.describe AcademicYear, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
ollietreend, line 16: Likewise, although it seems like the AcademicYear model doesn't have much (any?) logic right now.

ollietreend (commented):
I appreciate you're trying to get this deployed quickly, so these comments aren't blockers. But it would be good to address them ASAP.

app/services/claims/claim_window/build.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    year = claim_window.starts_on.month > 1 ? claim_window.starts_on.year - 1 : claim_window.starts_on.year - 2
+
+    AcademicYear.create_with(
+      starts_on: Date.new(year, 8, 1),
+      ends_on: Date.new(year + 1, 7, 31),
```
ollietreend, line 27: I'm uncomfortable with the these [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) here, which are also hardcoded into the data migration.

app/services/claims/claim_window/build.rb
```diff
@@ -0,0 +1,30 @@
...
+    ).find_or_create_by!(name: "#{year} to #{year + 1}")
```
ollietreend, line 28: I also think this is slightly overstepping the boundaries of what this class should be responsible for.

How about an interface more like:

```ruby
AcademicYear.for_date(claim_window.starts_on)
```

and then make the AcademicYear model responsible for knowing about term dates and handling the 'find or create by' bit?

```ruby
class AcademicYear < ApplicationRecord
  def self.for_date(date)
    # find where(starts_on: ..date, ends_on: date..)
    # or automatically create one
  end
end
```

db/migrate/20240712103537_create_claims_claim_windows.rb
Nitemaeric: I'll move this into a data migration.

Nitemaeric (commented):
I've had a chat with @ollietreend and we've decided on what needs to be addressed today and what will be picked up next week.

ollietreend (approved):

ollietreend (approved):
Approved on the basis of the following changes:

### Today:

- [x] Remove `optional: true` from Claim Window → Academic Year association
- [x] Move data changes to data migrations
- [x] Academic year runs 1 September – 31 August

### Next week:

- [x] Remove automatic academic year magic (make support user choose the AY manually)
  - [x] Don't worry that new Academic Years won't be auto-created in future
- [x] Unit tests for models

PR #857, opened 2024-07-17
Prevent all users from making new claims when there is no claim windo…

Nitemaeric:
## Context

As we now have configurable claim windows, we want these claim windows to take effect on preventing the creation of claims once the window is closed.

- Update the `Claims::ClaimPolicy` to only allow the creation of claims if there is a current `Claims::ClaimWindow`.
- Update the `Claims::ClaimPolicy` to only allow editing claims if they can be created. (This logic works with the current constraints, technically, it may not be accurate. There is possibility that we allow changes to be made whilst a new claim cannot be created in the future.)
- If a user cannot make a claim, they will see a different guidance text within the inset text above.

## Guidance to review

- Try visiting the Claim Window settings and playing around with dates and windows.
- Try creating/submitting claims when a window is currently open or all are closed.

## Link to Trello card

https://trello.com/c/ULmE5aCc/571-as-a-user-i-should-not-be-able-to-submit-a-claim-outside-of-the-claim-window

## Screenshots

CleanShot 2024-07-17 at 10 24 41

### DIFF
--- a/app/policies/claims/claim_policy.rb
+++ b/app/policies/claims/claim_policy.rb
@@ -1,6 +1,10 @@
 class Claims::ClaimPolicy < Claims::ApplicationPolicy
+  def create?
+    Claims::ClaimWindow.current.present?
+  end
+
   def edit?
-    !record.submitted?
+    create? && !record.submitted?
   end
 
   def update?
@@ -12,7 +16,7 @@ class Claims::ClaimPolicy < Claims::ApplicationPolicy
   end
 
   def submit?
-    !user.support_user? && !record.submitted?
+    create? && !user.support_user? && !record.submitted?
   end
 
   def rejected?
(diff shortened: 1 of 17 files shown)


app/policies/claims/claim_policy.rb
```diff
@@ -1,6 +1,10 @@
 class Claims::ClaimPolicy < Claims::ApplicationPolicy
+  def create?
+    Claims::ClaimWindow.current.present?
+  end
+
   def edit?
-    !record.submitted?
+    create? && !record.submitted?
```
Nitemaeric, line 7: This needs to check if the claim's window it was submitted within is current or not.

ollietreend (approved):

ollietreend (approved):
Approved on the basis of the following changes:

### Today:

- [x] Get the tests passing

### Next week:

- [x] Update `Claims::ClaimPolicy#edit?` to take the claim's claim window into account, rather than always looking at the current claim window
  - [x] To prevent people editing historic drafts from previous claim windows, etc.
- [x] Make notification banner text dynamic

PR #860, opened 2024-07-17
Fix redirect after log out

Kizr:
## Things to check


JamieCleare2525 (approved):

spec/system/sign_in_redirect_spec.rb
```diff
@@ -0,0 +1,118 @@
+require "rails_helper"
+
+RSpec.describe "Sign in redirect", service: :placements, type: :system do
```
Nitemaeric, line 3: This is _potentially_ different per service. If we were to change this behaviour in the Claims app, no test would currently cover it.

Can we duplicate this test and put these in according directories please?

spec/system/sign_in_redirect_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+  before do
+    user_exists_in_dfe_sign_in(user:)
+  end
+
+  context "when I am a user" do
+    let(:user) { create(:placements_user) }
+    let(:school) { create(:placements_school, users: [user]) }
+
+    scenario "Logging in via the root path redirects to the placements page" do
```
ollietreend, line 12: Should this be "organisations" rather than "placements"?

```suggestion
    scenario "Logging in via the root path redirects to the organisations page" do
```

ollietreend: And from what I can tell, it looks like the user goes to the "organisations" page rather than "placements" because they have two organisations.

1. Is auto-assigned by the factory when we call `create(:placements_user)`
2. The other is assigned when we add the user to the school `create(:placements_school, users: [user])`

So the user always has two organisations, hence they see the "organisations" index page.

ollietreend (commented):
I think there's a typo in the scenario name, but otherwise this looks good to me 👍🏻

spec/system/sign_in_redirect_spec.rb
Kizr: This feels like a change that is outside the scope of this PR as it would make this test _different_ to the existing system specs, it is also unlikely to add any value to the integrity of the codebase given that the logic tested here will not differ between our services.

If this is something you feel strongly about, we can look into refactoring all of the tests to be specific to each service, but we should avoid doing so unless we can prove that it's adding value to the codebase.

spec/system/sign_in_redirect_spec.rb
Kizr: You're correct on both counts, this was from an earlier implementation that I ended up changing :)

spec/system/sign_in_redirect_spec.rb
Nitemaeric: If I add the following line into the code, would it break any specs?

```ruby
class ApplicationController < ActionController::Base
   def requested_path
    return if [sign_in_path, sign_out_path].include?(session["requested_path"])

    return session["requested_path"] if service == :claims

    @requested_path ||= session.delete("requested_path")
  end
end
```

spec/system/claims/sign_in_as_a_claims_user_spec.rb
```diff
@@ -142,4 +242,30 @@ def i_do_not_have_access_to_support_page
...
+
+  def then_i_see_a_list_of_organisations
+    expect(page).to have_current_path claims_support_schools_path, ignore_query: true
+    expect(page).to have_content("Claims School")
+  end
+
+  def and_i_see_an_empty_schools_page
+    expect(page).to have_current_path(claims_schools_path)
+    expect(page).to have_content("Organisations")
+  end
```
ollietreend, line 258: This is an awkward test because there is no "empty state" for this page in the Claims service that we can positively assert on 😕

How about something like this, to test the page **does not have links**?

```suggestion
  def and_i_see_an_empty_schools_page
    expect(page).to have_current_path(claims_schools_path)
    expect(page).to have_content("Organisations")
    expect(find("#main-content")).not_to have_link
  end
```

spec/system/claims/sign_in_as_a_claims_user_spec.rb
```diff
@@ -66,8 +66,113 @@
...
+        scenario "I am redirected to the list of my organisations" do
+          given_there_is_an_existing_user_for("Anne")
+          and_the_user_is_part_of_an_organisation(organisation)
+          and_the_user_is_part_of_an_organisation(another_organisation)
+          when_i_visit_my_account_page
+          then_i_am_redirected_to_the_sign_in_page
+          when_i_click_sign_in
+          then_i_see_user_details_for_anne
+          when_i_visit_the claims_root_path
+          # requested_path is not it again due to app/controllers/claims/pages_controller.rb:6
```
ollietreend, line 133: Is this comment still helpful? I'm not sure it's needed.

```suggestion
```

ollietreend (commented):
Nice work – thanks for consolidating the tests into the existing sign-in system specs.

ollietreend (approved):

PR #863, opened 2024-07-18
Refactor edit school contact to use wizard pattern

JamieCleare2525:
## Context

- Refactor School -> Edit school contact flow to a wizard

- Removed unnecessary code
- Implemented an EditSchoolContactWizard and related Steps

## Guidance to review

- Sign in as Anne (School user)
- Follow the process to update school contact details

## Link to Trello card

https://trello.com/c/kzsXaBg1/573-refactor-the-school-edit-school-contact-flow-to-use-the-wizard-pattern


app/controllers/placements/schools/school_contacts/edit_school_contact_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+  private
+
+  def set_school
+    school_id = params.require(:school_id)
+    @school = current_user.schools.find(school_id)
+  end
+
+  def set_school_contact
+    @school_contact = @school.school_contact
+  end
```
ollietreend, line 36: This controller assumes that the school has a contact. Do you think we should raise a `404 Not Found` error if it doesn't? 🤔

```suggestion
  def set_school_contact
    @school_contact = @school.school_contact
    
    raise ActiveRecord::RecordNotFound if @school_contact.nil?
  end
```

Otherwise I'm assuming something further downstream will explode and result in a `500 Internal Server Error`. I wonder if this is a useful case to catch at this point instead.

What do you think?

config/locales/en/placements/schools/school_contacts.yml
```diff
@@ -2,29 +2,11 @@ en:
...
-          cancel: Cancel
-        update:
-          school_contact_updated: Placement contact updated
         remove:
           page_title: Are you sure you want to remove this placement contact? - %{school_contact_name}
           are_you_sure: Are you sure you want to remove this placement contact?
           cancel: Cancel
           remove_school_contact: Remove placement contact
         destroy:
           school_contact_removed: Placement contact removed
```
ollietreend, line 11: Presumably these translations can be removed, too? It's not possible to remove a school contact.

I expect the entire file can be deleted.

app/wizards/placements/add_school_contact_wizard/school_contact_step.rb
```diff
@@ -3,13 +3,14 @@ class Placements::AddSchoolContactWizard::SchoolContactStep < Placements::BaseSt
   attribute :last_name
   attribute :email_address
 
-  validates :first_name, :last_name, presence: true
```
ollietreend, line 6: Why are we removing this validation rule? Was it not supposed to be here in the first place?

app/wizards/placements/add_school_contact_wizard/school_contact_step.rb
```diff
@@ -3,13 +3,14 @@ class Placements::AddSchoolContactWizard::SchoolContactStep < Placements::BaseSt
...
   validates :email_address, presence: true, format: { with: URI::MailTo::EMAIL_REGEXP }
   validate :new_school_contact
 
   delegate :school, to: :wizard
 
   def new_school_contact
+    return if wizard.is_a?(Placements::EditSchoolContactWizard)
+
```
ollietreend, line 13: I wonder if a more intention revealing name for this validation method would be `only_one_contact_per_school`?

Also, I think we could consolidate these two guard clauses into one (thanks to Rails 7's [excluding](https://blog.saeloun.com/2021/03/08/rails-6-1-adds-excluding-to-active-record-relation/) method):

```ruby
def only_one_contact_per_school
  another_contact = Placements::SchoolContact.where(school:).excluding(school_contact)
  return unless another_contact.exists?

  errors.add(:email_address, :taken)
end
```

This implementation should cover both scenarios – i.e. creating new contacts, and editing existing ones – because `excluding` works correctly with both `self.new_record?` and `self.persisted?` objects.

It also means the coupling between this step class and its parent wizard(s) becomes looser. It feels a bit awkward for the step to change its validation behaviour based on the wizard's class name.

ollietreend: If you _do_ choose to go with this approach, you may also want to memoize the `school_contact` method.

```ruby
def school_contact
  @school_contact ||= school.build_school_contact(first_name:, last_name:, email_address:)
end
```

Or actually, since we might be editing the existing one...

```ruby
def school_contact
  return wizard.school_contact if wizard.respond_to?(:school_contact)

  @school_contact ||= school.build_school_contact(first_name:, last_name:, email_address:)
end
```

app/wizards/placements/add_school_contact_wizard/school_contact_step.rb
JamieCleare2525: There is no validation on the SchoolContact model for these attributes. This is to allow for shared inboxes.

ollietreend (approved):
Looks good 👍🏻

PR #865, opened 2024-07-19
Missing annotations and schema changes after claims window migrations

JamieCleare2525:
## Context

- Missing annotations and schema changes after claims window migrations


ollietreend (approved):
Thanks for doing this @JamieCleare2525

PR #867, opened 2024-07-22
Add ClaimWindow to Claims

Nitemaeric:
## Context

We want to associate Claims to Claim windows when they are created. This is step 1 of 2 to implement this functionality.

1. Add foreign key column
2. Associate data and back-populate claim windows onto Claims

- Add `claim_window_id` foreign key column

## Guidance to review

This commit adds the association, but does not populate, nor perform any association.

This is to perform a safe step-by-step migration where we are sure the column is present before we start assigning data.


app/models/claims/claim.rb
```diff
@@ -38,14 +41,16 @@ class Claims::Claim < ApplicationRecord
 
   belongs_to :school
   belongs_to :provider
+  belongs_to :claim_window, optional: true
```
JamieCleare2525, line 44: Can you add a TODO comment, to remove this `optional: true`

db/migrate/20240722120510_add_claim_window_to_claims.rb
```diff
@@ -0,0 +1,5 @@
+class AddClaimWindowToClaims < ActiveRecord::Migration[7.1]
+  def change
+    safety_assured { add_reference :claims, :claim_window, foreign_key: true, type: :uuid }
```
JamieCleare2525, line 3: I don't think this should be wrapped in `safety_assured`.
`strong_migration` should suggest you include `algorithm: :concurrently`.

Nitemaeric: It does suggest that, as it assumes there is existing table that will need to be locked in order to populate the index. As this is a relatively small table, I'd urge for the simplicity of this migration and go forward with the `safety_assured` route.

There is very low risk this will be problematic or long-running.

app/models/claims/claim.rb
Nitemaeric: This follow-up PR removes it on this line: https://github.com/DFE-Digital/itt-mentor-services/pull/868/files#diff-af5fcd79400b248a8923c036a23d85f33328cb0021add465590354c9ffc58902R44

Is that sufficient enough?

db/migrate/20240722120510_add_claim_window_to_claims.rb
JamieCleare2525: I've just run it myself and it suggests:

```
=== Dangerous operation detected #strong_migrations ===

Adding a foreign key blocks writes on both tables. Instead, use:

class AddClaimWindowToClaims < ActiveRecord::Migration[7.1]
  disable_ddl_transaction!

  def change
    add_reference :claims, :claim_window, type: :uuid, index: {algorithm: :concurrently}
  end
end
```

I would recommend you do what strong migrations suggests, rather than wrap it in `safety_assured`

Nitemaeric: This is now addressed.

```diff
  === Dangerous operation detected #strong_migrations ===
  
  Adding a foreign key blocks writes on both tables. Instead, use:
  
  class AddClaimWindowToClaims < ActiveRecord::Migration[7.1]
    disable_ddl_transaction!
  
    def change
      add_reference :claims, :claim_window, type: :uuid, index: {algorithm: :concurrently}
    end
  end

+ Then add the foreign key in separate migrations.
```

There is an additional step. I'll add the foreign key in a future PR. This is why I wanted to assume safety and skip this additional commit.

JamieCleare2525 (approved):

PR #868, opened 2024-07-22
Populate claim windows on claims

Nitemaeric:
## Context

Now that we have a `claim_window` association on the `Claim` model, we can start associating the claim windows when a claim is created.

- Back-populate claim windows onto all existing claims.
- When a claim is being created, assign the current claim window to the claim being created.

## Guidance to review

- Locally, try creating a Claim. It will associate the "current" claim window to the claim.

### DIFF
--- a/db/data/20240722123910_add_claim_window_to_claims.rb
+++ b/db/data/20240722123910_add_claim_window_to_claims.rb
@@ -1,0 +1,12 @@
+class AddClaimWindowToClaims < ActiveRecord::Migration[7.1]
+  def up
+    date = Date.parse("2 May 2024")
+    claim_window = Claims::ClaimWindow.find_by_date(date)
+
+    Claims::Claim.update_all(claim_window_id: claim_window.id) if claim_window
+  end
+
+  def down
+    Claims::Claim.update_all(claim_window_id: nil)
+  end
+end
--- a/app/controllers/claims/schools/claims_controller.rb
+++ b/app/controllers/claims/schools/claims_controller.rb
@@ -80,7 +80,7 @@ class Claims::Schools::ClaimsController < Claims::ApplicationController
   end
 
   def default_params
-    { school: @school, current_user: }
+    { school: @school, current_user:, claim_window: Claims::ClaimWindow.current }
   end
 
   def claim_provider_form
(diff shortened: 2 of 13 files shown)


spec/factories/academic_year.rb
```diff
@@ -0,0 +1,11 @@
+FactoryBot.define do
+  factory :academic_year do
+    trait :current do
+      starts_on { Date.new(Date.current.year - 1, 9, 1) }
+      ends_on { Date.new(Date.current.year, 8, 31) }
```
JamieCleare2525, line 5:
```suggestion
      starts_on { Date.parse("1 September #{Date.current.year - 1}") }
      ends_on { Date.parse("31 August #{Date.current.year}") }
```
This will increase readability.

I would also thing about doing something conditional when calculating these dates. So if you run these tests on 2nd September 2024, this will generate an academic year for 1st September 2023 - 31st August 2024, which might break some tests.

db/data/20240722123910_add_claim_window_to_claims.rb
```diff
@@ -0,0 +1,9 @@
+class AddClaimWindowToClaims < ActiveRecord::Migration[7.1]
+  def up
+    Claims::Claim.update_all(claim_window_id: Claims::ClaimWindow.first.id)
```
JamieCleare2525, line 3:
```suggestion
    Claims::Claim.update_all(claim_window_id: Claims::ClaimWindow.current.id)
```

Nitemaeric: This is incorrect. The correct claim window has now closed.

JamieCleare2525: I which case I would first find the ClaimWindow based on the start and end date you expect, rather than rely on `.first` which might return an incorrect ClaimWindow.

Nitemaeric: 👍 I'll update.

spec/factories/academic_year.rb
Nitemaeric: I've tried running the specs with the following `around` hook.

```ruby
config.around do |example|
  Timecop.travel(Time.zone.local(2024, 9, 2)) do
    example.run
  end
end
```

All pass :)

JamieCleare2525: In which case I'm assuming your tests/code doesn't validate that the ClaimWindow dates must be within the AcademicYear `starts_on` and `ends_on` dates. 
So I don't think this will break your implementation.

JamieCleare2525 (approved):

PR #876, opened 2024-07-25
Google routes spike

Kizr:
## Context

Travel distances and duration are required for upcoming features, no gems currently exist for the Google Routes API.

- Enable caching on the test environment
- Create Google Routes API library
- Create TravelTime service

## Guidance to review

- Checkout this branch
- Have a go in your local console

## Link to Trello card

[Create a client for accessing the Google Maps Routes API](https://trello.com/c/nmu1FZtV/576-create-a-client-for-accessing-the-google-maps-routes-api)


ollietreend (commented):
This is looking really nice – great work @JamieCleare2525 & @Kizr 🙌🏻 I can't wait to see it powering a user-facing feature!! 😬

I like the approach you've taken towards 'enriching' the collection so it's easy for the view layer to use these travel times in the same way as any other attribute.

I've added a few comments with some things to think about. Happy to talk them through!

config/environments/test.rb
```diff
@@ -25,7 +25,7 @@
   # Show full error reports and disable caching.
   config.consider_all_requests_local = true
   config.action_controller.perform_caching = false
-  config.cache_store = :null_store
+  config.cache_store = :memory_store
```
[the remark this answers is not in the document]
ollietreend: @Kizr I think it should be possible to enable the cache just as part of the unit test, using something like this:

https://dev.to/epigene/simple-testing-of-rails-cache-with-rspec-j5

So we can verify that the cache gets written to / read from as part of the unit test, and then the rest of the test suite can continue to run without caching.

app/services/placements/travel_time.rb
```diff
@@ -1,43 +1,53 @@
...
 
-      destination.update!(travel_duration:, travel_distance:)
+  def travel_data(index, travel_mode:)
+    results = Rails.cache.fetch(cache_key(travel_mode:), expires_in: 1.day) do
+      routes_client.travel_time(origin_address, destinations, travel_mode:)
     end
 
-    destinations.sort_by(&:travel_duration)
+    results.find { |datum| datum["destinationIndex"] == index }
   end
```
ollietreend, line 46: My concern with this approach is that we end up calling `travel_data` multiple times – once for each destination – because it's inside the `destinations.map` loop. So we'll call it 25 times given a collection of 25 destinations.

While this should in theory be okay because `travel_data` makes use of the Rails cache, it will still result in multiple reads & deserialisations of the cached values. This feels a bit wasteful.

Instead, perhaps we could refactor it so that `travel_data` returns the entire result-set as an array, and we hold that outside of the `destinations.map` loop. And then within the loop we can use the `index` just to pluck the correct result from the overall array.

What do you think?

spec/services/placements/travel_time_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+    it "returns the school collection, sorted by travel duration" do
+      results = service
+      expect(results).to eq [placements_schools[1], placements_schools[0], placements_schools[2]]
+    end
+
+    it "caches the travel time data" do
+      fingerprint = Digest::SHA256.hexdigest(["SW1A 1AA", destinations.pluck(:latitude, :longitude), "DRIVE"].to_json)
+      service
+
+      expect(Rails.cache.exist?("placements_travel_time_#{fingerprint}")).to be true
```
ollietreend, line 75: Rather than (or in addition to?) checking for the existence of a particular cache key, I wonder if we could also check that we only actually hit the API just once?

The fact that we cache API responses seems like an important behaviour, whereas the specific cache key used feels more like a side effect/implementation detail (handled by a `private` method).

You could also argue that the existence of a cache key doesn't necessarily mean that you _use_ it in subsequent calls.

```suggestion
      fingerprint = Digest::SHA256.hexdigest(["SW1A 1AA", destinations.pluck(:latitude, :longitude), "DRIVE"].to_json)

      expect(google_routes_client).to receive(:travel_time).once

      # this should run the service twice
      expect(service).to eq(described_class.call(origin_address:, destinations:))

      expect(Rails.cache.exist?("placements_travel_time_#{fingerprint}")).to be true
```

ollietreend (approved):

PR #891, opened 2024-08-02
Added validation to academic year model

dp-daly:
## Context

Added validation rules to the AcademicYear model. 

- Rules confirming presence of name and years added to AcademicYear model.
- Added claims association to ClaimWindow model.
- Added accompanying RSpec tests.

## Guidance to review

Run tests on models folder.

## Link to Trello card

https://trello.com/c/BkTUj28v/634-add-validation-rules-for-fields-on-the-academic-year-model

### DIFF
--- a/app/models/academic_year.rb
+++ b/app/models/academic_year.rb
@@ -10,9 +10,6 @@
 #  updated_at :datetime         not null
 #
 class AcademicYear < ApplicationRecord
-  has_many :claim_windows, class_name: "Claims::ClaimWindow"
-  has_many :claims, through: :claim_windows
-
   validates :name, presence: true
   validates :starts_on, presence: true
   validates :ends_on, presence: true, comparison: { greater_than_or_equal_to: :starts_on }
--- a/app/models/academic_year.rb
+++ b/app/models/academic_year.rb
@@ -10,4 +10,10 @@
 #  updated_at :datetime         not null
 #
 class AcademicYear < ApplicationRecord
+  has_many :claim_windows, class_name: "Claims::ClaimWindow"
+  has_many :claims, through: :claim_windows
+
+  validates :name, presence: true
+  validates :starts_on, presence: true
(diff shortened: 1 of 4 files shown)


app/models/academic_year.rb
```diff
@@ -10,4 +10,10 @@
 #  updated_at :datetime         not null
 #
 class AcademicYear < ApplicationRecord
+  has_many :claim_windows, class_name: "Claims::ClaimWindow"
+  has_many :claims, through: :claim_windows
```
Nitemaeric, line 14: Can we avoid leaking the claims-specific logic into the top-level `AcademicYear`, please? 🙏

app/models/claims/claim_window.rb
```diff
@@ -25,6 +25,7 @@ class Claims::ClaimWindow < ApplicationRecord
   default_scope -> { kept }
 
   belongs_to :academic_year
+  has_many :claims, class_name: "Claims::Claim"
```
Nitemaeric, line 28: We don't really have a need for this just yet. It might be better to leave out this code until we have a use for it.

Nitemaeric (commented):
A couple comments :)

app/models/academic_year.rb
JamieCleare2525: Similar to my other comment:
https://github.com/DFE-Digital/itt-mentor-services/pull/891/files#r1702024680

I would assume you would want the reverse of `belongs_to :academic_year` in ClaimWindow, in AcademicYear. Generally this is what is done when creating associations.

JamieCleare2525: @dp-daly We've discussed this with @Nitemaeric and have come to an agreement to remove the associations work from this PR. The reasoning is to implement them if/when the associations are required to be used as part of the code.

JamieCleare2525 (approved):

Nitemaeric (approved):

PR #893, opened 2024-08-05
Add a migration step to the deployment

ollietreend:
## Context

We've been experiencing issues with review apps due to the way database migrations are run by the deployment pipeline.

Migrations currently run inline within the web app container. The web app is started with the command `rails db:migrate && rails server`, which is a straightforward but slightly hacky way to run migrations before starting the server. It works on the assumption that your web app is the only service which depends on the database schema being up to date.

However in our case, we have a separate service called [Good Job](https://github.com/bensheldon/good_job) which runs alongside the web app to process background jobs. This service also has a dependency on the same Postgres database that our web app uses. Therefore it's important that our deployment pipeline runs migrations _before_ attempting to start the web app and Good Job worker, since they both depend on schema migrations having been successfully run.

For more details about the problem, refer to [the Trello card](https://trello.com/c/LcIqY4Fo/630-update-terraform-configuration-to-run-migrations-before-starting-the-app-and-worker) and [this Slack thread](https://ukgovernmentdfe.slack.com/archives/C011EM7HU85/p1722270014308649).

This pull request makes the following changes:

### 1. Run migrations before starting the web app and Good Job worker

This is done through the introduction of a new Kubernetes Job resource in the Terraform template.

When deployed, Terraform will wait for the migration job to complete before spinning up the web app and worker. This works because I've configured the migration job with `wait_for_completion = true`, and made the web app & worker modules dependent on it with `depends_on = [kubernetes_job.migrations]`.

### 2. Use `rails db:prepare` to speed up review app initialisation

Rather than using `rails db:migrate`, I've switched to using `rails db:prepare`.

This means that new review apps should spin up quicker because the database schema will be loaded from `schema.rb` rather than generated by running every migration sequentially.

It also means that we no longer need to explicitly seed the database in review apps. That's taken care of by `rails db:prepare` too.

For more details about what the command does, see the [Rails migrations guide](https://guides.rubyonrails.org/active_record_migrations.html#preparing-the-database).

### 3. Remove `rails db:migrate` from `Dockerfile`

The default image command no longer needs to include `rails db:migrate` because migrations are now handled separately at deploy time.

## Guidance to review

See the review app for this PR.

## Link to Trello card

https://trello.com/c/LcIqY4Fo/630-update-terraform-configuration-to-run-migrations-before-starting-the-app-and-worker

### DIFF
--- a/.github/workflows/build_and_deploy.yml
+++ b/.github/workflows/build_and_deploy.yml
@@ -50,19 +50,6 @@ jobs:
           azure-credentials: ${{ secrets.AZURE_CREDENTIALS }}
           pull-request-number: ${{ github.event.number }}
 
-      - uses: azure/login@v2
-        with:
-          creds: ${{ secrets.AZURE_CREDENTIALS }}
-
-      - name: Seed Review App
-        shell: bash
-        if: github.event.number != ''
-        run: |
-          make ci review get-cluster-credentials
-          kubectl exec -n bat-qa deployment/itt-mentor-services-${{ github.event.number }} -- sh -c "export GIAS_CSV_BASE_URL=https://ea-edubase-api-prod.azurewebsites.net/edubase/downloads/public  && export PUBLISH_BASE_URL=https://qa.api.publish-teacher-training-courses.service.gov.uk// && cd /app && /usr/local/bin/rake db:seed"
-        env:
-          PR_NUMBER: ${{ github.event.number }}
-
       - name: Post comment to Pull Request ${{ github.event.number }}
         if: ${{ github.event_name == 'pull_request' }}
         uses: marocchino/sticky-pull-request-comment@v2
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -9,11 +9,11 @@ on:
         type: choice
         default: review
         options:
-        - qa
-        - review
-        - staging
-        - sandbox
-        - dv
+          - qa
+          - review
+          - staging
+          - sandbox
+          - dv
       sha:
         description: Commit sha to be deployed
         required: true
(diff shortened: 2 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -9,11 +9,11 @@ on:
...
-        - qa
-        - review
-        - staging
-        - sandbox
-        - dv
+          - qa
+          - review
+          - staging
+          - sandbox
+          - dv
```
JamieCleare2525, line 16: This seems like an odd change.

ollietreend: My IDE did it automatically when I updated that file 🤷🏻 presumably it thought the formatting was wrong.

I can revert this particular chunk if needed. But VS Code will do it again next time I edit the file 🫠

.github/workflows/build_and_deploy.yml
```diff
@@ -50,19 +50,6 @@ jobs:
           azure-credentials: ${{ secrets.AZURE_CREDENTIALS }}
           pull-request-number: ${{ github.event.number }}
 
-      - uses: azure/login@v2
-        with:
-          creds: ${{ secrets.AZURE_CREDENTIALS }}
-
-      - name: Seed Review App
```
saliceti, line 57: Why remove db:seed?

terraform/application/application.tf
```diff
@@ -22,8 +22,57 @@ module "application_configuration" {
   }
 }
 
+# Run database migrations
+# https://guides.rubyonrails.org/active_record_migrations.html#preparing-the-database
+# https://github.com/ilyakatz/data-migrate
+# Terraform waits for this to complete before starting web_application and worker_application
+resource "kubernetes_job" "migrations" {
```
saliceti, line 29: Could you add the [security_context](https://github.com/DFE-Digital/terraform-modules/blob/main/aks/application/resources.tf#L169C11-L169C27) and [resources](https://github.com/DFE-Digital/terraform-modules/blob/main/aks/application/resources.tf#L103) blocks?

saliceti: module.cluster_data.configuration_map.cpu_min

.github/workflows/build_and_deploy.yml
ollietreend: @saliceti see this commit message https://github.com/DFE-Digital/itt-mentor-services/pull/893/commits/c78dea254a580f06fc44328c60a3e1c2924d2e1c

TL;DR: `rails db:prepare` automatically runs seeds for you when initialising a new database.

terraform/application/application.tf
ollietreend: Added in f982d30529ef675d756d457c1b7ca1867a498b6b

saliceti (approved):

JamieCleare2525 (approved):

PR #856, opened 2024-07-17
Refactor partnerships to use wizard pattern

JamieCleare2525:
## Context

Refactor Organisation (School/Provider) -> Add Partnership (Partner provider/Partner school) flow to a wizard

## Changes proposed in this pull request

- Removed unnecessary code
- Implemented an AddPartnershipWizard and related Steps

## Guidance to review

- Log in as Anne (School User)
- Follow the process to add a new partner provider

- Log in as Patricia (Provider User)
- Follow the process to add a new partner school

## Link to Trello card

https://trello.com/c/jvwvq5De/564-refactor-the-organisation-add-partnership-flow-to-use-the-wizard-pattern

PR #761, opened 2024-06-12
Persist date of birth on validation error

Kizr:
## Context

When entering an invalid TRN or Date of Birth whilst adding a mentor, the date of birth does not persist.

## Changes proposed in this pull request

- [x] Use a struct to populate the date of birth when an invalid date is present.

## Guidance to review

- Log in as Anne
- Add a mentor
- Enter a number into one of the three date of birth fields
- Click continue
- See a validation error and your entry has persisted

## Link to Trello card

[Add mentor → preserve partially entered DOB fields](https://trello.com/c/c7hMIOEu/486-add-mentor-%E2%86%92-preserve-partially-entered-dob-fields)

## Screenshots

image

PR #905, opened 2024-08-08
Select academic year when creating a claim window

Nitemaeric:
## Context

Follows up on the committed task here: https://github.com/DFE-Digital/itt-mentor-services/pull/854#pullrequestreview-2186352476

We want to avoid automatically associating academic years to claim windows, as it reduces flexibility but also adds additional logic to the service.

- Require user selecting the academic year for a claim window.
- Add `Claims::ClaimWindowForm` form object to handle `Claims::ClaimWindow` validations.
  - As the `belongs_to :academic_year` adds a `required` validation, a lot of error-movement is required to show the `:academic_year` error on the `:academic_year_id` field, which is present on the claim window form.
  - Moving the claim window form data object from a model to a form object allows us to validate the form submission in isolation.
- Allow over year-long windows at the Model level. Prevent over year-long windows at the FormObject level.
- Add a `DateAttributes` concern for FormObjects to clean up the logic. This is adopted in the new `Claims::ClaimWindowForm` form object, but not elsewhere.

## Guidance to review

- Sign in a a support user.
- Visit the settings tab.
- Visit the "Claim windows" link.
- Try adding and editing (only active or upcoming) claim windows.

## Link to Trello card

https://trello.com/c/9OcRpTbF/639-claim-windows-should-select-an-academic-year-and-not-automatically-associate-together

## Screenshots

| Page | Screenshot |
| ---- | ---------- |
| New  | CleanShot 2024-08-13 at 15 52 00 |
| Edit | CleanShot 2024-08-13 at 15 51 10 |

### DIFF
--- a/app/controllers/claims/support/claim_windows_controller.rb
+++ b/app/controllers/claims/support/claim_windows_controller.rb
@@ -2,50 +2,36 @@ class Claims::Support::ClaimWindowsController < Claims::Support::ApplicationCont
   before_action :set_claim_window, only: %i[show edit edit_check update remove destroy]
   before_action :authorize_claim_window
 
+  helper_method :claim_window_form
+
   def index
     @claim_windows = Claims::ClaimWindow.order(starts_on: :desc)
   end
 
-  def new
-    @claim_window = if params[:claims_claim_window]
-                      Claims::ClaimWindow.new(claim_window_params)
-                    else
-                      Claims::ClaimWindow.new
-                    end
-  end
+  def new; end
 
   def new_check
-    @claim_window = Claims::ClaimWindow::Build.call(claim_window_params:)
-
-    render :new unless @claim_window.valid?
+    render :new unless claim_window_form.valid?
   end
 
   def create
-    @claim_window = Claims::ClaimWindow::Build.call(claim_window_params:)
-    @claim_window.save!
+    claim_window_form.save!
 
-    redirect_to claims_support_claim_windows_path, flash: { success: "Claim window created" }
+    redirect_to claims_support_claim_windows_path, flash: { success: t(".success") }
   end
 
   def show; end
 
-  def edit
-    if params[:claims_claim_window]
-      @claim_window.assign_attributes(claim_window_params)
-    end
-  end
+  def edit; end
 
   def edit_check
-    @claim_window = Claims::ClaimWindow::Build.call(claim_window: @claim_window, claim_window_params:)
-
-    render :edit unless @claim_window.valid?
+    render :edit unless claim_window_form.valid?
   end
 
   def update
-    @claim_window = Claims::ClaimWindow::Build.call(claim_window: @claim_window, claim_window_params:)
-    @claim_window.save!
+    claim_window_form.save!
 
-    redirect_to claims_support_claim_window_path(@claim_window), flash: { success: "Claim window updated" }
+    redirect_to claims_support_claim_window_path(@claim_window), flash: { success: t(".success") }
   end
 
   def remove; end
@@ -53,13 +39,13 @@ class Claims::Support::ClaimWindowsController < Claims::Support::ApplicationCont
   def destroy
     @claim_window.discard!
 
-    redirect_to claims_support_claim_windows_path, flash: { success: "Claim window removed" }
+    redirect_to claims_support_claim_windows_path, flash: { success: t(".success") }
   end
 
   private
 
-  def claim_window_params
-    params.require(:claims_claim_window).permit(:starts_on, :ends_on)
+  def claim_window_form_params
+    params.fetch(:claims_claim_window_form, {}).permit(:starts_on, :ends_on, :academic_year_id)
   end
 
   def set_claim_window
@@ -68,5 +54,9 @@ class Claims::Support::ClaimWindowsController < Claims::Support::ApplicationCont
 
   def authorize_claim_window
     authorize @claim_window || Claims::ClaimWindow
+  end
+
+  def claim_window_form
+    @claim_window_form ||= Claims::ClaimWindowForm.new(id: params[:id], **claim_window_form_params)
   end
 end
--- a/app/forms/application_form.rb
+++ b/app/forms/application_form.rb
@@ -1,7 +1,9 @@
 class ApplicationForm
   include ActiveModel::Model
+  include ActiveModel::Attributes
   include ActionView::Helpers::TranslationHelper
   include Rails.application.routes.url_helpers
+  include DateAttributes
 
   def initialize(attributes = {})
     @virtual_path = "forms/#{self.class.name.underscore}"
--- a/app/forms/claims/claim_window_form.rb
+++ b/app/forms/claims/claim_window_form.rb
@@ -1,0 +1,60 @@
+class Claims::ClaimWindowForm < ApplicationForm
+  attr_accessor :id
+
+  date_attribute :starts_on
+  date_attribute :ends_on
+
+  attribute :academic_year_id
+
+  validates :starts_on, date_presence: true
+  validates :ends_on, date_presence: true
+  validates :ends_on, comparison: { greater_than_or_equal_to: :starts_on }, if: -> { starts_on_is_date? && ends_on_is_date? }
+  validates :academic_year_id, presence: true
+
+  validate :does_not_start_within_another_claim_window, if: :starts_on_is_date?
+  validate :does_not_end_within_another_claim_window, if: :ends_on_is_date?
+
+  def initialize(attributes = {})
+    super
+
+    if id
+      self.starts_on ||= claim_window.starts_on
+      self.ends_on ||= claim_window.ends_on
+      self.academic_year_id ||= claim_window.academic_year_id
+    end
+  end
+
+  def persist
+    claim_window.update!(starts_on:, ends_on:, academic_year_id:)
+  end
+
+  def new_back_path
+    new_claims_support_claim_window_path(claims_claim_window_form: slice(:starts_on, :ends_on, :academic_year_id))
+  end
+
+  def edit_back_path
+    edit_claims_support_claim_window_path(id, claims_claim_window_form: slice(:starts_on, :ends_on, :academic_year_id))
+  end
+
+  def academic_year_name
+    AcademicYear.find_by(id: academic_year_id)&.name
+  end
+
+  private
+
+  def claim_window
+    @claim_window ||= id ? Claims::ClaimWindow.find(id) : Claims::ClaimWindow.new
+  end
+
+  def does_not_start_within_another_claim_window
+    return unless Claims::ClaimWindow.where.not(id:).where("starts_on <= :starts_on AND ends_on >= :starts_on", starts_on:).exists?
+
+    errors.add(:starts_on, :overlap)
+  end
+
+  def does_not_end_within_another_claim_window
+    return unless Claims::ClaimWindow.where.not(id:).where("starts_on <= :ends_on AND ends_on >= :ends_on", ends_on:).exists?
+
+    errors.add(:ends_on, :overlap)
+  end
+end
--- a/app/forms/concerns/date_attributes.rb
+++ b/app/forms/concerns/date_attributes.rb
@@ -1,0 +1,45 @@
+module DateAttributes
+  extend ActiveSupport::Concern
+
+  IncompleteDate = Struct.new(:year, :month, :day)
+
+  class_methods do
+    private
+
+    def date_attribute(attribute_name)
+      attr_accessor :"#{attribute_name}_year", :"#{attribute_name}_month", :"#{attribute_name}_day"
+
+      alias_method :"#{attribute_name}(1i)", :"#{attribute_name}_year"
+      alias_method :"#{attribute_name}(2i)", :"#{attribute_name}_month"
+      alias_method :"#{attribute_name}(3i)", :"#{attribute_name}_day"
+
+      alias_method :"#{attribute_name}(1i)=", :"#{attribute_name}_year="
+      alias_method :"#{attribute_name}(2i)=", :"#{attribute_name}_month="
+      alias_method :"#{attribute_name}(3i)=", :"#{attribute_name}_day="
+
+      define_method "#{attribute_name}=" do |value|
+        attribute_value = ActiveModel::Type::Date.new.cast(value)
+
+        raise ArgumentError, "Invalid date format" unless attribute_value.is_a?(Date) || attribute_value.nil?
+
+        send("#{attribute_name}_year=", attribute_value&.year)
+        send("#{attribute_name}_month=", attribute_value&.month)
+        send("#{attribute_name}_day=", attribute_value&.day)
+      end
+
+      define_method attribute_name do
+        year, month, day = %w[year month day].map { |date_part| send("#{attribute_name}_#{date_part}") }
+
+        if year.present? && month.present? && day.present?
+          Date.new(year.to_i, month.to_i, day.to_i)
+        else
+          IncompleteDate.new(year, month, day)
+        end
+      end
+
+      define_method "#{attribute_name}_is_date?" do
+        send(attribute_name).is_a?(Date)
+      end
+    end
+  end
+end
--- a/spec/forms/application_form_spec.rb
+++ b/spec/forms/application_form_spec.rb
@@ -54,4 +54,102 @@ RSpec.describe ApplicationForm do
       end
     end
   end
+
+  describe ".date_attribute" do
+    subject(:example_form) { ExampleForm.new }
+
+    before do
+      example_form_class = Class.new(ApplicationForm) do
+        date_attribute :date_of_birth
+      end
+
+      stub_const("ExampleForm", example_form_class)
+    end
+
+    it "adds getters and setters for the attribute suffixed with (1i), (2i), and (3i)" do
+      example_form.public_send("date_of_birth(1i)=", 2024)
+      example_form.public_send("date_of_birth(2i)=", 5)
+      example_form.public_send("date_of_birth(3i)=", 2)
+
+      expect(example_form.public_send("date_of_birth(1i)")).to eq(2024)
+      expect(example_form.public_send("date_of_birth(2i)")).to eq(5)
+      expect(example_form.public_send("date_of_birth(3i)")).to eq(2)
+    end
+
+    it "aliases the (1i), (2i), and (3i) suffixed attribute accessors to year, month, and day, respectively" do
+      example_form.public_send("date_of_birth(1i)=", 2024)
+      example_form.public_send("date_of_birth(2i)=", 5)
+      example_form.public_send("date_of_birth(3i)=", 2)
+
+      expect(example_form.date_of_birth_year).to eq(2024)
+      expect(example_form.date_of_birth_month).to eq(5)
+      expect(example_form.date_of_birth_day).to eq(2)
+
+      example_form.public_send("date_of_birth_year=", 2025)
+      example_form.public_send("date_of_birth_month=", 6)
+      example_form.public_send("date_of_birth_day=", 3)
+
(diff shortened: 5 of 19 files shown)


app/forms/claims/claim_window_form.rb
```diff
@@ -0,0 +1,68 @@
...
+    @claim_window ||= id ? Claims::ClaimWindow.find(id) : Claims::ClaimWindow.new
+  end
+
+  def is_not_longer_than_an_academic_year
+    return if starts_on.blank? || ends_on.blank?
+
+    if (ends_on - starts_on).to_i > 365
+      errors.add(:base, :longer_than_academic_year)
+    end
+  end
```
JamieCleare2525, line 55: You've removed this validation from the claim window model. I imagine it needs removing here too.

app/forms/claims/claim_window_form.rb
```diff
@@ -0,0 +1,68 @@
...
+    return unless Claims::ClaimWindow.where.not(id:).where("starts_on <= :starts_on AND ends_on >= :starts_on", starts_on:).exists?
+
+    errors.add(:starts_on, :overlap)
+  end
+
+  def does_not_end_within_another_claim_window
+    return unless Claims::ClaimWindow.where.not(id:).where("starts_on <= :ends_on AND ends_on >= :ends_on", ends_on:).exists?
+
+    errors.add(:ends_on, :overlap)
+  end
```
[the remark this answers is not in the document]
Nitemaeric: Hmm, there are some challenges with this 🤔 

The `ClaimWindow` model validates the presence of `academic_year` with the `belongs_to` line. If we were to put all model errors onto the form object errors, the page would show 2 "academic year" errors - 1, which links to the radio field correctly - `academic_year_id` presence, and 2, the `academic_year` association presence - which only appears in the top summary.

This also defeats the purpose I had for making the form object, as it doesn't do anything special, it just redefines how validations should work at the form level.

app/forms/application_form.rb
```diff
@@ -1,7 +1,9 @@
 class ApplicationForm
   include ActiveModel::Model
+  include ActiveModel::Attributes
   include ActionView::Helpers::TranslationHelper
   include Rails.application.routes.url_helpers
+  include DateAttributes
```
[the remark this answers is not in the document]
Nitemaeric: Is it more beneficial to standardise the tooling available for our `FormObjects`?

I'd also want to add `ActiveModel::Attributes` to the base `ApplicationForm` really, rather than pick and choose which class methods it has access to, but may not use.

Implicit behaviours definitely need to be included case by case though.

So like, having us have a standard `ActiveRecord` base, but choosing whether or not to use the provided tooling, e.g. `belongs_to` and `validates` etc, makes more sense to me.

What do you think?

app/forms/claims/claim_window_form.rb
Nitemaeric: > Allow over year-long windows at the Model level. Prevent over year-long windows at the FormObject level.

No, this is intentional. In the case that this behaviour is desired, it can be overridden by a developer.

JamieCleare2525: Whilst I take issue that a Developer would need to change this data, and that I believe if this is valid then a Support User should be trusted enough to extend a ClaimWindow without assistance from Developers.

If this is intentional I will concede.

However, this validation checks that a ClaimWindow is less than a whole year (even though a leap year is 366 days). A better check would be to check if the start and end dates are within the boundaries of an actual AcademicYear. 
Otherwise a SupportUser can make a ClaimWindow for 29 Aug 2024 - 28 Aug 2025, which doesn't seem like the desired outcome.

app/forms/claims/claim_window_form.rb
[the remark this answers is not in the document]
JamieCleare2525: I would argue that you are probably correct in that this form object isn't really necessary, and is likely a method of getting around a form naming issue you are having.
In any case, you can chose to remove the academic_year validation from the errors being added using the method I have suggested. 
It shouldn't be the responsibility of the form to duplicate validations of the model it is intended to represent.

app/forms/application_form.rb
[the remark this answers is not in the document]
JamieCleare2525: This kind of reaches into the code pollution argument. As no placements forms require this `DateAttributes` concern added, and if added to the base model would be included into every form object.
Given this is also not being introduced across the board, such as with MentorForm which has date attributes, it's my opinion that the `DateAttributes` concern only be added to forms which require it.

I have no issues with adding `ActiveModel::Attributes` as the only form currently used by Placements is Placements::Placements::FilterForm, which includes them anyway.

app/forms/concerns/date_attributes.rb
ollietreend, line 1: Can we have a unit test to cover this concern, please?

The heavy use of dynamic programming here makes it difficult to parse. I'd like to see a unit test which clearly demonstrates the behaviour it's supposed to add.

app/controllers/claims/support/claim_windows_controller.rb
```diff
@@ -69,4 +55,12 @@ def set_claim_window
   def authorize_claim_window
     authorize @claim_window || Claims::ClaimWindow
   end
+
+  def claim_window_form
+    @claim_window_form ||= if params[:claims_claim_window_form]
+                             Claims::ClaimWindowForm.new(id: params[:id], **claim_window_form_params)
+                           else
+                             Claims::ClaimWindowForm.new(id: params[:id])
+                           end
```
ollietreend, line 64: Since you never want to raise a [`ActionController::ParameterMissing`](https://api.rubyonrails.org/classes/ActionController/ParameterMissing.html) error, it would be more succinct to use `params.fetch` rather than `params.require`.

```diff
def claim_window_form_params
-  params.require(:claims_claim_window_form).permit(:starts_on, :ends_on, :academic_year_id)
+  params.fetch(:claims_claim_window_form, {}).permit(:starts_on, :ends_on, :academic_year_id)
end
```

This way, the `claim_window_form_params` method will always return a params object that can be safely expanded using the double splat operator.

```suggestion
    @claim_window_form ||= Claims::ClaimWindowForm.new(id: params[:id], **claim_window_form_params)
```

app/forms/claims/claim_window_form.rb
Nitemaeric: I'll make this adjustment.

Nitemaeric: I made a mistake here. This isn't quite right. A claim window does not need to be within the bounds of an academic year. A claim window for a given year will be opened after the academic year has ended.

app/forms/concerns/date_attributes.rb
Nitemaeric: I've added some unit tests to the ApplicationForm specs.

app/forms/claims/claim_window_form.rb
JamieCleare2525: When then at the very least, there should be validation to ensure that the placement window begins within the bounds of an academic year

Nitemaeric: That's also not quite right. There's no necessity for the window to begin, nor end, within the bounds of its associated academic year.

JamieCleare2525: So given your last comment "There's no necessity for the window to begin, nor end, within the bounds of its associated academic year." I would return to my original comment, that this validation should be removed, as this validation is not shared with the model this form is intended to represent. Support Users should "given what you've said" be able to create an ClaimWindow which branches over academic years.

This does lead me to question the necessary association of a ClaimWindow with an AcademicYear though.

Nitemaeric: > Support Users should "given what you've said" be able to create an ClaimWindow which branches over academic years.

Arguably, yes. I can remove the limitation if you don't think it is reasonable.

It was introduced to help add sensible limits, similar to cases of not allowing people to add date of births before 1900.

> This does lead me to question the necessary association of a ClaimWindow with an AcademicYear though.

All claims made within a claim window should all be associated to the same academic year. This allows us to query claims against both claim windows _and_ academic years.

JamieCleare2525: Screenshot 2024-08-21 at 16 03 03

But given your current implementation I can still do ^

JamieCleare2525: In summary, I think the validations should reflect exactly what's on the model. The form shouldn't be adding additional validation, especially when it's been removed from the model itself given the existing use case that a claim window, can in fact, be longer than an year in length.

Nitemaeric: In that case, let's remove the _sensibility_ validations, as it is likely hard to determine what is _actually_ sensible.

We'll rely on the Support team to decide how they want to schedule windows. Technically, we'll validate that the records are logically sound - such as the presence of fields, no overlap, and date ranges that start and end respectively.

app/forms/claims/claim_window_form.rb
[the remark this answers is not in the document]
Nitemaeric: I've run into a few issues with the implementation here.

```ruby
validates :validate_claim_window

def validate_claim_window
  if claim_window.invalid?
    claim_window.errors.each do |err|
      errors.add(err.attribute, err.message)
    end
  end
end
```

```ruby
form = Claims::ClaimWindowForm.new(starts_on: Date.parse("21 Aug 2024"))
form.errors # As the `claim_window.starts_on` value hasn't been populated yet, the claim_window's starts_on missing error is propagated to the form object even though we've set the `starts_on` value on the form object.
```

This implementation relies on the underlying `ClaimWindow` being populated alongside the `ClaimWindowForm` attributes.

This means setting `starts_on` on the claim window form object also needs to set it on the `claim_window` model.

```ruby
def starts_on=(value)
  super(value)

  claim_window.starts_on = value
end
```

We would need to do this for all of the form object's attributes. 🤔 

The alternative is to perform either a soft `persist` or split out the `persist` method into "attribute assignment" and then "save" in order to perform the validation between these two steps.

-----

A lot of these steps seem counterintuitive.

What do you think?

spec/forms/application_form_spec.rb
```diff
@@ -54,4 +54,110 @@
...
+      expect(example_form).to respond_to("date_of_birth(2i)=")
+      expect(example_form).to respond_to("date_of_birth(3i)=")
+
+      example_form.send("date_of_birth(1i)=", 2024)
+      example_form.send("date_of_birth(2i)=", 5)
+      example_form.send("date_of_birth(3i)=", 2)
+
+      expect(example_form.send("date_of_birth(1i)")).to eq(2024)
+      expect(example_form.send("date_of_birth(2i)")).to eq(5)
+      expect(example_form.send("date_of_birth(3i)")).to eq(2)
```
ollietreend, line 84: It would be worth using `public_send` to demonstrate that these are public methods. I think it's generally better to use `public_send` whenever possible, unless you really need to use `send`.

```suggestion
      example_form.public_send("date_of_birth(1i)=", 2024)
      example_form.public_send("date_of_birth(2i)=", 5)
      example_form.public_send("date_of_birth(3i)=", 2)

      expect(example_form.public_send("date_of_birth(1i)")).to eq(2024)
      expect(example_form.public_send("date_of_birth(2i)")).to eq(5)
      expect(example_form.public_send("date_of_birth(3i)")).to eq(2)
```

The same applies to the other examples in this spec.

spec/forms/application_form_spec.rb
```diff
@@ -54,4 +54,110 @@
...
+
+        expect(example_form.date_of_birth).to eq(Date.new(2024, 5, 2))
+      end
+
+      it "returns nil if any of the date components are missing" do
+        example_form.date_of_birth_year = 2024
+        example_form.date_of_birth_month = 5
+
+        expect(example_form.date_of_birth).to be_nil
+      end
```
ollietreend, line 160: We have a slightly different implementation in placements, which came about as a result of a content review done on the service.

A content designer fed back that partially entered dates – for example, if a user filled in the "day" and "month" fields, but left the "year" field blank – resulted in all three fields being reset following validation. That was happening because invalid dates were returned as `nil` for the entire attribute – i.e. exactly as this code does.

The content designer considered this behaviour to be a bug. Instead, the desired behaviour was for populated fields to be persisted, even if others were left blank.

For more details, see these artefacts:

- Trello: [Add mentor → preserve partially entered DOB fields](https://trello.com/c/c7hMIOEu/486-add-mentor-%E2%86%92-preserve-partially-entered-dob-fields)
- https://github.com/DFE-Digital/itt-mentor-services/pull/761

FWIW I believe this 'buggy' behaviour just wasn't noticed on the Claims side, but I think it's fair to expect consistent behaviour between the two services here.

Let's assume this behaviour is fine for now – after all, that's the existing behaviour in Claims, and I don't think it'd be in scope to change as part of this PR. However I think it would be good to follow up and standardise all our existing date attribute behaviour to use this concern, at which point the behaviour would need to change accordingly.

spec/system/claims/support/create_a_claim_window_spec.rb
```diff
@@ -54,14 +54,16 @@ def then_i_see_the_form_to_create_a_claim_window
...
   end
 
-  def when_i_fill_in_the_form(starts_on:, ends_on:)
-    fill_in "claims_claim_window[starts_on(3i)]", with: starts_on.day
-    fill_in "claims_claim_window[starts_on(2i)]", with: starts_on.month
-    fill_in "claims_claim_window[starts_on(1i)]", with: starts_on.year
+  def when_i_fill_in_the_form(starts_on:, ends_on:, academic_year:)
+    fill_in "claims_claim_window_form[starts_on(3i)]", with: starts_on.day
+    fill_in "claims_claim_window_form[starts_on(2i)]", with: starts_on.month
+    fill_in "claims_claim_window_form[starts_on(1i)]", with: starts_on.year
```
ollietreend, line 60: I think it would be cleaner to specify labels rather than HTML IDs in these specs. Ideally system specs should interact with the UI as if they were a user – e.g. looking for buttons and labels, rather than the underlying HTML IDs of attributes.

```suggestion
    within_fieldset("Starts on") do
      fill_in "Day", with: starts_on.day
      fill_in "Month", with: starts_on.month
      fill_in "Year", with: starts_on.year
    end
```

We went through a similar wave over in Placements where we changed IDs to text labels because we found it made for more readable specs – for example, see https://github.com/DFE-Digital/itt-mentor-services/pull/856#discussion_r1684554702.

app/forms/claims/claim_window_form.rb
```diff
@@ -0,0 +1,59 @@
...
+    claim_window.update!(starts_on:, ends_on:, academic_year_id:)
+  end
+
+  def new_back_path
+    new_claims_support_claim_window_path(claims_claim_window_form: slice(:starts_on, :ends_on, :academic_year_id))
+  end
+
+  def edit_back_path
+    edit_claims_support_claim_window_path(id, claims_claim_window_form: slice(:starts_on, :ends_on, :academic_year_id))
+  end
```
ollietreend, line 36: Knowledge of routing and path helpers feels like a responsibility for the Controller layer.

For example, this is how we've handled it in Placements:

https://github.com/DFE-Digital/itt-mentor-services/blob/c7a56194a1cc3e249e967a2456aabb407b5afba9/app/controllers/placements/schools/mentors/add_mentor_controller.rb#L37-L55

However I appreciate Form Objects are intrinsically coupled to the View layer, so these methods aren't necessarily introducing additional coupling.

Not a strong opinion – no change necessarily required.

app/forms/claims/claim_window_form.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    errors.add(:starts_on, :overlap)
+  end
+
+  def does_not_end_within_another_claim_window
+    return unless Claims::ClaimWindow.where.not(id:).where("starts_on <= :ends_on AND ends_on >= :ends_on", ends_on:).exists?
+
+    errors.add(:ends_on, :overlap)
+  end
+end
```
JamieCleare2525, line 59:
```suggestion
class Claims::ClaimWindowForm < ApplicationForm
  attr_accessor :id

  date_attribute :starts_on
  date_attribute :ends_on

  attribute :academic_year_id

  validates :starts_on, presence: true
  validates :ends_on, presence: true, comparison: { greater_than_or_equal_to: :starts_on }
  validates :academic_year_id, presence: true
  validate :validate_claim_window

  def initialize(attributes = {})
    super

    if id
      self.starts_on ||= claim_window.starts_on
      self.ends_on ||= claim_window.ends_on
      self.academic_year_id ||= claim_window.academic_year_id
    end
  end

  def persist
    assign_claim_window_attributes.save!
  end

  def new_back_path
    new_claims_support_claim_window_path(claims_claim_window_form: slice(:starts_on, :ends_on, :academic_year_id))
  end

  def edit_back_path
    edit_claims_support_claim_window_path(id, claims_claim_window_form: slice(:starts_on, :ends_on, :academic_year_id))
  end

  def academic_year_name
    AcademicYear.find_by(id: academic_year_id)&.name
  end

  private

  def claim_window
    @claim_window ||= id ? Claims::ClaimWindow.find(id) : Claims::ClaimWindow.new
  end

  def assign_claim_window_attributes
    @assign_claim_window_attributes ||= claim_window.tap do |window|
      window.starts_on = starts_on
      window.ends_on = ends_on
      window.academic_year_id = academic_year_id
    end
  end

  def validate_claim_window
    if assign_claim_window_attributes.invalid?
      assign_claim_window_attributes.errors.each do |err|
        next if err.attribute == :academic_year

        errors.add(err.attribute, err.message)
      end
    end
  end
end
```
I've changed your code so that the validation is removed and is dependant on the validation used on the model.

app/forms/claims/claim_window_form.rb
Nitemaeric: We'll leave this as is for now and revisit in the future if needed.

spec/forms/application_form_spec.rb
Nitemaeric: I've addressed this in the latest fixup.

CleanShot 2024-08-23 at 11 37 14

Practically speaking, it's a bit strange, which makes the implementation also strange.

Ideally, we'd have a partial-`Date`-like class that interacts with other `Date` instances.

This would allow us to avoid this code:

```ruby
if year.present? && month.present? && day.present?
  Date.new(year.to_i, month.to_i, day.to_i)
else
  IncompleteDate.new(year, month, day)
end
```

and just return...

```ruby
PotentiallyIncompleteDate.new(year, month, day)
```

`PotentiallyIncompleteDate` would behave exactly like a `Date` when it's components are complete but would respond to comparisons with other `Date` or `Date`-like objects, but return `false`.

`Date.new(2024, 5, 1) < PotentiallyIncompleteDate.new(2024, nil, 1) == false # always`

This would remove the need for this attribute method:

```ruby
define_method "#{attribute_name}_is_date?" do
  send(attribute_name).is_a?(Date)
end
```

and the usage of this method on all date-related validations (that are usually skipped when they're `nil`):

```ruby
validates :ends_on, comparison: { greater_than_or_equal_to: :starts_on }, if: -> { starts_on_is_date? && ends_on_is_date? }

validate :does_not_start_within_another_claim_window, if: :starts_on_is_date?
validate :does_not_end_within_another_claim_window, if: :ends_on_is_date?
```

Unfortunately, `#present?` returning `false` would be weird, so a `date_presence` validator, or similar, would be required to show "presence" errors without being blank.

```ruby
validates :starts_on, date_presence: true
validates :ends_on, date_presence: true
```

-----

To avoid investing too much time into creating a `Date`-interoperable class, I've added the attribute method and the conditional validations.

app/forms/claims/claim_window_form.rb
Nitemaeric: Do you mind if I revisit this in another PR? I agree with the point of leaning on model validations, as the `#persist` call would fail anyway. I'd like to give a good think about what the API can best be presented as.

JamieCleare2525 (approved):

PR #919, opened 2024-08-13
Add policy scopes

Kizr:
## Context

As our service expands it is important that we are scoping our data correctly based on who is accessing the application. This is a first pass to add some basic policy scopes to ensure the correct data is visible to the correct users.

- Adds an after callback to check for the presence of a policy scope for all placements index pages
- Adds multiple new policy scopes to enable this behaviour
- Updates index actions to make use of the new policy scopes
- Adds a missed translation which caused an error 😲 

## Guidance to review

Heavy smoke testing of _everything_ placements related. The tests should give some confidence, but test everything... manually... twice.

## Link to Trello card

[Add pundit policy scopes](https://trello.com/c/oOOUTNaY/660-add-pundit-policy-scopes)

### DIFF
--- a/app/policies/provider_policy.rb
+++ b/app/policies/provider_policy.rb
@@ -3,7 +3,7 @@ class ProviderPolicy < ApplicationPolicy
     def resolve
       return scope if user.support_user?
 
-      scope.where(id: user.current_organisation.partner_providers.ids)
+      scope.where(id: user.current_organisation.partner_providers.select(:id))
     end
   end
 end
--- a/app/policies/provider_policy.rb
+++ b/app/policies/provider_policy.rb
@@ -1,0 +1,9 @@
+class ProviderPolicy < ApplicationPolicy
+  class Scope < ApplicationPolicy::Scope
+    def resolve
+      return scope if user.support_user?
+
+      scope.where(id: user.current_organisation.partner_providers.ids)
+    end
+  end
+end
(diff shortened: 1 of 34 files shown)


spec/policies/placement_policy_spec.rb
```diff
@@ -44,4 +44,55 @@
...
+      end
+    end
+
+    context "when the user is a school user" do
+      let(:user) { create(:placements_user, schools: [school]) }
+      let(:school) { create(:placements_school) }
+
+      before do
+        user.current_organisation = school
+        school.placements << create(:placement)
```
JamieCleare2525, line 69:
```suggestion
        create(:placement, school: school)
```

spec/policies/user_membership_policy_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      create_list(:user_membership, 3)
+    end
+
+    context "when the user is a school user" do
+      let(:user) { create(:placements_user, schools: [school]) }
+      let(:school) { create(:placements_school) }
+
+      before do
+        user.current_organisation = school
+        school.user_memberships << create(:user_membership)
```
JamieCleare2525, line 19:
```suggestion
        create(:user_membership, school:)
```

JamieCleare2525 (approved):

app/policies/provider_policy.rb
```diff
@@ -0,0 +1,9 @@
+class ProviderPolicy < ApplicationPolicy
+  class Scope < ApplicationPolicy::Scope
+    def resolve
+      return scope if user.support_user?
+
+      scope.where(id: user.current_organisation.partner_providers.ids)
```
ollietreend, line 6: Let's use a subquery for these instead of causing extra queries:

```suggestion
      scope.where(id: user.current_organisation.partner_providers.select(:id))
```

ollietreend (approved):

PR #945, opened 2024-08-20
Add secondary navigation to Support claims page

Nitemaeric:
## Context

We want to add in the following flows:

- Payments (Downloading CSV of claims, uploading CSV of payments)
- Sampling (Uploading CSV of claims to be sampled)
- Clawback (Uploading CSV of claims to be clawed back)

This is the initial PR to add in the tabs, without adding the functional parts to keep future change sets to a minimum.

- This commit adds the "Payments", "Sampling", and "Clawback" tabs to the Claims page.

## Guidance to review

- Content will change in a future PR.
- This PR is a starting point to add in these flows in an incremental manner.

## Screenshots

CleanShot 2024-08-20 at 17 17 13

CleanShot 2024-08-20 at 17 17 26

### DIFF
--- a/app/views/claims/support/claims/_secondary_navigation.html.erb
+++ b/app/views/claims/support/claims/_secondary_navigation.html.erb
@@ -1,0 +1,8 @@
+<% if ENV["FEATURE_FLAG_CLAIMS_TABS"] == "true" %>
+  <%= render SecondaryNavigationComponent.new do |component| %>
+    <% component.with_navigation_item t(".claims"), claims_support_claims_path %>
+    <% component.with_navigation_item t(".payment"), claims_support_claims_payment_path %>
+    <% component.with_navigation_item t(".sampling"), claims_support_claims_sampling_path %>
+    <% component.with_navigation_item t(".clawback"), claims_support_claims_clawback_path %>
+  <% end %>
+<% end %>
--- a/config/routes/claims.rb
+++ b/config/routes/claims.rb
@@ -62,6 +62,14 @@ scope module: :claims, as: :claims, constraints: {
 
     resources :claims, only: %i[index show] do
       get :download_csv, on: :collection
+
+      if ENV["FEATURE_FLAG_CLAIMS_TABS"] == "true"
+        collection do
+          resource :payment, only: %i[show create], module: :claims, as: :claims_payment
+          resource :sampling, only: %i[show create], module: :claims, as: :claims_sampling
+          resource :clawback, only: %i[show create], module: :claims, as: :claims_clawback
+        end
+      end
     end
 
     resources :support_users, path: "support-users" do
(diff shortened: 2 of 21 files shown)


app/views/claims/support/claims/_secondary_navigation.html.erb
```diff
@@ -0,0 +1,8 @@
+<% unless HostingEnvironment.env.production? %>
```
JamieCleare2525, line 1: Have you considered using a feature flag instead of this environment check? I'm assuming this feature is actually something you want enabled at some point.

Nitemaeric: I did consider it, but it seemed like more effort to use a feature flag in the case of where we only want to prevent it from being accessible in production, not in any of the pre-production environments.

config/routes/claims.rb
```diff
@@ -56,6 +56,14 @@
 
     resources :claims, only: %i[index show] do
       get :download_csv, on: :collection
+
+      unless HostingEnvironment.env.production?
```
JamieCleare2525, line 60: Instead of using this environment check, you should think about using feature flag for each specific feature/route.

app/views/claims/support/claims/_secondary_navigation.html.erb
JamieCleare2525: Whilst it might be more effort this is the very purpose for feature flags.
It means that these 3 features can be developed and released independently of each other.
Otherwise, you'll need to complete the work for all 3 features before removing this `HostingEnvironment.env.production?` clause.

In my opinion, `HostingEnvironment.env.production?` should only be used if this feature is intended to never be used on production. Such as your database cleanup work.

Nitemaeric: 👍 I'll make the change.

Nitemaeric: I've gone ahead with using an ENV var as a feature flag to get around Flipflop's limitation of requiring a database connection.

JamieCleare2525: Whilst I understand your change to use ENV variables instead of feature flags. I would have expected a flag per feature (Sampling, Clawback, etc.). 
Is the intention onto to release these features to production only once all 3 features are complete?

Nitemaeric: Yes. The intention is to release all of these features all together.

That being said, our recent revisions have broken down the processes into distinct stages.

I'll remove Sampling and Clawback for now.

Nitemaeric:
Actually, I've decided to drop this PR.

As we're now moving to implement 1 flow at a time, we can add the tabs in as the flows are added.

PR #951, opened 2024-08-22
Create Placements::AcademicYear helper methods

dp-daly:
## Context

- We have new views that will allow users to see their placements for their previous academic year, current academic year and next academic year. 
- The UI should be unaware of the logic involved to calculate which academic years are shown.

- Add a current_academic_year method to `Placements::AcademicYear`.
- Add a next_academic_year method to `Placements::AcademicYear`.
- Add a previous_academic_year method to `Placements::AcademicYear`.
- Write new specs to cover this functionality.

## Guidance to review

- Run `bundle exec rspec spec/models/placements/academic_year_spec.rb`

## Link to Trello card

https://trello.com/c/ztMmmitf/703-create-nextacademicyear-and-previousacademicyear-methods-for-placementsacademicyear


app/models/placements/academic_year.rb
```diff
@@ -1,2 +1,24 @@
...
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
 class Placements::AcademicYear < AcademicYear
+  def self.current
+    for_date(Date.current)
+  end
+
+  def next
+    Placements::AcademicYear.for_date(starts_on + 1.year)
```
JamieCleare2525, line 18:
```suggestion
    for_date(starts_on + 1.year)
```

app/models/placements/academic_year.rb
```diff
@@ -1,2 +1,24 @@
...
+  end
+
+  def previous
+    Placements::AcademicYear.for_date(starts_on - 1.year)
```
JamieCleare2525, line 22:
```suggestion
    for_date(starts_on - 1.year)
```

spec/factories/placements/academic_years.rb
```diff
@@ -0,0 +1,20 @@
...
+# Table name: academic_years
+#
+#  id         :uuid             not null, primary key
+#  ends_on    :date
+#  name       :string
+#  starts_on  :date
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+FactoryBot.define do
```
JamieCleare2525, line 12: I would move this to the `spec/factories/academic_years.rb`, similar to the `spec/factories/users.rb`
```
factory :placements_academic_year, class: "Placements::AcademicYear", parent: :academic_year
```

spec/models/placements/academic_year_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      create(:placements_academic_year,
+             starts_on: Date.parse("1 September 2023"),
+             ends_on: Date.parse("31 August 2024"),
+             name: "2023 to 2024")
+    end
+
+    it "returns the previous academic year" do
+      expect(current_academic_year.previous).to eq(previous_academic_year)
+    end
+  end
```
JamieCleare2525, line 59: Could you also add tests for it the previous/next academic year does not exist, then the academic year is created.

app/models/placements/academic_year.rb
dp-daly: @JamieCleare2525 This is an instance method rather than a class method so tests break if `Placements::AcademicYear` is removed.

app/models/placements/academic_year.rb
dp-daly: @JamieCleare2525 This is an instance method rather than a class method so tests break if `Placements::AcademicYear` is removed.

JamieCleare2525 (approved):

PR #953, opened 2024-08-22
[Part 2] associate placements to placements academic year

Kizr:
## Context

Placements now have a concept of their academic year, this work allows us to reference that data from the placements model.

- Add the `belongs_to` association to the `Placement` for `Placements::AcademicYear`
- Adds the `has_many` association to the `Placements::AcademicYear` for `Placement`
- Converts the `null: true` to `null: false` for the `academic_year_id` column
- Adds the foreign key for the `academic_year_id` column
- Populates all new placements via the placements wizard with the current academic year

## Guidance to review

- Smoke test the app, there are no user facing changes.

## Link to Trello card

[Associate Placement to Placements::AcademicYear](https://trello.com/c/avhLfMpn/704-associate-placement-to-placementsacademicyear)

### DIFF
--- a/app/wizards/placements/add_placement_wizard.rb
+++ b/app/wizards/placements/add_placement_wizard.rb
@@ -33,7 +33,7 @@ module Placements
 
     def build_placement
       placement = school.placements.build
-      placement.academic_year = ::Placements::AcademicYear.current_academic_year
+      placement.academic_year = ::Placements::AcademicYear.current
       placement.subject = steps[:subject].subject
 
       if steps[:additional_subjects].present?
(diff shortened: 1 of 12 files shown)


app/wizards/placements/add_placement_wizard.rb
```diff
@@ -33,6 +33,7 @@ def create_placement
 
     def build_placement
       placement = school.placements.build
+      placement.academic_year = ::Placements::AcademicYear.current
```
JamieCleare2525, line 36: Can we add a comment that this will change

Kizr: Due to the fact that this enables placements to be created with the same functionality they had before. When this changes it is a new feature so there is no reason to call this out.

JamieCleare2525 (approved):

PR #994, opened 2024-09-03
Add mentors training type

Kizr:
## Context

Now we’re headed in to a new academic year, we need to introduce the concept of training types for mentor training.

- Initial training: The first time a mentor trains with a provider, they receive up to 20 hours of ‘initial training’.
- Refresher training: Then every year after that, the mentor is eligible for up to 6 hours of ‘refresher training’ with the provider.

- Back fills the `training_type` column on `Claims::MentorTraining` with "initial"
- Changes the `training_type` column to `null: false` and the default to `"initial"`
- Adds the `training_type` enum to the `Claims::MentorTraining` model
- Updates the specs to reflect these changes

## Guidance to review

- Check out this branch locally and run the migrations with some previously created claims records
- Create a new claim after running the migrations to make sure everything continues to work as expected

## Link to Trello card

[Add 'training type' to mentor training records](https://trello.com/c/qw8e2hb4/738-add-training-type-to-mentor-training-records)


app/models/claims/mentor_training.rb
```diff
@@ -31,6 +31,11 @@ class Claims::MentorTraining < ApplicationRecord
 
   audited associated_with: :claim
 
+  enum :training_type, {
+    initial: "initial",
+    refresher: "refresher",
+  }, default: :initial, validate: { allow_nil: false }
```
Nitemaeric, line 37:
```suggestion
  }, default: :initial, validate: true
```

This will validate that any value is one of the enum values, `nil` will be invalid. Ref: https://blog.saeloun.com/2024/01/02/rails-enums-validate-option/#after

db/migrate/20240903123622_back_fill_mentor_training_training_type.rb
```diff
@@ -0,0 +1,8 @@
+class BackFillMentorTrainingTrainingType < ActiveRecord::Migration[7.1]
+  def up
+    # We require callbacks for BigQuery to stay in sync with the data.
+    Claims::MentorTraining.find_each do |mentor_training|
+      mentor_training.update(training_type: Claims::MentorTraining.training_types[:initial])
+    end
+  end
+end
```
Nitemaeric, line 8: I believe we've decided to keep data migrations outside of schema migrations. That being said, you'll need to add the data migration in a separate PR to ensure that the `null` db values are cleared out before the validations are added.

Nitemaeric: I've had a conversation with @ollietreend about the removal of data migration. I'm all on board with the approach.

One thing important to me is the re-playability of migrations though. In the case here, we use application code within the migration file. As application code is prone to change, for example:

2 years from now, we decide to rename the `Claims::MentorTraining` model. This breaks the migration. Alternatively, we could change the `training_type` enum name or values. This could also break the migration.

The workaround is to avoid using application code from within migrations. Executing raw SQL is a sure way to ensure that no application code is used, but as a Rails developer, `ActiveRecord` is available for a reason, no? 😉 

```ruby
class SomeMigration < ActiveRecord::Migration[7.1]
  class MentorTrainingForMigration < ActiveRecord::Base # Note that this doesn't inherit from ApplicationRecord
    self.table_name = :mentor_trainings
  end

  def up
    MentorTrainingForMigration.update_all(training_type: :initial)
  end

  def down
    MentorTrainingForMigration.update_all(training_type: nil)
  end
end
```

As for pushing the updated values from our Prod database to DfE Analytics, let's perform a manual update following the guidance [here](https://github.com/DFE-Digital/dfe-analytics?tab=readme-ov-file#7-import-existing-data).

Better yet, you can tie this data migration to the schema change too.

```ruby
class ChangeMentorTrainingTrainingTypeDefault < ActiveRecord::Migration[7.1]
  class MentorTrainingForMigration < ActiveRecord::Base
    self.table_name = :mentor_trainings
  end

  def change
    change_column_default :mentor_trainings, :training_type, from: nil, to: "initial"

    up_only do
      MentorTrainingForMigration.where(training_type: nil).update_all(training_type: :initial)
    end
  end
end
```

Note: This would update the values when migrating forward, but leaves them as is when reverting.

https://github.com/testdouble/good-migrations is probably something we can add to prevent application code from slipping into migrations.

Kizr: Some great insights @Nitemaeric thank you, I've done similar previously when a model hasn't been available so this is a great idea 👍

db/migrate/20240903123307_change_mentor_training_training_type_default.rb
```diff
@@ -0,0 +1,16 @@
...
+    self.table_name = :mentor_trainings
+  end
+
+  def change
+    change_column_default :mentor_trainings, :training_type, from: nil, to: "initial"
+
+    up_only do
+      # We require callbacks for BigQuery to stay in sync with the data.
+      MentorTrainingForMigration.where(training_type: nil).find_each do |mentor_training|
+        mentor_training.update(training_type: Claims::MentorTraining.training_types[:initial])
```
Nitemaeric, line 12: Let's use a hardcoded string here. :)

```suggestion
        mentor_training.update(training_type: "initial")
```

Nitemaeric (approved):
🙌

ollietreend (approved):
Looks good to me 👍🏻

PR #995, opened 2024-09-03
Add status and type of establishment headers and outputs to CSV

dp-daly:
## Context

ESFA has requested the type_of_establishment of each school in order to assist lookup the bank details of a School. If a school belongs to a local authority, their payment will go there.

The status field is required for ESFA to confirm that they have received Claims that are all in the correct status.

- Add the type_of_establishment field from School the Claims::Claim::GenerateCSV output.
- Add the status of the Claim to the Claims::Claim::GenerateCSV service.

## Guidance to review

- Log in as a school user and create a new claim.
- Log out and log in as support user Colin.
- Click on the claims tab.
- Press 'Download CSV' in the top-left corner of the claims index page.
- Review the resulting CSV file.

## Link to Trello card

https://trello.com/c/drfnuwOA/741-add-typeofestablishment-and-status-to-claims-csv-download

## Screenshots

Screenshot 2024-09-03 at 14 53 05

### DIFF
--- a/app/services/claims/claim/generate_csv.rb
+++ b/app/services/claims/claim/generate_csv.rb
@@ -1,7 +1,7 @@
 require "csv"
 
 class Claims::Claim::GenerateCSV < ApplicationService
-  HEADERS = %w[claim_reference urn school_name local_authority claim_amount establishment_type date_submitted].freeze
+  HEADERS = %w[claim_reference urn school_name local_authority claim_amount type_of_establishment establishment_type date_submitted claim_status].freeze
 
   def initialize(claims:)
     @claims = claims
@@ -18,8 +18,10 @@ class Claims::Claim::GenerateCSV < ApplicationService
           claim.school.name,
           claim.school.local_authority_name,
           claim.amount.format(symbol: false, decimal_mark: ".", no_cents: false),
+          claim.school.type_of_establishment,
           claim.school.group,
           claim.submitted_at&.iso8601,
+          claim.status,
         ]
       end
     end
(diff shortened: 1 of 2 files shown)


app/services/claims/claim/generate_csv.rb
```diff
@@ -1,7 +1,7 @@
 require "csv"
 
 class Claims::Claim::GenerateCSV < ApplicationService
-  HEADERS = %w[claim_reference urn school_name local_authority claim_amount establishment_type date_submitted].freeze
+  HEADERS = %w[claim_reference urn school_name local_authority claim_amount type_of_establishment establishment_type date_submitted claim_status].freeze
```
ollietreend, line 4: @Nitemaeric can I just double check that this is correct?

There's already a column called `establishment_type`, and this PR adds a new column with a very similar name `type_of_establishment` but a different value.

| Header                  | Value                                |
|-------------------------|--------------------------------------|
| `type_of_establishment` | `claim.school.type_of_establishment` |
| `establishment_type`    | `claim.school.group`                 |

This is probably going to be a bit confusing unless ESFA are expecting both these columns.

ollietreend (commented):
Technically, this looks good to me. But I'll defer to @Nitemaeric to give the official ✅

app/services/claims/claim/generate_csv.rb
Nitemaeric: ESFA have specifically requested the `type_of_establishment` header as additional information.

As far as I'm aware, they're expecting an additional column to the existing pattern.

If it does cause any confusion moving forward, we can drop it easily enough :)

Nitemaeric (approved):
All is good to me! 🚀

Nitemaeric:
[As ESFA, I would like to the establishment type look up to the next level down](https://trello.com/c/XRVjK40d/652-as-esfa-i-would-like-to-the-establishment-type-look-up-to-the-next-level-down)

PR #999, opened 2024-09-04
DateAttributes should return nil when all date components are nil

Nitemaeric:
## Context

Date attributes are returning empty `IncompleteDate` objects when all date components are `nil`. This causes some strange behaviours, such as the the `Claims::ClaimWindowForm#initialize` method not setting the date attributes as they are not `nil`, they are `IncompleteDate` objects by default.

- A date attribute will return `nil` when all date components are `nil` instead of an `IncompleteDate` with all components as `nil`.

## Guidance to review

- Try editing a claim window.
- The date should now be pre-populated.

## Screenshots

| Before | After |
| ------ | ----- |
| CleanShot 2024-09-04 at 11 20 54 | CleanShot 2024-09-04 at 11 21 27 |

### DIFF
--- a/app/forms/concerns/date_attributes.rb
+++ b/app/forms/concerns/date_attributes.rb
@@ -30,9 +30,9 @@ module DateAttributes
       define_method attribute_name do
         year, month, day = %w[year month day].map { |date_part| send("#{attribute_name}_#{date_part}") }
 
-        if year.present? && month.present? && day.present?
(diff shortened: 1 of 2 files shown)


Nitemaeric:
[`DateAttributes` concern bug - date components not being populated](https://trello.com/c/X4FywN7K/247-dateattributes-concern-bug-date-components-not-being-populated)

Kizr (approved):

PR #1002, opened 2024-09-05
Add placements navigation

Kizr:
## Context

School users need to be able to filter their placments by academic year

- [x] Add academic year nav to placements index
- [x] Update links to placements index to default to current academic year 

## Guidance to review

- Log in as Anne
- View placements
- Click on "Next academic year"
- Add a placement
- Click on "Next academic year"
- Placement should be visible

## Link to Trello card

[Add academic year navigation](https://trello.com/c/wGqnm8Lt/778-add-academic-year-navigation)

## Screenshots

image
image

### DIFF
--- a/app/controllers/placements/schools/placements_controller.rb
+++ b/app/controllers/placements/schools/placements_controller.rb
@@ -69,6 +69,6 @@ class Placements::Schools::PlacementsController < Placements::ApplicationControl
   end
 
   def index_path
-    placements_school_placements_path(@school, params: { year: :current })
+    placements_school_placements_path(@school)
   end
 end
--- a/app/controllers/placements/schools/placements_controller.rb
+++ b/app/controllers/placements/schools/placements_controller.rb
@@ -2,12 +2,12 @@ class Placements::Schools::PlacementsController < Placements::ApplicationControl
   before_action :set_school
   before_action :set_placement, only: %i[show remove destroy preview]
   before_action :set_decorated_placement, only: %i[show remove preview]
+  before_action :current_academic_year, only: :index
 
   helper_method :edit_attribute_path, :add_provider_path, :add_mentor_path
 
   def index
-    scope = policy_scope(@school.placements)
-    @pagy, placements = pagy(scope.includes(:subject, :mentors, :additional_subjects, :provider).order("subjects.name"))
+    @pagy, placements = pagy(placements_scope.includes(:subject, :mentors, :additional_subjects, :provider).order("subjects.name"))
     @placements = placements.decorate
   end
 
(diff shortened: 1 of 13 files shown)


app/controllers/placements/schools/placements_controller.rb
```diff
@@ -2,12 +2,12 @@ class Placements::Schools::PlacementsController < Placements::ApplicationControl
   before_action :set_school
   before_action :set_placement, only: %i[show remove destroy preview]
   before_action :set_decorated_placement, only: %i[show remove preview]
+  before_action :current_academic_year, only: :index
```
ollietreend, line 5: Does this need to be a `before_action`?

```suggestion
```

I'm expecting it should be safe to remove because it's already called by `index` through this call chain:

```
index → placements_scope → academic_year_scope → current_academic_year
```

spec/system/placements/schools/placements/view_placements_list_spec.rb
```diff
@@ -141,4 +172,8 @@ def expect_to_see_table_headings
       expect(page).to have_content "Start date"
     end
   end
+
+  def and_i_click_on(link_text)
+    click_on link_text
+  end
```
ollietreend, line 178: Sorry to be nitpicky but for proper "Given, When, Then" style syntax, this should be "when I click on"

```suggestion
  def when_i_click_on(link_text)
    click_on link_text
  end
```

spec/system/placements/schools/placements/view_placements_list_spec.rb
```diff
@@ -53,6 +53,29 @@
...
+        then_i_see_placement(current_academic_year_placement)
+        and_i_do_not_see_placement(next_academic_year_placement)
+      end
+
+      scenario "when I view placements for the next academic year" do
+        given_i_sign_in_as_anne
+        and_i_click_on("Next year (#{next_academic_year.name})")
+        then_i_see_placement(next_academic_year_placement)
+        and_i_do_not_see_placement(current_academic_year_placement)
+      end
```
ollietreend, line 77: If I'm interpreting the system specs correctly, I think we test the paths:

- `/placements`
- `/placements?year=next`

But we never actually test:

- `/placements?year=current`

I was trying to work out why that doesn't result in a coverage failure. But it turns out it's because:

```ruby
def academic_year_scope
  params[:year] == "next" ? current_academic_year.next : current_academic_year
end
```

This code :point_up: only has two branches. Either `params[:year] == "next"` or it doesn't.

So we're implicitly covering it already. Although I do think it might be nice so explicitly click the "This year" tab just to prove that it works.

ollietreend: What do you think about something like this?

```ruby
scenario "I can switch between academic years" do
  given_i_sign_in_as_anne

  then_i_see_placement(current_academic_year_placement)
  and_i_do_not_see_placement(next_academic_year_placement)

  when_i_click_on("Next year (#{next_academic_year.name})")
  then_i_see_placement(next_academic_year_placement)
  and_i_do_not_see_placement(current_academic_year_placement)

  when_i_click_on("This year (#{current_academic_year.name})")
  then_i_see_placement(current_academic_year_placement)
  and_i_do_not_see_placement(next_academic_year_placement)
end
```

ollietreend (approved):

PR #1014, opened 2024-09-10
Create training allowance object

Kizr:
## Context

Create a class that will calculate how much training can be claimed for a given mentor.

- Adds a new `Claims::TrainingAllowance` model that combines the features of several existing classes into a single class.

## Guidance to review

Work through the spec file and check that all of the possible scenarios are covered, alternatively this can be tested via a rails console with existing claims data to confirm the expected results.

## Link to Trello card

[Introduce a Training Allowance object](https://trello.com/c/l2QqUCci/739-introduce-a-training-allowance-object)

### DIFF
--- a/app/models/claims/training_allowance.rb
+++ b/app/models/claims/training_allowance.rb
@@ -7,7 +7,7 @@ class Claims::TrainingAllowance
   end
 
   def training_type
-    @training_type ||= has_initial_training? ? REFRESHER_TRAINING_TYPE : INITIAL_TRAINING_TYPE
+    @training_type ||= has_trained_in_a_previous_year? ? REFRESHER_TRAINING_TYPE : INITIAL_TRAINING_TYPE
   end
 
   def total_hours
@@ -38,17 +38,20 @@ class Claims::TrainingAllowance
     @mentor_training_scope ||= provider.mentor_trainings
                                        .where(mentor:)
                                        .joins(claim: { claim_window: :academic_year })
-                                       .merge(Claims::Claim.not_draft_status)
   end
 
-  def has_initial_training?
-    mentor_training_scope.where(academic_years: { ends_on: ..academic_year.starts_on }).exists?
+  def has_trained_in_a_previous_year?
+    mentor_training_scope
+      .where(academic_years: { ends_on: ..academic_year.starts_on })
+      .merge(Claims::Claim.not_draft_status)
+      .exists?
   end
 
   def hours_completed
     mentor_training_scope
       .where(claims: { claim_windows: { academic_year: } })
       .where.not(claim_id: [claim_to_exclude&.id, claim_to_exclude&.previous_revision_id])
+      .merge(Claims::Claim.active)
       .sum(:hours_completed)
   end
 end
--- a/app/models/claims/training_allowance.rb
+++ b/app/models/claims/training_allowance.rb
@@ -1,8 +1,9 @@
 class Claims::TrainingAllowance
-  def initialize(mentor:, provider:, academic_year:)
+  def initialize(mentor:, provider:, academic_year:, claim_to_exclude: nil)
     @mentor = mentor
     @provider = provider
     @academic_year = academic_year
+    @claim_to_exclude = claim_to_exclude
   end
 
   def training_type
@@ -16,6 +17,9 @@ class Claims::TrainingAllowance
       6
     end
   end
+
+  def remaining_hours
+    total_hours - hours_completed
   end
 
   private
--- a/app/models/claims/training_allowance.rb
+++ b/app/models/claims/training_allowance.rb
@@ -11,7 +11,7 @@ class Claims::TrainingAllowance
   end
 
   def total_hours
-    training_type == INITIAL_TRAINING ? 20 : 6
+    training_type == INITIAL_TRAINING ? INITIAL_TRAINING_HOURS : REFRESHER_TRAINING_HOURS
   end
 
   def remaining_hours
@@ -27,9 +27,12 @@ class Claims::TrainingAllowance
   attr_reader :mentor, :provider, :academic_year, :claim_to_exclude
 
   INITIAL_TRAINING = :initial
+  INITIAL_TRAINING_HOURS = 20
   REFRESHER_TRAINING = :refresher
+  REFRESHER_TRAINING_HOURS = 6
 
-  private_constant :INITIAL_TRAINING, :REFRESHER_TRAINING
+  private_constant :INITIAL_TRAINING, :INITIAL_TRAINING_HOURS, :REFRESHER_TRAINING,
+                   :REFRESHER_TRAINING_HOURS
 
   def mentor_training_scope
     @mentor_training_scope ||= provider.mentor_trainings
--- a/app/models/claims/training_allowance.rb
+++ b/app/models/claims/training_allowance.rb
@@ -7,15 +7,11 @@ class Claims::TrainingAllowance
   end
 
   def training_type
-    has_initial_training? ? :refresher : :initial
+    has_initial_training? ? REFRESHER_TRAINING : INITIAL_TRAINING
   end
 
   def total_hours
-    if training_type == :initial
-      20
-    elsif training_type == :refresher
-      6
-    end
+    training_type == INITIAL_TRAINING ? 20 : 6
   end
 
   def remaining_hours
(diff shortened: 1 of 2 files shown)


app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,51 @@
...
+    @academic_year = academic_year
+    @claim_to_exclude = claim_to_exclude
+  end
+
+  def training_type
+    has_initial_training? ? REFRESHER_TRAINING : INITIAL_TRAINING
+  end
+
+  def total_hours
+    training_type == INITIAL_TRAINING ? 20 : 6
```
ollietreend, line 14: Can you make `20` and `6` constants too, please?

```suggestion
    training_type == INITIAL_TRAINING ? INITIAL_TRAINING_HOURS : REFRESHER_TRAINING_HOURS
```

app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  private
+
+  attr_reader :mentor, :provider, :academic_year, :claim_to_exclude
+
+  INITIAL_TRAINING = :initial
+  REFRESHER_TRAINING = :refresher
+
+  private_constant :INITIAL_TRAINING, :REFRESHER_TRAINING
```
ollietreend, line 32: ❤️

app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,54 @@
...
+    has_initial_training? ? REFRESHER_TRAINING : INITIAL_TRAINING
+  end
+
+  def total_hours
+    training_type == INITIAL_TRAINING ? INITIAL_TRAINING_HOURS : REFRESHER_TRAINING_HOURS
+  end
+
+  def remaining_hours
+    total_hours - hours_completed
+  end
```
ollietreend, line 19: Likewise, this method also performs an expensive query.

```suggestion
  def remaining_hours
    @remaining_hours ||= total_hours - hours_completed
  end
```

ollietreend (commented):
Just a couple of comments to optimise things.

It would also be really good to get @Nitemaeric to review the logic to make sure it checks out & no edge cases have been missed.

app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,54 @@
...
+                   :REFRESHER_TRAINING_HOURS
+
+  def mentor_training_scope
+    @mentor_training_scope ||= provider.mentor_trainings
+                                       .where(mentor:)
+                                       .joins(claim: { claim_window: :academic_year })
+  end
+
+  def has_initial_training?
+    mentor_training_scope.where(claims: { status: :submitted, claim_windows: { academic_years: { ends_on: ..academic_year.starts_on } } }).exists?
```
ollietreend, line 44: We need to be careful about hardcoding `status: :submitted` here. Dan is currently in the process of adding new statuses which presumably need to be included in this query – for example, see #1027.

I wonder if we'd be better off excluding the statuses we know should be ignored:

```ruby
.where.not(claims: { status: %i[draft internal_draft] })
```

But then I'm not entirely confident that every other status should be considered as valid in this context.

@Nitemaeric can you advise which statuses to filter out here?

I also wonder if there's a meaningful grouping for these statuses – for example, could they be defined as a scope on the `Claim` model, or even an array defined as a constant? Or is this only relevant here in this particular class?

app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def hours_completed
+    mentor_training_scope
+      .where(claims: { status: :submitted, claim_windows: { academic_year: } })
```
ollietreend, line 49: Ditto.

ollietreend (changes requested):
Now that #1027 has been merged in, I think we need to adjust these query conditions.

app/models/claims/training_allowance.rb
Nitemaeric: Yeah, I'd typically urge towards explicitly listing the statuses, rather than excluding them.

But in this case, the list of statuses isn't complete.

For statuses used for calculations, anything that gets to the "submitted" status and beyond would all be valid.

So in this instance, the code block you've suggested works.

app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,54 @@
...
+  REFRESHER_TRAINING_HOURS = 6
+
+  private_constant :INITIAL_TRAINING, :INITIAL_TRAINING_HOURS, :REFRESHER_TRAINING,
+                   :REFRESHER_TRAINING_HOURS
+
+  def mentor_training_scope
+    @mentor_training_scope ||= provider.mentor_trainings
+                                       .where(mentor:)
+                                       .joins(claim: { claim_window: :academic_year })
+  end
```
ollietreend, line 41: If we add `.merge(Claims::Claim.not_draft_status)`...

```suggestion
  def mentor_training_scope
    @mentor_training_scope ||= provider.mentor_trainings
                                       .where(mentor:)
                                       .joins(claim: { claim_window: :academic_year })
                                       .merge(Claims::Claim.not_draft_status)
  end
```

app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def has_initial_training?
+    mentor_training_scope.where(claims: { claim_windows: { academic_years: { ends_on: ..academic_year.starts_on } } }).where.not(claims: { status: %i[draft internal_draft] }).exists?
+  end
```
ollietreend, line 45: ...then we could drop this hardcoded condition...

```suggestion
  def has_initial_training?
    mentor_training_scope.where(claims: { claim_windows: { academic_years: { ends_on: ..academic_year.starts_on } } }).exists?
  end
```

app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def hours_completed
+    mentor_training_scope
+      .where(claims: { claim_windows: { academic_year: } }).where.not(claims: { status: %i[draft internal_draft] })
+      .merge(Claims::Claim.active)
+      .where.not(claim_id: [claim_to_exclude&.id, claim_to_exclude&.previous_revision_id])
+      .sum(:hours_completed)
+  end
```
ollietreend, line 53: ...and here.

```suggestion
  def hours_completed
    mentor_training_scope
      .where(claims: { claim_windows: { academic_year: } })
      .where.not(claim_id: [claim_to_exclude&.id, claim_to_exclude&.previous_revision_id])
      .sum(:hours_completed)
  end
```

spec/models/claims/training_allowance_spec.rb
```diff
@@ -0,0 +1,168 @@
+require "rails_helper"
+
+RSpec.describe Claims::TrainingAllowance, type: :model do
```
ollietreend, line 3: I think we might be missing scenarios where there are `internal_draft` or `draft` claims which need to be ignored.

I don't think those status are currently covered in this spec. 👀

ollietreend (commented):
Just a few more thoughts on how we handle claims with various different statuses.

1. How about promoting the Claim status condition up into the `mentor_training_scope` method? I think we need the status as a blanket condition across everything in this class.

2. It looks like there's [an existing scope](https://github.com/DFE-Digital/itt-mentor-services/blob/e09f77b66196b7377e0e455b7240913f4744ef48/app/models/claims/claim.rb#L66) we can use on the Claim model to set the condition, rather than needing to hardcode the values ourselves.

ollietreend (approved):
Great work! 🎉

Chef kiss emoji

app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  private_constant :INITIAL_TRAINING_TYPE, :INITIAL_TRAINING_HOURS, :REFRESHER_TRAINING_TYPE,
+                   :REFRESHER_TRAINING_HOURS
+
+  def mentor_training_scope
+    @mentor_training_scope ||= provider.mentor_trainings
+                                       .where(mentor:)
+                                       .joins(claim: { claim_window: :academic_year })
+                                       .merge(Claims::Claim.not_draft_status)
+  end
```
ollietreend, line 42:
```suggestion
  def mentor_training_scope
    @mentor_training_scope ||= provider.mentor_trainings
                                       .where(mentor:)
                                       .joins(claim: { claim_window: :academic_year })
  end
```

app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def has_initial_training?
+    mentor_training_scope.where(academic_years: { ends_on: ..academic_year.starts_on }).exists?
+  end
```
ollietreend, line 46: Only look at training on claims **which were submitted**. This means ignore `internal_draft` and `draft` statuses.

```suggestion
  def has_initial_training?
    mentor_training_scope
      .where(academic_years: { ends_on: ..academic_year.starts_on })
      .merge(Claims::Claim.not_draft_status)
      .exists?
  end
```

### Maybe tweak the name?

Also, _maybe_ a more descriptive / less ambiguous name for the method would be `has_trained_in_a_previous_year?`. That would avoid any ambiguity because "initial training" would usually mean `mentor_training.where(training_type: :initial)` (which in this case it's not).

app/models/claims/training_allowance.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def hours_completed
+    mentor_training_scope
+      .where(claims: { claim_windows: { academic_year: } })
+      .where.not(claim_id: [claim_to_exclude&.id, claim_to_exclude&.previous_revision_id])
+      .sum(:hours_completed)
+  end
```
ollietreend, line 53: Look at training on all claims **except for `internal_draft`**. We should _include_ `draft` claims in the hours completed calculation (unlike on `has_initial_training?`).

```suggestion
  def hours_completed
    mentor_training_scope
      .where(claims: { claim_windows: { academic_year: } })
      .where.not(claim_id: [claim_to_exclude&.id, claim_to_exclude&.previous_revision_id])
      .merge(Claims::Claim.active)
      .sum(:hours_completed)
  end
```

ollietreend (commented):
After clarifying the expected behaviour with @Nitemaeric, let's make these final changes to the query conditions and then merge in the PR.

We'll pick up integration into the wider codebase in the next card: [Refactor the claim form to use Training Allowance](https://trello.com/c/mSePcH76/740-refactor-the-claim-form-to-use-training-allowance).

ollietreend (approved):

PR #1018, opened 2024-09-11
Use sub-queries in policy scopes

Kizr:
## Context

The joins in our policy scopes were returning duplicate records unintentioanlly, we've moved to subqueries to avoid this.

- [x] Refactor the `user_policy.rb` scope to use a subquery.
- [x] Refactor the `mentor_policy.rb` scope to use a subquery.

## Guidance to review

- There should be no discernable differences to users
- Test suite should pass

## Link to Trello card

[Duplicates are appearing in the users list for schools and providers](https://trello.com/c/5Apc51qs/781-duplicates-are-appearing-in-the-users-list-for-schools-and-providers)

### DIFF
--- a/app/policies/placements/mentor_policy.rb
+++ b/app/policies/placements/mentor_policy.rb
@@ -3,13 +3,7 @@ class Placements::MentorPolicy < ApplicationPolicy
     def resolve
       return scope if user.support_user?
 
-      scope.where(
-        mentor_memberships: { school: user.current_organisation },
-      ).or(
-        scope.where(
-          id: Placements::MentorMembership.select(:mentor_id).where(school: user.current_organisation),
-        ),
-      )
+      scope.where(id: Placements::MentorMembership.select(:mentor_id).where(school: user.current_organisation))
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/policies/placements/mentor_policy.rb
```diff
@@ -3,7 +3,13 @@ class Scope < ApplicationPolicy::Scope
...
       return scope if user.support_user?
 
-      scope.joins(:mentor_memberships).where(mentor_memberships: { school: user.current_organisation })
+      scope.where(
+        mentor_memberships: { school: user.current_organisation },
+      ).or(
+        scope.where(
+          id: Placements::MentorMembership.select(:mentor_id).where(school: user.current_organisation),
+        ),
+      )
```
ollietreend, line 12: Why do we have an `OR` condition in this query now? 🤔 Shouldn't it work with just the subquery?

Right now it looks like we're doubling up with the same `WHERE` conditions twice.

```suggestion
      scope.where(
        id: Placements::MentorMembership.select(:mentor_id).where(school: user.current_organisation),
      )
```

ollietreend (approved):
Looks good to me 👍🏻

PR #1025, opened 2024-09-13
Add line break opportunity for support email addresses

Kizr:
## Context

Long email addresses cause horizontal scrolling when zoomed in. This is because the browser doesn’t break words across multiple lines by default, so it instead introduces x-overflow.

- [x] Adds line break opportunities for support email addresses in the application, this does not include emails.

## Guidance to review

- This one is a bit tricky to review, you can have a look through the review app to confirm but not all of them are easily accessible.

## Link to Trello card

[Add word breaks to email links](https://trello.com/c/Cb8MJNUZ/786-add-word-breaks-to-email-links)

### DIFF
--- a/app/views/claims/pages/terms.en.html.erb
+++ b/app/views/claims/pages/terms.en.html.erb
@@ -38,7 +38,7 @@
         </p>
 
         <p class="govuk-body">
-          You can contact us using the following email address: <%= govuk_mail_to t("claims.support_email") %>
+          You can contact us using the following email address: <%= govuk_mail_to t("claims.support_email_html") %>
         </p>
       </section>
 
--- a/app/views/claims/pages/terms.en.html.erb
+++ b/app/views/claims/pages/terms.en.html.erb
@@ -38,7 +38,7 @@
         </p>
 
         <p class="govuk-body">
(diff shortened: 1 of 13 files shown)


app/views/claims/pages/terms.en.html.erb
```diff
@@ -38,7 +38,7 @@
         </p>
 
         <p class="govuk-body">
-          You can contact us using the following email address: <%= govuk_mail_to t("claims.support_email") %>
+          You can contact us using the following email address: <%= govuk_mail_to t("claims.support_email_html") %>
```
Nitemaeric, line 41: This doesn't work. The display and mailto address need to be distinct, unfortunately.

There are a couple more like this in the PR.

Kizr: Ah yes, you're right! I tried playing with the `govuk_mail_to` method but it doesn't allow me to use it in the same way as the `mail_to` method does, I meant to switch all of these back to the original but I've clearly missed some, nice spot!

ollietreend (approved):

Nitemaeric (approved):

PR #1031, opened 2024-09-16
Add Ruby LSP RSpec addon

ollietreend:
Enables support for running RSpec tests inside your IDE (VS Code, in my case).

See [ruby-lsp-rspec][1] for an example of how this works.

Can be used in combination with the [Ruby LSP extension for VS Code][2].

[1]: https://github.com/st0012/ruby-lsp-rspec
[2]: https://marketplace.visualstudio.com/items?itemName=Shopify.ruby-lsp


Nitemaeric (approved):
Doesn't break Zed. 😃

PR #1041, opened 2024-09-17
Remove the tag from the reset database page title

Nitemaeric:
## Context

We don't want to have tags within page titles.

- Swap out tag for plaintext.
- Add "environment" word in for clarity.

## Guidance to review

- Reset database page no longer shows a tag in the page title.

### DIFF
--- a/app/views/claims/support/databases/reset.en.html.erb
+++ b/app/views/claims/support/databases/reset.en.html.erb
@@ -7,7 +7,7 @@
 
 <div class="govuk-width-container">
   <span class="govuk-caption-l">Reset database</span>
-  <h1 class="govuk-heading-l">Are you sure you want to reset the <%= govuk_tag text: hosting_environment_phase(current_service), colour: hosting_environment_color, classes: "govuk-tag--middle" %> database?</h1>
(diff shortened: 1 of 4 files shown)


app/views/claims/support/databases/reset.en.html.erb
```diff
@@ -7,7 +7,7 @@
 
 <div class="govuk-width-container">
   <span class="govuk-caption-l">Reset database</span>
-  <h1 class="govuk-heading-l">Are you sure you want to reset the <%= govuk_tag text: hosting_environment_phase(current_service), colour: hosting_environment_color, classes: "govuk-tag--middle" %> database?</h1>
```
ollietreend, line 10: Can we also tidy up the CSS that was created for this?

https://github.com/DFE-Digital/itt-mentor-services/blob/76d9faace9b8b8147751d8e3f5aa8e54f88c0f85/app/assets/stylesheets/components/_tag.scss#L1-L5

We should be able to drop the entire `components/_tag.scss` file.

Nitemaeric: Done! :D

ollietreend (approved):

PR #1046, opened 2024-09-18
Add missing `visually_hidden_text` to change links

Kizr:
## Context

For all “Change” links in the app (i.e. across both services), make sure visually_hidden_text is present.

Re-use the ‘key’ text as the hidden text.

For example, if the summary list key is “First name” then the visually hidden text should also be “First name”. That will produce the link “Change First name” which means there is context for screen reader users.

- [x] Adds visually hidden text to all change links that did not have it.

## Guidance to review

- Have a look at the updated views in the review app and inspect the HTML to see the visually hidden text

## Link to Trello card

[Add visually hidden text to all "change" links](https://trello.com/c/HQnpex17/788-add-visually-hidden-text-to-all-change-links)

### DIFF
--- a/app/views/claims/support/schools/mentors/check.html.erb
+++ b/app/views/claims/support/schools/mentors/check.html.erb
@@ -35,7 +35,8 @@
                                href: new_claims_support_school_mentor_path(params: mentor_form.as_form_params),
                                html_attributes: {
                                  class: "govuk-link--no-visited-state",
-                               }) %>
+                               },
+                               visually_hidden_text: t(".trn")) %>
           <% end %>
           <% summary_list.with_row do |row| %>
             <% row.with_key(text: t(".date_of_birth")) %>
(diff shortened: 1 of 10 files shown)


app/views/claims/support/schools/mentors/check.html.erb
```diff
@@ -44,7 +45,8 @@
                                href: new_claims_school_mentor_path(params: mentor_form.as_form_params),
                                html_attributes: {
                                  class: "govuk-link--no-visited-state",
-                               }) %>
+                               },
+                               visually_hidden_text: t(".trn")) %>
```
ollietreend, line 49:
```suggestion
                               visually_hidden_text: t(". date_of_birth")) %>
```

ollietreend (commented):
Looks good 👍🏻 nice work.

ollietreend (approved):

PR #1047, opened 2024-09-19
Replace hardcoded link with `:back`

Kizr:
## Context

There’s a bug where the ‘Back’ link on the placement details page doesn’t return you to the correct tab on the placements index page.

This applies to school users where they have ‘Current year’ and ‘Next year’ tabs for switching between placements in different academic years.

- [x] Made use of the `:back` helper

## Guidance to review

- Log in as Anne
- Navigate to placements
- Select "Next year (2025 to 2026)"
- View a placement
- Click back
- "Next year (2025 to 2026)" should still be selected

## Link to Trello card

[School user → Placements → Placement details: Make sure the 'back' link retains the currently selected academic year tab
](https://trello.com/c/hbZ4f3rH/796-school-user-%E2%86%92-placements-%E2%86%92-placement-details-make-sure-the-back-link-retains-the-currently-selected-academic-year-tab)

### DIFF
--- a/app/models/placements/academic_year.rb
+++ b/app/models/placements/academic_year.rb
@@ -16,6 +16,10 @@ class Placements::AcademicYear < AcademicYear
     for_date(Date.current)
   end
 
+  def current?
+    self == self.class.current
+  end
+
   def next
     Placements::AcademicYear.for_date(starts_on + 1.year)
   end
--- a/spec/system/placements/schools/placements/navigate_placements_spec.rb
+++ b/spec/system/placements/schools/placements/navigate_placements_spec.rb
@@ -1,0 +1,79 @@
+require "rails_helper"
+
+RSpec.describe "Placements / Schools / Placements / Navigate placements",
+               service: :placements, type: :system do
+  let!(:academic_year) { Placements::AcademicYear.current }
+  let!(:school) { create(:placements_school, name: "School 1", phase: "Primary") }
+  let!(:placement_for_current_academic_year) { create(:placement, school:).decorate }
+  let!(:placement_for_next_academic_year) { create(:placement, school:, academic_year: academic_year.next).decorate }
+
+  before do
+    given_i_sign_in_as_anne
+  end
+
+  context "when viewing placements from the current academic year" do
+    scenario "User can view a placement and return to the placements page" do
+      when_i_visit_the_placements_page
+      then_i_see_the_placements_page_for(academic_year)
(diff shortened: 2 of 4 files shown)


ollietreend (approved):
This fix looks good.

It'd be nice to have a system spec which covers navigation from the placements index page → show page → index page, returning with the correct tab selected.

However I've had a look at the existing system specs, and it seems like we don't actually test that journey at all right now. We only test [the index page](https://github.com/DFE-Digital/itt-mentor-services/blob/f829b6e7c6d2f8439a6104fc92588d648057e6dd/spec/system/placements/schools/placements/view_placements_list_spec.rb) and [the show page](https://github.com/DFE-Digital/itt-mentor-services/blob/1a79a43555222247f22a952a477c75c277a5bad9/spec/system/placements/schools/placements/view_a_placement_spec.rb) individually. ☹️

I'd say that's probably out of scope for this PR, but I think it'd be a nice thing to add coverage for in a future PR. What do you think @Kizr?

app/models/placements/academic_year.rb
```diff
@@ -11,11 +11,16 @@
...
 class Placements::AcademicYear < AcademicYear
   has_many :placements
+  delegate :current, to: :class
 
   def self.current
     for_date(Date.current)
   end
 
+  def current?
+    self == current
```
ollietreend, line 21:
```suggestion
    self == self.class.current
```

app/models/placements/academic_year.rb
```diff
@@ -11,11 +11,16 @@
 #
 class Placements::AcademicYear < AcademicYear
   has_many :placements
+  delegate :current, to: :class
```
ollietreend, line 14:
```suggestion
```

spec/system/placements/schools/placements/navigate_placements_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    and_i_visit_the_sign_in_path
+    and_i_click_sign_in
+  end
+
+  def when_i_visit_the_placements_page
+    visit placements_school_placements_path(school)
+  end
+
+  def then_i_see_the_placements_page_for(academic_year)
+    placement = academic_year.current? ? placement_for_current_academic_year : placement_for_next_academic_year
```
ollietreend, line 63: Might be worth adding this too, to prove that the correct tab is highlighted.

```ruby
find("a[aria-current=page]", text: "This year")
```

PR #1062, opened 2024-09-25
Refactor claim form to use training allowance

Kizr:
## Context

We have combined these services into a new unified service that handles all queries regarding training allowance.

- [x] Changes the base scope for `Claims::TrainingAllowance` to use `mentor` instead of `provider`
- [x] Replaces the existing use of services for calculating training hours
- [x] Removes the now redundant services

## Guidance to review

- There should be no discerable difference to users for this pull request, however @Nitemaeric with his additional knowledge of the Claims service should do some testing on the review app to make sure the specs haven't missed anything!

## Link to Trello card

[Refactor the claim form to use Training Allowance
](https://trello.com/c/mSePcH76/740-refactor-the-claim-form-to-use-training-allowance)


Nitemaeric:
@Kizr Yeah, I've found a bug. Here are the steps to reproduce:

- Sign in as Support User.
- Create a claim for an organisation with 4 (arbitrary) hours of training for a mentor.
  - 4 is just easy to verify. 
- Go to change the hours. (Only drafts made by support users can be changed).
- The allowance shows 16 hours left, when it should be 20, as you're editing a claim, not making a new one - The 4 hours from the existing claim should be re-added to the allowance.

app/forms/claims/claim/mentor_training_form.rb
```diff
@@ -62,11 +60,11 @@ def custom_hours_selected?
...
 
   def max_hours
-    @max_hours ||= Claims::Mentor::CalculateRemainingMentorTrainingHoursForProvider.call(mentor:, provider:, mentor_training:)
+    @max_hours ||= training_allowance.remaining_hours
   end
 
   def max_hours_equals_maximum_claimable_hours?
-    max_hours == MAXIMUM_CLAIMABLE_HOURS_PER_PROVIDER
+    max_hours == training_allowance.total_hours
   end
```
ollietreend, line 68: I appreciate this is probably out of scope for this PR, but I wanted to point it out nonetheless. I'm not a fan of these method names. I think we could improve upon them.

### `#max_hours`

`max_hours` is just `remaining_hours` under a different name. It doesn't seem to be a meaningful or intentional change in terminology – in fact, I think it's unnecessarily confusing.

I'd suggest we just delegate this method to the underlying training allowance object and avoid the name change altogether. Especially because the underlying `remaining_hours` method is already memoized – there's no reason _not to_ delegate it.

```ruby
delegate :remaining_hours, to: :training_allowance
```

### `#max_hours_equals_maximum_claimable_hours?`

This method name seems doubly confusing – _"What's the difference between 'max hours' and 'maximum claimable hours'? They sound like the same thing. And what does it mean if they're equal?"_.

This method is **only used in one place** in the app. It's used to decide whether this component gets rendered:

https://github.com/DFE-Digital/itt-mentor-services/blob/2bc7df6e8a739e45961c27bc46104e78cc634740/app/components/claims/claim/mentor_training_form/disclaimer_component.rb#L17-L19

I'd suggest a more meaningful method would be `some_hours_already_claimed?`, to indicate that some hours have already been claimed – in other words, because the 'remaining hours' is less than 'total hours' of the training allowance. Which would be a 'positive' use of logic in the component (i.e. we wouldn't need to `!` flip the boolean value):

```diff
def render?
-  !mentor_training_form.max_hours_equals_maximum_claimable_hours?
+  mentor_training_form.some_hours_already_claimed?
end
```

Additionally, I'm not _convinced_ this method _really belongs_ in the form object. Since the the component is the only thing that uses the method, is there really a legitimate need for it elsewhere?

If it's needed elsewhere, it should probably live directly on the `Claims::TrainingAllowance` object.

If it's only needed by the component itself, then the logic should probably just live directly inside the `#render?` method.

---

These would be nice improvements to make to the code, but as I said I appreciate they might not be technically in scope for this PR.

app/models/claims/training_allowance.rb
```diff
@@ -35,8 +35,8 @@ def available?
                    :REFRESHER_TRAINING_HOURS
 
   def mentor_training_scope
-    @mentor_training_scope ||= provider.mentor_trainings
-                                       .where(mentor:)
+    @mentor_training_scope ||= mentor.mentor_trainings
+                                       .where(provider:)
                                        .joins(claim: { claim_window: :academic_year })
```
ollietreend, line 40: Doesn't this produce the exact same query? 🤔 Any reason why it changed in this PR?

```ruby
provider.mentor_trainings.where(mentor:)
# vs
mentor.mentor_trainings.where(provider:)
```

No strong preference as to which is better. If you prefer the newer version, then we can keep it! 🙂

app/views/claims/support/schools/claims/rejected.html.erb
```diff
@@ -5,7 +5,7 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <span class="govuk-caption-l"> <%= t(".add_claim") %></span>
-      <h1><%= t(".page_title") %></h1>
+      <h1 class="govuk-heading-l"><%= t(".page_title") %></h1>
```
ollietreend, line 8: Nice spot

spec/system/claims/schools/claims/edit_draft_claim_spec.rb
ollietreend, line 1: Nice one ⭐

app/models/claims/training_allowance.rb
Kizr: There's an error raised when starting from `provider` due to the STI (If I recall correctly) but this is mitigated by using `mentor`, seemed the easiest solution for the issue

ollietreend: 🤨 How odd. I've just dug into it and it looks like a couple of misconfigured factories.

I'll push up a commit to fix them, and then tests pass with the code remaining as before.

ollietreend: Resolved by d033117cb2b5a93e44386e2e8694c24468f1a6bd

ollietreend (approved):
I'm happy with these changes.

@Nitemaeric can you please give the final ✅ before we merge this in?

app/forms/claims/claim/mentor_training_form.rb
ollietreend: Follow-up PR created to address this #1086

app/forms/claims/claim/mentor_training_form.rb
```diff
@@ -78,4 +76,13 @@ def mentor_trainings
   def previous_mentor_training
     mentor_trainings[mentor_trainings.index(mentor_training) - 1]
   end
+
+  def training_allowance
+    @training_allowance ||= Claims::TrainingAllowance.new(
+      mentor:,
+      provider:,
+      academic_year: Claims::ClaimWindow.current.academic_year,
```
Nitemaeric, line 84: Technically speaking, this _could_ be incorrect, right?

Excluding the external constraints, being unable to edit/submit a claim when the window is closed, it would be better to contain the responsibilities of those modules, right?

Nitemaeric (commented):
Just 1 comment. All else looks great!

app/forms/claims/claim/mentor_training_form.rb
ollietreend: @Nitemaeric resolved by 0b2ea03cb41652c5a67a508e548c2cf4a7b41c07 🙂

app/queries/claims/mentors_with_remaining_claimable_hours_query.rb
```diff
@@ -9,9 +9,13 @@ def call
...
-      Claims::Mentor::CalculateRemainingMentorTrainingHoursForProvider.call(mentor:, provider: params[:provider], claim: params[:claim]).positive?
+      training_allowance_for(mentor).available?
     end
 
     scope.where(id: mentor_with_claimable_hours)
   end
+
+  def training_allowance_for(mentor)
+    Claims::TrainingAllowance.new(mentor:, provider: params[:provider], academic_year: Claims::ClaimWindow.current.academic_year)
+  end
```
Nitemaeric, line 20: I think this also needs to follow the same pattern as changed above.

`claim: params[:claim]` in this case.

ollietreend: Resolved in 27493236d162f01f71c911d458302096f86fc0ee

Nitemaeric (approved):
Nice! Thanks for making the changes!

PR #1063, opened 2024-09-25
Update validation rules to support claiming for refresher training

Kizr:
## Context

This PR completes the implementation of functionality required for claiming refresher training. It's just the tip of the iceberg, and it follows on from #994, #1014, #1166, #1167, and #1168.

After this PR is merged, all the required logic and validation should be in place to support users claiming up to 20 hours of initial training, or up to 6 hours of refresher training, depending on whether or not the mentor has trained with that particular provider before.

> [!NOTE]  
> **There are still areas throughout the app which refer to a 20 hour limit of training.** This content will need updating ahead of refresher training becoming available to end users. This PR completes the development work required to support refresher training, but it doesn't update any content.

- **Make the validation rule for `Claims::MentorTraining#hours_completed` dynamic**
  So it's limited to either 20 hours, or 6 hours, depending on the training type. This removes the hardcoded references to 20 hours in favour of using the Training Allowance calculator, further solidifying that as the source of truth for calculating the number of claimable hours.
- **Store the training type in the Mentor Training record**
  Updates the `Claims::MentorTraining#training_type` field so it gets persisted to the database correctly. Before this, it would always persist to the database as "initial" training. This could be useful in future when looking back on historic claims.
- **Add a system spec for refresher training**
  To provide confidence that the service is now ready to accept claims for refresher training.

## Guidance to review

- Read through the spec and make sure the proposed expectations cover the scenarios

## Link to Trello card

https://trello.com/c/TUWOipzZ/824-add-test-coverage-for-refresher-training-and-remove-some-hardcoded-training-allowance-limits

### DIFF
--- a/app/forms/claims/claim/mentors_form.rb
+++ b/app/forms/claims/claim/mentors_form.rb
@@ -53,7 +53,7 @@ class Claims::Claim::MentorsForm < ApplicationForm
         mentor_id:,
         provider_id: claim.provider_id,
         claim_id: claim.id,
-      )
+      ).tap(&:set_training_type)
     end
   end
 end
--- a/app/models/claims/mentor_training.rb
+++ b/app/models/claims/mentor_training.rb
@@ -36,15 +36,50 @@ class Claims::MentorTraining < ApplicationRecord
     refresher: "refresher",
   }, default: :initial, validate: true
 
-  validates(
-    :hours_completed,
-    numericality: { greater_than: 0, less_than_or_equal_to: 20, only_integer: true },
-    allow_nil: true,
-  )
+  before_validation :set_training_type
+
+  validates :hours_completed,
+            allow_nil: true,
+            # TODO: Remove this 'unless' condition once claim revisions have been removed.
+            # The 'internal draft' state that powers revisions is confusing, but in this case we need to
+            # skip this validation rule so that Claims::Claim::CreateRevision can successfully duplicate
+            # the claim and its mentor training records without breaching the remaining training allowance.
+            # The validity of mentor training records is still checked before claims can be submitted, so this
+            # doesn't open up a loophole. It's just a quirk due to the way claim revisions are modelled.
+            # https://trello.com/c/4riw52nH/230-replace-claim-revisions-with-the-wizard-pattern
+            unless: -> { claim&.internal_draft? },
+            numericality: {
+              greater_than: 0,
+              less_than_or_equal_to: :max_hours,
+              only_integer: true,
+            }
 
   scope :without_hours, -> { where(hours_completed: nil).order_by_mentor_full_name }
   scope :order_by_mentor_full_name, -> { joins(:mentor).merge(Mentor.order_by_full_name) }
 
   delegate :full_name, to: :mentor, prefix: true, allow_nil: true
   delegate :name, to: :provider, prefix: true, allow_nil: true
+
+  private
+
+  def set_training_type
+    return if training_allowance.nil?
+
+    self.training_type = training_allowance.training_type
+  end
+
+  def max_hours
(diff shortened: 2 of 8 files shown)


app/models/claims/mentor_training.rb
```diff
@@ -36,15 +36,50 @@ class Claims::MentorTraining < ApplicationRecord
...
   delegate :full_name, to: :mentor, prefix: true, allow_nil: true
   delegate :name, to: :provider, prefix: true, allow_nil: true
+
+  private
+
+  def set_training_type
+    return if training_allowance.nil?
+
+    self.training_type = training_allowance.training_type
+  end
```
Nitemaeric, line 69: I'd suggest moving this into a Service Object. This is the only callback-based piece of data population. I've done this a lot in the past, and while it is seemingly harmless, it can get out of hand without caution.

By moving this logic into a Service Object, you can:

1. test this logic in isolation.
2. uncouple the assignment of this field within factories.

Nitemaeric: Right now, Claims are updated/created at https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/forms/claims/claim/provider_form.rb

app/models/claims/mentor_training.rb
```diff
@@ -36,15 +36,50 @@ class Claims::MentorTraining < ApplicationRecord
...
+
+  def max_hours
+    training_allowance.remaining_hours
+  end
```
JamieCleare2525, line 73: Given it's possible for `training_allowance` to be `nil`, should this be:
```suggestion
  def max_hours
    training_allowance&.remaining_hours
  end
```

JamieCleare2525 (commented):
Generally happy. Just a small question.

app/models/claims/mentor_training.rb
ollietreend: Good question @JamieCleare2525 🤔 I originally thought the same, and included logic to handle `nil` values, but `undercover` told me I never exercised that code path. 🤖 

I'm not entirely sure under what circumstances `training_allowance` will be `nil`, actually. I'll take a look!

app/models/claims/mentor_training.rb
Nitemaeric: That'll be this one: https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/forms/claims/claim/mentor_training_form.rb

Nitemaeric: That'll be this one: https://github.com/DFE-Digital/itt-mentor-services/blob/main/app/forms/claims/claim/mentors_form.rb

app/forms/claims/claim/mentors_form.rb
```diff
@@ -53,7 +53,7 @@ def mentor_trainings
         mentor_id:,
         provider_id: claim.provider_id,
         claim_id: claim.id,
-      )
+      ).tap(&:set_training_type)
```
Nitemaeric, line 56: This will attempt to re-set the training type each time it is called. I _think_ Rails has some _magic_ to avoid `update` calls to values that are unchanged, but I don't think it is that clever in the case of assigning values and then calling `#save`.

Maybe renaming to `#set_initial_training_type` and using `||=` addresses this.

Nevertheless, I don't see it as a blocker.

Nicely done everyone!

Nitemaeric (approved):

app/forms/claims/claim/mentors_form.rb
ollietreend: Active Record will avoid re-saving the model if an attribute has been assigned but its value hasn't changed. So we should be all good! 🙂

PR #1064, opened 2024-09-25
Update sandbox DSI configuration

Nitemaeric:

### DIFF
--- a/terraform/application/config/sandbox_app_env.yml
+++ b/terraform/application/config/sandbox_app_env.yml
@@ -18,8 +18,10 @@ TEACHING_RECORD_BASE_URL: https://preprod.teacher-qualifications-api.education.g
 TEACHING_RECORD_API_MINOR_VERSION: 20240416
 
 # User authentication
-SIGN_IN_METHOD: persona
+SIGN_IN_METHOD: dfe-sign-in
+DFE_SIGN_IN_ISSUER_URL: https://oidc.signin.education.gov.uk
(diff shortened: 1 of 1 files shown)


ollietreend (commented):
> Writing to external partners, such as ESFA, should be configured to not go to actual external partners.

Is this already the case? Or does that need adding to this PR?

I'm also wondering about integrations like GOV.UK Notify. Will that send real-world emails to users, or are they caught?

Nitemaeric:
> > Writing to external partners, such as ESFA, should be configured to not go to actual external partners.
> 
> Is this already the case? Or does that need adding to this PR?
> 
> I'm also wondering about integrations like GOV.UK Notify. Will that send real-world emails to users, or are they caught?

Sorry, didn't realise you responded to this.

It's the case for ESFA, where we don't want actual payments to go through. I'd assume things like, updating mentor TRNs and others, we wouldn't want to _really_ be updated either.

That being said, I think emails are still important to actually be sent. This is because, we want the ability to invite users onto the sandbox platform.

Nitemaeric:
@ollietreend Right now, the Sandbox environment isn't being utilised at all. I think it's best to bring it into some kind of usage or drop it entirely.

Sandbox is intended to be an end-user playground, similar to Staging for internal DfE. The key difference is that end-users expect data _similar_ to Production. They want to see their own school and associated providers.

That being said, I'll hand over the torch for this one to your team.

PR #1023, opened 2024-09-12
Add missing h1s

Kizr:
## Context

All pages in the application should hierachical headers, our accessibility testing revealed that we were missing h1's on a number of pages.

## Changes proposed in this pull request

- [x] Ensured that there are h1's on every page
- [x] Fixed some alignment and missing tags along the way

## Guidance to review

This will be difficult to test due to the number of pages that have been altered, there shouldn't be any real differences on pages but a quick jaunt through the pages is advisable.

## Link to Trello card

[Add H1 tags where they've been missed](https://trello.com/c/ta7njVcd/783-add-h1-tags-where-theyve-been-missed)

PR #1068, opened 2024-09-26
Refactor notification banners

dp-daly:
## Context

Refactor the way flash messages are rendered so it’s possible to provide both a heading and body content for the notification banner.

**1. Restructure the flash hash**

Change the structure of the hash so it more closely resembles the options accepted by the [notification banner component](https://govuk-components.netlify.app/components/notification-banner/).

This is passed through to the view where corresponding changes have been made to the way we read the flash variable now that it has different keys.

**2. Update translation keys to differentiate ‘heading’ and ‘body’**

The convention has been to use the translation key .success for the flash message content. These have now been renamed to ".success_heading".

Later, we’ll be able to use .success_body to define the body content for flash messages that need it. 

**3. Move rendering to a view partial**

**4. Tidy up: unnecessary flash rendering in the component preview layout**

The component_preview.html.erb layout does not need to render flash messages so code has been removed.

## Guidance to review

- Related frontend changes for review are detailed in PR[#1073](https://github.com/DFE-Digital/itt-mentor-services/pull/1073).

## Link to Trello card

https://trello.com/c/CunVgrTw/820-refactor-flash-messages-to-support-heading-and-body-content

## Screenshots

- See screenshot video on banner content changes (PR[#1073](https://github.com/DFE-Digital/itt-mentor-services/pull/1073)).


app/views/layouts/_notification_banner.html.erb
```diff
@@ -0,0 +1,11 @@
+<div class="govuk-!-margin-bottom-0">
+    <%= render(GovukComponent::NotificationBannerComponent.new(
+    title_text: flash[:success] != false ? t(".success") : t(".alert"),
+    success: flash[:success] != false,
+  )) do %>
+    <h3 class="govuk-heading-m"><%= flash[:heading] %></h3>
```
ollietreend, line 6: We should not be using a `<h3>` tag here.

Sorry 🙈 I thought we'd fixed this in a previous PR, because it was part of the accessibility work. And now I'm just confused.

ollietreend: Oh, I've just checked. It looks like I raised it here on a different PR https://github.com/DFE-Digital/itt-mentor-services/pull/1023#discussion_r1763616749 

So my bad. We never actually did fix it for the main `application.html.erb` layout.

config/environments/development.rb
```diff
@@ -77,7 +77,7 @@
   config.action_view.annotate_rendered_view_with_filenames = true
 
   # Raise error when a before_action's only/except options reference missing actions.
-  config.action_controller.raise_on_missing_callback_actions = true
+  config.action_controller.raise_on_missing_callback_actions = false
```
ollietreend, line 80: Should this be here? 🤔

spec/system/placements/sign_in_as_a_placements_user_spec.rb
```diff
@@ -74,7 +74,7 @@
   end
 
   context "when normal user tries to access support user page" do
-    scenario "I sign in as user mary trying to access support user page" do
+    scenario "I sign in as user mary trying to access support user page", :js do
```
ollietreend, line 77: Does this need to be here?

```suggestion
    scenario "I sign in as user mary trying to access support user page" do
```

app/views/layouts/_notification_banner.html.erb
ollietreend: So, we want to use `nb.with_heading` here – a bit like this:

https://github.com/DFE-Digital/itt-mentor-services/blob/d242b2408286606a265f3bb05189b2bd51801137/app/views/placements/schools/placements/preview.html.erb#L10-L12

I haven't tried it, but I think something like this should work:

```ruby
<%= render(GovukComponent::NotificationBannerComponent.new(
  title_text: flash[:success] != false ? t(".success") : t(".alert"),
  success: flash[:success] != false,
)) do |nb| %>
  <% nb.with_heading(flash[:heading]) %>
  <% if flash[:body].present? %>
    <p class="govuk-body"><%= flash[:body] %></p>
  <% end %>
<% end %>
```

app/controllers/claims/schools/claims_controller.rb
```diff
@@ -68,7 +68,10 @@ def remove; end
   def destroy
     @claim.destroy!
 
-    redirect_to claims_school_claims_path(@school, @claim), flash: { success: t(".success") }
+    redirect_to claims_school_claims_path(@school, @claim), flash: {
+      heading: t(".success"),
+      success: true,
```
ollietreend, line 73: As Ben said (on Slack) – we don't need the explicit `success: true` because it defaults to showing a success banner. So let's drop it from all the controllers.

ollietreend:
I'm closing this PR as it has been combined into the overall way finding PR #1071.

PR #1069, opened 2024-09-26
Only verify_policy_scoped on controllers that have index actions

Kizr:
## Context

Fix the issue causing the warning messages to avoid us having to disable the config for the warning,

- [x] verify_policy_scoped conditionally if there is an index action

## Guidance to review

- [x] Developers should be able to use any of the wizard journeys without error/run the test suite and have it pass

## Link to Trello card

[Re-enable raise_on_missing_callback_actions config](https://trello.com/c/YQgg4Tm4/822-re-enable-raiseonmissingcallbackactions-config)

### DIFF
--- a/app/controllers/placements/application_controller.rb
+++ b/app/controllers/placements/application_controller.rb
@@ -1,5 +1,5 @@
 class Placements::ApplicationController < ApplicationController
-  after_action :verify_policy_scoped, only: %i[index]
+  after_action :verify_policy_scoped, if: ->(c) { c.action_name == "index" }
   before_action :authorize_support_user!
 
   def current_user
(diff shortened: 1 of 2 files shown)


ollietreend (changes requested):
We still need to revert the associated config back to [Rails 7.2 defaults](https://github.com/rails/rails/blob/97169912f197eee6e76fafb091113bddf624aa67/guides/source/configuring.md?plain=1#L1860-L1867).

### ✅ Development environment

The **development** environment is fine now.

https://github.com/DFE-Digital/itt-mentor-services/blob/dd10b05bc49604a8e5a6f4508f5a6efedaff4b57/config/environments/development.rb#L79-L80

### ❌ Test environment

But the **test** environment is still configured to `false`. It needs reverting back to `true`. Also note that the comment needs updating, from "Don't raise error..." to "Raise error..."

https://github.com/DFE-Digital/itt-mentor-services/blob/dd10b05bc49604a8e5a6f4508f5a6efedaff4b57/config/environments/test.rb#L68-L69

PR #1080, opened 2024-10-03
Multi tab placement creation

Kizr:
## Context

Wizard journeys in the application only support one tab as they use the session to store the data with non-unique keys. This PR creates unique keys for each wizard so that a user can open as many tabs as they wish and still have the data correctly persisted in each tab. This is accomplished by passing a state_key via the URL params.

- [x] Update the base wizard to take state rather than the entire session
- [x] Add the state key to the URL params for wizard journeys
- [x] Update wizard controllers to send the correct state to their respective wizard
- [x] Add multi-tab tests to all wizards to make sure that they have all been updated correctly

## Guidance to review

Unfortunately this will take a while to test.

You will need to log in as every user and try every wizard in the application using multiple tabs, excluding single step wizard journeys (e.g. edit journeys) as these do not have the same issues as the longer journeys.

## Link to Trello card

[Spike out how we could support multi-tab wizard journeys](https://trello.com/c/05kqUr3N/831-spike-out-how-we-could-support-multi-tab-wizard-journeys)


spec/requests/placements/schools/mentors/add_mentor_spec.rb
```diff
@@ -2,14 +2,15 @@
 
 RSpec.describe "'Add mentor' journey", service: :placements, type: :request do
   let(:current_user) { create(:placements_user, schools: [school]) }
-  let(:start_path) { new_add_mentor_placements_school_mentors_path(school_id:) }
+  let!(:state_key) { SecureRandom.uuid }
+  let(:start_path) { new_add_mentor_placements_school_mentors_path(state_key:, school_id:) }
```
ollietreend, line 6: The `/new` path doesn't have a `state_key` route param. We shouldn't need to provide one – `/new` will redirect us to `/edit/:state_key/name-of-first-step`, where `:state_key` has been generated automatically by the controller.

```suggestion
  let(:start_path) { new_add_mentor_placements_school_mentors_path(school_id:) }
```

Since `state_key` is not a route param, I expect it probably becomes a query string param when provided to this helper. For example:

```
/schools/:school_id/placements/new?state_key=03e173a4-2266-405e-8427-0410921f8ced
```

In fact, I'm wondering if we should also be altering this test to make sure it generates a new state key and redirects the user to it. What do you think? 🤔

https://github.com/DFE-Digital/itt-mentor-services/blob/c7a56194a1cc3e249e967a2456aabb407b5afba9/spec/support/shared_examples/requests/placements/add_a_mentor.rb#L7-L22

spec/requests/placements/schools/mentors/add_mentor_spec.rb
```diff
@@ -2,14 +2,15 @@
 
 RSpec.describe "'Add mentor' journey", service: :placements, type: :request do
   let(:current_user) { create(:placements_user, schools: [school]) }
-  let(:start_path) { new_add_mentor_placements_school_mentors_path(school_id:) }
+  let!(:state_key) { SecureRandom.uuid }
```
ollietreend, line 5: Just something about the fact that we pre-determine the state key here.. it feels weird, because presumably that means we never exercise the code that generates a _new_ state key from the controller. Right?

(Except for hitting it in the system specs. Where we sort of don't pay much attention to it. Maybe that's okay?)

spec/requests/placements/schools/placements/add_placement_spec.rb
```diff
@@ -7,7 +7,8 @@
   let(:current_academic_year) { create(:placements_academic_year) }
   let(:drama) { create(:subject, :secondary, name: "Drama") }
   let(:summer_term) { create(:placements_term, :summer) }
-  let(:start_path) { new_add_placement_placements_school_placements_path(school_id:) }
+  let(:start_path) { new_add_placement_placements_school_placements_path(state_key:, school_id:) }
```
ollietreend, line 10: Same here

spec/requests/placements/schools/placements/edit_placement_spec.rb
```diff
@@ -8,7 +8,8 @@
   let(:provider) { create(:placements_provider) }
   let(:school_id) { school.id }
   let(:year_group) { :year_6 }
-  let(:start_path) { new_edit_placement_placements_school_placement_path(school, placement, step:) }
+  let(:start_path) { new_edit_placement_placements_school_placement_path(school, placement, state_key:, step:) }
```
ollietreend, line 11: Same here

spec/requests/placements/support/schools/placements/add_placement_spec.rb
```diff
@@ -7,7 +7,8 @@
   let(:current_academic_year) { create(:placements_academic_year) }
   let(:drama) { create(:subject, :secondary, name: "Drama") }
   let(:summer_term) { create(:placements_term, :summer) }
-  let(:start_path) { new_add_placement_placements_support_school_placements_path(school_id:) }
+  let!(:state_key) { SecureRandom.uuid }
+  let(:start_path) { new_add_placement_placements_support_school_placements_path(state_key:, school_id:) }
```
ollietreend, line 11: Same here

spec/requests/placements/support/schools/placements/edit_placement_spec.rb
```diff
@@ -8,7 +8,8 @@
   let(:provider) { create(:provider) }
   let(:school_id) { school.id }
   let(:year_group) { :year_6 }
-  let(:start_path) { new_edit_placement_placements_support_school_placement_path(school, placement, step:) }
+  let(:start_path) { new_edit_placement_placements_support_school_placement_path(school, placement, state_key:, step:) }
```
ollietreend, line 11: Same here

ollietreend (commented):
This all looks good to me, except for one thing which I think would be worth discussing in the request specs. Right now they're not _quite_ realistic – would be up for chatting this through.

spec/requests/placements/schools/mentors/add_mentor_spec.rb
Kizr: I added this originally because we can't match the path without the state key for the expectation, how do you feel about stubbing the value instead?

```ruby
  let(:state_key) { SecureRandom.uuid }

  before do
    sign_in_as current_user
    allow(SecureRandom).to receive(:uuid).and_return(state_key)
  end
```

spec/wizards/placements/base_wizard_spec.rb
```diff
@@ -257,33 +247,40 @@
...
+  describe ".generate_state_key" do
+    before do
+      allow(SecureRandom).to receive(:uuid).and_call_original
+      described_class.generate_state_key
+    end
+
+    it "returns a SecureRandom UUID" do
+      expect(SecureRandom).to have_received(:uuid).once
+    end
+  end
```
ollietreend, line 285: Right now the test doesn't actually prove that we return the UUID – we only assert that it was called.

This might be a nicer way to test it and cover both the `SecureRandom.uuid` call and the expected return value.

```suggestion
  describe ".generate_state_key" do
    it "returns a SecureRandom UUID" do
      expect(SecureRandom).to receive(:uuid).once.and_return("pineapple")
      expect(described_class.generate_state_key).to eq("pineapple")
    end
  end
```

ollietreend (commented):
This looks great. Thanks for being so patient with my reviews! 😅

I've got one final comment which I think would tighten up the unit test for `BaseWizard.generate_state_key` – but otherwise I think this is good to go.

PR #170, opened 2024-01-26
Update App Header & Banner

Nitemaeric:
## Context

- We want to be able to configure the hosting environment separately from the Rails environment.
  - This is to create a distinction between configuring Rails and configuring a hosted environment.
- We want to add colors to the app's header bottom border and phase banner tag background based on the current hosting environment.

As I was going through the changes, I also found some bugs and opportunities to improve what was closely related.

- Clicking on the Logo breaks the Claims app when logged in as a support user.
- The Phase Banner text didn't match the prototype.
- The feedback link in the Phase Banner didn't go anywhere.
- The existing overridden styles for the header did not support the mobile menu.

## Changes proposed in this pull request

- Add a `HOSTING_ENV` env var to control which environment we're targeting.
- Update header bottom border to be full width.
- Update `HostingEnvironment.env` to provide a matching experience to `Rails.env`.
  - [ActiveSupport::EnvironmentInquirer](https://www.rubydoc.info/github/rails/rails/ActiveSupport/EnvironmentInquirer)
- Pull out the phase banner localisation from `HostingEnvironment`.
- Update phase banner text to match prototype.
- Clicking on the header logo should take you to the `sign_in_path` or the user's `after_sign_in_path`.
- Add a _placeholder_ feedback page.
- Add [Climate Control](https://github.com/thoughtbot/climate_control) to safely set and reset ENV vars in tests.

## Guidance to review

- Try running the app with `HOSTING_ENV=production` and etc.
- By default, you should see the header and tags in grey.
- The header navigation menu is now properly full width on mobile.
- Clicking on the header's logo should redirect to path of least surprise now.

## Screenshots

| `HOSTING_ENV` | Screenshot |
| ----------------- | ------------ |
| production        | CleanShot 2024-01-30 at 16 15 34 |
| staging | CleanShot 2024-01-30 at 16 16 46 |
| sandbox | CleanShot 2024-01-30 at 16 17 13 |
| qa | CleanShot 2024-01-30 at 16 17 36 |
| review | CleanShot 2024-01-30 at 16 18 00 |
| development | CleanShot 2024-01-30 at 16 18 28 |

PR #1092, opened 2024-10-08
Change support console views to use multi org

JamieCleare2525:
## Context

- Change the support console, so that support users use the multi-org views.

- Move `set_school` and `set_provider` methods to the `Placements::ApplicationController`
- Allow Support Users to view school and provider views.
- Remove unnecessary routes, controllers and views which are no longer required since moving support users to the multi-org views.

## Guidance to review

- Sign in as Colin (Support User)
- Select a School from the Organisations list.
- Test to check that a Support User can perform all the features of the School User
- Change organisation
- Select a Provider from the Organisations list.
- Test to check that a Support User can perform all the features of the Provider User

## Link to Trello card

https://trello.com/c/R3I4rmO3/845-investigate-what-changes-will-need-to-be-made-for-the-support-console-changes


Kizr (approved):
This is big enough that I could have easily overlooked something, I think this'll need 2 reviews!

ollietreend (commented):
@JamieCleare2525 could we please drop the `app-phase-banner-—no-border-bottom` class on the phase banner, please?

Otherwise we end up with the phase banner _sometimes_ having an underline, and not having an underline in other places.

Two support console pages – one shows the phase banner with an underline, the other shows it without an underline

While we're at it, I'd support removing [that custom CSS rule](https://github.com/DFE-Digital/itt-mentor-services/blob/67cf8c69969ff6b2ea066579f4d742c73caf6fc7/app/assets/stylesheets/components/_phase_banner.scss#L1-L5) altogether. I think it was only slipped in to avoid an "unsightly" line in the support console (added in #170, but not mentioned until #199). I'm not really sure how valuable it is, and it seems to [unnecessarily complicate the views](https://github.com/DFE-Digital/itt-mentor-services/blob/e395ad86611b7f9cea6d57b16ee8379276032700/app/views/claims/schools/_primary_navigation.html.erb#L11-L13) IMHO.

app/controllers/placements/application_controller.rb
```diff
@@ -4,9 +4,24 @@ class Placements::ApplicationController < ApplicationController
...
+              end
+  end
+
+  def set_provider
+    @provider = if current_user.support_user?
+                  Placements::Provider.find(params.require(:provider_id))
+                else
+                  current_user.providers.find(params.require(:provider_id))
+                end
+  end
```
ollietreend, line 21: This seems like an ideal use case for policy scopes. Although it looks like we might not actually have suitable scopes defined for these use cases yet.

Maybe one to follow up with in a separate PR to avoid doing too much here?

app/controllers/placements/organisations_controller.rb
```diff
@@ -7,24 +7,27 @@ def index
...
-
-    redirect_to landing_page_path(organisation)
+  def organisation
+    @organisation ||= if !current_user.support_user?
+                        memberships.find(params.require(:id)).organisation
+                      elsif params.fetch(:type) == "School"
+                        School.find(params.require(:id))
+                      else
+                        Provider.find(params.require(:id))
+                      end
```
ollietreend, line 30: Ditto re: policy scopes.

app/controllers/placements/providers/partner_schools_controller.rb
```diff
@@ -14,7 +14,7 @@ def destroy
   private
 
   def set_organisation
-    @provider = current_user.providers.find(params.fetch(:provider_id))
+    @provider = set_provider
   end
```
ollietreend, line 18: It seems a shame this method is called `set_organisation` but it actually sets `@provider`.

Given the method name, I would have expected it to set `@organisation`, as happens in these other places:

https://github.com/DFE-Digital/itt-mentor-services/blob/89874b12bb0a01ed3526cee2da96fc57a7ac71c1/app/controllers/placements/providers/partner_schools/add_partner_school_controller.rb#L4-L6

https://github.com/DFE-Digital/itt-mentor-services/blob/89874b12bb0a01ed3526cee2da96fc57a7ac71c1/app/controllers/placements/providers/users/add_user_controller.rb#L4-L6

https://github.com/DFE-Digital/itt-mentor-services/blob/89874b12bb0a01ed3526cee2da96fc57a7ac71c1/app/controllers/placements/providers/users_controller.rb#L4-L6

https://github.com/DFE-Digital/itt-mentor-services/blob/89874b12bb0a01ed3526cee2da96fc57a7ac71c1/app/controllers/placements/schools/partner_providers/add_partner_provider_controller.rb#L4-L6

app/controllers/placements/schools/partner_providers_controller.rb
```diff
@@ -14,7 +14,7 @@ def destroy
   private
 
   def set_organisation
-    @school = current_user.schools.find(params.fetch(:school_id))
+    @school = set_school
   end
```
ollietreend, line 18: Same same.

app/controllers/placements/providers_controller.rb
```diff
@@ -1,5 +1,11 @@
...
   def show
-    @provider = current_user.providers.find(params.require(:id)).decorate
+    @provider = providers.find(params.require(:id)).decorate
+  end
+
+  private
+
+  def providers
+    current_user.support_user? ? Placements::Provider.all : current_user.providers
   end
```
ollietreend, line 10: Let's definitely lean in to policy scopes for this stuff! I have a feeling you've already planned to do this in a separate PR?

app/controllers/placements/schools_controller.rb
```diff
@@ -1,5 +1,11 @@
...
   def show
-    @school = current_user.schools.find(params.require(:id)).decorate
+    @school = schools.find(params.require(:id)).decorate
+  end
+
+  private
+
+  def schools
+    current_user.support_user? ? Placements::School.all : current_user.schools
   end
```
ollietreend, line 10: very scope

app/views/placements/providers/_primary_navigation.html.erb
```diff
@@ -1,9 +1,13 @@
...
       title: provider.name,
-      actions: [govuk_link_to(t(".change_organisation"), placements_organisations_path, no_visited_state: true)],
+      actions: [govuk_link_to(
+        t(".change_organisation"),
+        current_user.support_user? ? placements_support_organisations_path : placements_organisations_path,
+        no_visited_state: true,
+      )],
     )) %>
   <% end %>
 <% end %>
```
ollietreend, line 13: This could be a good candidate for refactoring as an 'organisation switcher' View Component:

- The [conditional rendering logic](https://github.com/DFE-Digital/itt-mentor-services/blob/89874b12bb0a01ed3526cee2da96fc57a7ac71c1/app/views/placements/providers/_primary_navigation.html.erb#L2) would work well as a `#render?` method
- The [link path logic](https://github.com/DFE-Digital/itt-mentor-services/blob/89874b12bb0a01ed3526cee2da96fc57a7ac71c1/app/views/placements/providers/_primary_navigation.html.erb#L8) would also fit nicely into a component class
- ...which would then be easier to unit test in a component spec

Anyway, probably another thing that's out of scope for this PR 🙂

app/views/placements/schools/_primary_navigation.html.erb
```diff
@@ -1,9 +1,13 @@
...
       title: school.name,
-      actions: [govuk_link_to(t(".change_organisation"), placements_organisations_path, no_visited_state: true)],
+      actions: [govuk_link_to(
+        t(".change_organisation"),
+        current_user.support_user? ? placements_support_organisations_path : placements_organisations_path,
+        no_visited_state: true,
+      )],
     )) %>
   <% end %>
 <% end %>
```
ollietreend, line 13: Okay – now that I've seen it in two places, it's definitely looking like something worth pulling into its own reusable component.

app/views/placements/wizards/add_user_wizard/_user_step.html.erb
```diff
@@ -1,12 +1,11 @@
-<% caption = current_user.support_user? ? "#{t(".caption")} #{contextual_text}" : t(".caption") %>
-<% content_for :page_title, user_step.errors.any? ? t(".title_with_error", caption:) : t(".title", caption:) %>
+<% content_for :page_title, user_step.errors.any? ? t(".title_with_error", caption: t(".caption")) : t(".title", caption: t(".caption")) %>
```
ollietreend, line 1: Okay 🫠 well this is definitely not looking fun. Three layers deep of translation nesting is a bit much for my mere mortal brain.

Again, probably not something for this PR. But I wonder if we need to think about if/how this could become easier to understand.

config/routes/placements.rb
ollietreend, line 1: Chef kiss emoji

ollietreend (commented):
Super work @JamieCleare2525! ⭐

This is going to make it _so much easier_ to maintain the support console in future.

+238 -1741 lines of code 😍 

It's also glorious to see this much code removed while simultaneously improving the UX and paying down product debt. AMAZING! 🙌🏻

app/controllers/placements/providers_controller.rb
ollietreend: Oh! Is it #1101?? 👀

app/controllers/placements/application_controller.rb
JamieCleare2525: I will address in https://github.com/DFE-Digital/itt-mentor-services/pull/1101

app/controllers/placements/organisations_controller.rb
JamieCleare2525: I will address in https://github.com/DFE-Digital/itt-mentor-services/pull/1101

app/controllers/placements/schools_controller.rb
JamieCleare2525: I will address this in https://github.com/DFE-Digital/itt-mentor-services/pull/1101

PR #1100, opened 2024-10-09
Rename sent_to_esfa claim status to payment_in_progress

Nitemaeric:
## Context

After some revision, we've decided to exclude explicitly showing ESFA to users, as some (or many) may not be familiar with the acronym.

- Rename `sent_to_esfa` enum value of claim statuses to `payment_in_progress`.

## Guidance to review

- The `ESFA` acronym is still required in inflections, as it is also used in the migration file name/class mapping.


JamieCleare2525 (approved):
LGTM

PR #1101, opened 2024-10-10
Improve coverage of policy scopes in controllers

JamieCleare2525:
## Context

- Improve use of `policy_scope` throughout the controllers.

- Ensure all controller methods use `policy_scope` where possible.

## Guidance to review

- Sign in as each user type and test that all features still work as expected.

## Link to Trello card

https://trello.com/c/Mi2stP5t/848-expand-and-improve-on-the-use-of-pundit-scopes

### DIFF
--- a/app/controllers/placements/organisations/users_controller.rb
+++ b/app/controllers/placements/organisations/users_controller.rb
@@ -5,8 +5,7 @@ class Placements::Organisations::UsersController < Placements::ApplicationContro
   before_action :authorize_user, only: %i[remove destroy]
 
   def index
-    scope = policy_scope(users)
-    @users = scope.order_by_full_name
+    @users = users.order_by_full_name
   end
 
   def show; end
@@ -26,7 +25,7 @@ class Placements::Organisations::UsersController < Placements::ApplicationContro
   end
 
   def users
-    @users = @organisation.users
+    policy_scope(@organisation.users)
   end
 
   def set_user_membership
--- a/app/controllers/placements/organisations_controller.rb
+++ b/app/controllers/placements/organisations_controller.rb
@@ -24,9 +24,9 @@ class Placements::OrganisationsController < Placements::ApplicationController
     @organisation ||= if !current_user.support_user?
                         memberships.find(params.require(:id)).organisation
                       elsif params.fetch(:type) == "School"
-                        School.find(params.require(:id))
+                        policy_scope(Placements::School).find(params.require(:id))
                       else
-                        Provider.find(params.require(:id))
+                        policy_scope(Placements::Provider).find(params.require(:id))
                       end
   end
 
--- a/app/controllers/placements/organisations_controller.rb
+++ b/app/controllers/placements/organisations_controller.rb
@@ -21,9 +21,7 @@ class Placements::OrganisationsController < Placements::ApplicationController
   end
 
   def organisation
-    @organisation ||= if !current_user.support_user?
-                        memberships.find(params.require(:id)).organisation
-                      elsif params.fetch(:type) == "School"
+    @organisation ||= if params.fetch(:type) == "School"
                         policy_scope(Placements::School).find(params.require(:id))
                       else
                         policy_scope(Placements::Provider).find(params.require(:id))
--- a/app/controllers/placements/providers/partner_schools/placements_controller.rb
+++ b/app/controllers/placements/providers/partner_schools/placements_controller.rb
@@ -1,16 +1,15 @@
 class Placements::Providers::PartnerSchools::PlacementsController < Placements::ApplicationController
-  skip_after_action :verify_policy_scoped, only: %i[index]
   before_action :set_provider, only: %i[index show]
   before_action :set_partner_school, only: %i[index show]
 
   def index
-    @placements_assigned_to_provider = placements_scope.where(provider: @provider).decorate
-    @other_placements = placements_scope.where.not(provider: @provider)
-      .or(placements_scope.where(provider: nil)).decorate
+    @placements_assigned_to_provider = placements.where(provider: @provider).decorate
+    @other_placements = placements.where.not(provider: @provider)
+      .or(placements.where(provider: nil)).decorate
   end
 
   def show
-    @placement = placements_scope.find(params.require(:id)).decorate
+    @placement = placements.find(params.require(:id)).decorate
     @partner_school = @placement.school
   end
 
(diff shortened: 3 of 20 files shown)


Kizr (approved):

app/controllers/placements/organisations/users_controller.rb
```diff
@@ -26,7 +25,7 @@ def set_user
   end
 
   def users
-    @users = @organisation.users
+    @users = policy_scope(@organisation.users)
```
ollietreend, line 28: I don't think this needs to be memoized:

- `index` already sets `@users` to what it should be
- `show` calls this method and then sets `@user` accordingly

This method is not a `set_users` method, just a helper method to get the scoped `users`.

```suggestion
    policy_scope(@organisation.users)
```

app/controllers/placements/schools/mentors_controller.rb
```diff
@@ -31,7 +31,7 @@ def destroy
   private
 
   def set_mentor
-    @mentor = @school.mentors.find(params.require(:id))
+    @mentor = policy_scope(@school.mentors).find(params.require(:id))
   end
```
ollietreend, line 35: How about following the same pattern here?

```ruby
class MentorsController
  def index
    @page, @mentors = page(mentors.order_by_full_name)
  end
  
  private
  
  def mentors
    policy_scope(@school.mentors)
  end
  
  def set_mentor
    @mentor = mentors.find(params.require(:id))
  end
end
```

I quite like this approach because it defines the scope for this entire controller in a single place. That should make it easier to 'get it right' if we add more actions to this controller in future – we're nudging future developers towards the [pit of success](https://blog.codinghorror.com/falling-into-the-pit-of-success/).

app/controllers/placements/schools/partner_providers_controller.rb
```diff
@@ -22,7 +22,7 @@ def source_organisation
   end
 
   def set_decorated_partner_organisation
-    @partner_provider = @school.partner_providers.find(params.require(:id)).decorate
+    @partner_provider = policy_scope(@school.partner_providers).find(params.require(:id)).decorate
   end
```
ollietreend, line 26: Likewise here. It might be nice to have a scope method:

```ruby
def partner_providers
  policy_scope(@school.partner_providers)
end
```

app/controllers/placements/schools/placements_controller.rb
```diff
@@ -46,7 +46,7 @@ def academic_year_scope
   end
 
   def set_placement
-    @placement = @school.placements.find(params.require(:id))
+    @placement = policy_scope(@school.placements).find(params.require(:id))
   end
```
ollietreend, line 50: Ditto

app/controllers/placements/providers/partner_schools/placements_controller.rb
```diff
@@ -21,6 +20,6 @@ def set_partner_school
   end
 
   def placements_scope
-    @partner_school.placements
+    policy_scope(@partner_school.placements, policy_scope_class: Placements::Provider::PlacementPolicy::Scope)
   end
```
ollietreend, line 24: Hmm. Maybe a convention would be helpful here.

If we were following the other ones commented above ☝🏻 I guess it would mean renaming the method to simply `#placements`.

Or do you think the `_scope` suffix is useful?

app/controllers/placements/providers/partner_schools_controller.rb
```diff
@@ -22,7 +22,7 @@ def source_organisation
   end
 
   def set_decorated_partner_organisation
-    @partner_school = @provider.partner_schools.find(params.require(:id)).decorate
+    @partner_school = policy_scope(@provider.partner_schools).find(params.require(:id)).decorate
```
ollietreend, line 25:
```suggestion
    @partner_school = partner_schools.find(params.require(:id)).decorate
```

app/controllers/placements/providers/placements_controller.rb
```diff
@@ -19,7 +19,10 @@ def index
   end
 
   def show
-    @placement = Placement.find(params[:id]).decorate
+    @placement = policy_scope(
+      Placement.all,
+      policy_scope_class: Placements::Provider::PlacementPolicy::Scope,
+    ).find(params.require(:id)).decorate
```
ollietreend, line 25: This one might be trickier because `#index` uses a query object – but I wonder if [Active Record's merge method](https://api.rubyonrails.org/classes/ActiveRecord/SpawnMethods.html#method-i-merge) would work? 🤔 

```ruby
class PlacementsController
  def index
    query = Placements::PlacementsQuery.call(params: query_params)
    @pagy, @placements = pagy(placements.merge(query))
  end

  def show
    @placement = placements.find(params.require(:id))
    @school = @placement.school
  end

  private

  def placements
    policy_scope(
      Placement.all,
      policy_scope_class: Placements::Provider::PlacementPolicy::Scope,
    )
  end
end
```

ollietreend (commented):
This is looking good. It'll be great to have these policies in use throughout our controller actions.

As I'm looking at it, I'm wondering whether there's an opportunity to establish some conventions around how those scopes are defined. I think it could be nice for there to be a predictable place to get the scoped collection for the controller.

What do you think?

app/controllers/placements/schools/placements_controller.rb
```diff
@@ -6,8 +6,13 @@ class Placements::Schools::PlacementsController < Placements::ApplicationControl
...
   def index
-    @pagy, placements = pagy(placements_scope.includes(:subject, :mentors, :additional_subjects, :provider).order("subjects.name"))
-    @placements = placements.decorate
+    @pagy, scoped_placements = pagy(
+      placements
+        .where(academic_year: academic_year_scope)
+        .includes(:subject, :mentors, :additional_subjects, :provider)
+        .order("subjects.name"),
+    )
+    @placements = scoped_placements.decorate
```
ollietreend, line 15: It's a shame there's no `@placements.decorate!` method which would avoid needing to re-assign the variable.

But I wonder if this might be cleaner to avoid the awkwardness of needing a name for the intermediate variable? I think it should work.

```suggestion
    @pagy, @placements = pagy(
      placements
        .where(academic_year: academic_year_scope)
        .includes(:subject, :mentors, :additional_subjects, :provider)
        .order("subjects.name"),
    )
    @placements = @placements.decorate
```

ollietreend (commented):
Just a couple more comments 🙇🏻 but this is looking excellent, and I'm looking forward to it going in! Thanks.

app/controllers/placements/organisations_controller.rb
```diff
@@ -24,9 +24,9 @@ def organisation
     @organisation ||= if !current_user.support_user?
                         memberships.find(params.require(:id)).organisation
                       elsif params.fetch(:type) == "School"
-                        School.find(params.require(:id))
+                        policy_scope(Placements::School).find(params.require(:id))
                       else
-                        Provider.find(params.require(:id))
+                        policy_scope(Placements::Provider).find(params.require(:id))
                       end
```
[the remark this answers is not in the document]
JamieCleare2525: I had to make an adjustment to the `app/views/placements/organisations/index.html.erb` to make this work.

ollietreend (approved):
Chef kiss emoji

PR #1105, opened 2024-10-11
Refactor mailers

Kizr:
## Context

We need some more flexibility in our mailers so different services can send different emails to different types of users.

- [x] Create `Claims::ApplicationMailer` and `Placements::ApplicationMailer` which split out the service dependent logic.
- [x] Create `Claims::UserMailer` and `Placements::UserMailer` which inherit from thier service specific application mailer
- [x] Remove `UserMailer`
- [x] Create `Claims::SupportUserMailer` and `Placements::SupportUserMailer` which inherit from thier service specific application mailer
- [x] Remove `SupportUserMailer`
- [x] Create `user_mailer_class` and `support_user_mailer_class` methods to find the correct mailer to use for the current service
- [x] Refactor all mailer usages to point at their service specific mailer classes.
- [x] Refactor all mailer specs to mirror this logic

## Guidance to review

On the claims and placements services:
- Log in as Colin
- Navigate to Settings
- Ensure all mailers are displaying correctly

## Link to Trello card

[Refactor email notifications to be service specific](https://trello.com/c/xLi3O7G7/864-refactor-email-notifications-to-be-service-specific)


app/services/application_service.rb
```diff
@@ -8,4 +8,24 @@ def call
...
+  end
+
+  def support_user_mailer_class(service)
+    case service
+    when :placements
+      Placements::SupportUserMailer
+    when :claims
+      Claims::SupportUserMailer
+    end
+  end
```
Kizr, line 30: Where do we think these should live? They're only relevant to services and each method is used in two services. Maybe these could live in the application mailer instead?

app/mailers/claims/user_mailer.rb
```diff
@@ -0,0 +1,61 @@
...
+                   organisation_name: claim.school.name,
+                   reference: claim.reference,
+                   amount: claim.amount.format(symbol: true, decimal_mark: ".", no_cents: false),
+                   support_email:,
+                   link_to_claim:,
+                 )
+  end
+end
+
+class InvalidOrganisationError < StandardError; end
```
ollietreend, line 61: What's this class needed for? 🤔

I'm reviewing this on my phone, so it might be easier to see on a laptop. It feels like a bit of a "WTF" though.

app/services/placements/partnerships/notify/create.rb
```diff
@@ -6,7 +6,7 @@ class Create < Notify
 
         def notify_users
           partner_organisation_users.each do |user|
-            UserMailer.with(service: :placements)
+            ::Placements::UserMailer
```
ollietreend, line 9: Any reason why we have `Claims::UserMailer` but `::Placements::UserMailer` is prefixed with `::`?

I wonder why it works fine in one place but not the other.

app/mailers/claims/support_user_mailer.rb
```diff
@@ -0,0 +1,32 @@
...
+      slack_url:,
+    )
+
+    notify_email to: support_user.email, subject:, body:
+  end
+
+  private
+
+  def slack_url
+    "https://ukgovernmentdfe.slack.com/archives/C0657JE64HX"
```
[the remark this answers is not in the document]
ollietreend: Or, at the very least, we could put it in the same place as the support email address – i.e. the translations?

ollietreend (approved):

JamieCleare2525 (approved):

PR #1096, opened 2024-10-08
Placements index: Move location search to filter sidebar

dp-daly:
## Context

These design changes are in preparation of a new travel time feature on the placements index page. 

## Changes proposed in this pull request

- Page title and H1 change: “Find placements”.
- Location search bar moved to filter sidebar.
- Location search included as a selected filter tag. 
- Location search cleared by clearing filters.
- New paragraph tag shows "results sorted by" either alphabetical order or distance from the search location.
- Number of results move to results column.
- The structure of the titles of the search results has changed to [Subject] en dash [School name]. This means the links remain understandable for accessibility reasons, but puts the subject first.
- The size of the link becomes govuk-heading-m to make it stand out from the content and give visual hierarchy.

## Guidance to review

Please review the full travel time feature in this PR: https://github.com/DFE-Digital/itt-mentor-services/pull/1115.

## Link to Trello card

https://trello.com/c/oYsYxDX2/854-layout-changes-to-placements-search-page

PR #1115, opened 2024-10-15
Display travel time on placements in search results

dp-daly:
## Context

User is able to see travel times from the destination they have entered.

**Placements index**
- A travel time section is added to the listing content.
- This displays travel time for Public transport, Driving, Walking.
- Format should be:
> - XX minutes (up to an hour)
> - XX hour XX minutes (1-2 hours)
> - XX hours XX minutes (2+ hours)

## Guidance to review

Please use this PR to review content changes for the travel time feature, including from https://github.com/DFE-Digital/itt-mentor-services/pull/1096.

- Login as a Provider User.
- You will land on the placements index page.
- Search a location for which there will be a result (e.g. a town or city close to seeded placements).
- Check that the location appears as a selected filter tag.
- Check that results are rendered as expected (is the results number correct?; does the paragraph say "results sorted by distance from [the location you searched]"?)
- Check that the location clears when the filter tag is cancelled or the 'clear filters' button is pressed.
- Check that all three travel modes appear with different travel times.
- Search a location for which there will be no result.
- Check that a 'no results' message appears and the 'sorted by' paragraph is not visible.
- Use a combination of filters to make sure both the filter and search interact as you would expect.

## Link to Trello card

https://trello.com/c/J1W9oboz/856-display-travel-time-in-placement-search-results

## Screenshots

https://github.com/user-attachments/assets/e8fb29f9-d19a-467f-bfcb-789a3b25d0dd

### DIFF
--- a/app/decorators/school_decorator.rb
+++ b/app/decorators/school_decorator.rb
@@ -3,6 +3,13 @@ class SchoolDecorator < OrganisationDecorator
   attribute :walk_travel_duration
   attribute :drive_travel_duration
 
+  def formatted_duration(mode)
+    duration = send("#{mode}_travel_duration")
+    return "" if duration.blank?
+
+    duration.gsub(/\bmins\b/, I18n.t("components.placement.summary_component.minutes"))
+  end
+
   def formatted_inspection_date
     return "" if last_inspection_date.blank?
 
--- a/spec/system/placements/placements/searching_for_a_placement_spec.rb
+++ b/spec/system/placements/placements/searching_for_a_placement_spec.rb
@@ -37,6 +37,7 @@ RSpec.describe "Placements / Placements / Searching for a placements list",
     bath_placement
 
     given_i_sign_in_as_patricia
+    stub_travel_time_service
   end
 
   context "when searching for placements near a location (London)" do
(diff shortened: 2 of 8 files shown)


spec/system/placements/placements/searching_for_a_placement_spec.rb
```diff
@@ -60,7 +61,7 @@
       and_i_fill_in_location_search_with("Guildford")
       and_i_click_on("Apply filters")
       then_i_see_placements_for(school_name: "Guildford School", list_item: 0, distance: 0.0)
-      and_i_see_placements_for(school_name: "London School", list_item: 1, distance: 26.7)
+      then_i_see_placements_for(school_name: "London School", list_item: 1, distance: 26.7)
```
JamieCleare2525, line 64: I think you're missing tests to check the drive, walking and transit times appear.

JamieCleare2525 (commented):
I believe the test coverage is missing some steps.

app/decorators/school_decorator.rb
```diff
@@ -3,6 +3,13 @@ class SchoolDecorator < OrganisationDecorator
   attribute :walk_travel_duration
   attribute :drive_travel_duration
 
+  def formatted_duration(mode)
+    duration = send("#{mode}_travel_duration")
+    return "" if duration.blank?
+
+    duration.gsub(/\bmins\b/, "minutes")
```
JamieCleare2525, line 10: @Kizr alluded to a way a way of doing this with translations.
I would at least suggest moving "minutes" to a locale.

dp-daly: "minutes" has been added to a locale.

JamieCleare2525 (approved):
I'm happy with this. If @Kizr comes up with an idea to use translations for "minutes", we can address it in another PR

PR #1116, opened 2024-10-15
Add Azure Storage

Nitemaeric:
## Context

In preparation for the upcoming Payments/Sampling/Clawback flows, we want to be able to store CSV files attached to ActiveRecord records.

Namely, when we generate CSVs, we want to save a snapshot of the file sent to ESFA and Providers.

This PR implements the foundational ActiveStorage and Azure Storage functionality.

- Add Azure Storage to terraform configuration.
- Setup Rails' Azure Storage ActiveStorage service for testing.
- Add all ActiveStorage tables to the `analytics_blocklist.yml`.

## Guidance to review

I've run the following code to validate file read/write capabilities:

```ruby
class Claims::SupportUser < User
  has_one_attached :avatar
end

support_user = Claims::SupportUser.last
support_user.avatar.attach(io: URI.open("https://avatars.githubusercontent.com/u/4231530?s=40&v=4"), filename: "avatar.png") # My GitHub avatar
support_user.avatar.url # => "https://s189t01ittmsrvsa.blob.core.windows.net/storage/r1whzefgb4yj39dwedtwfhfsjt85?sp=r&sv=2024-05-04&se=2024-10-16T12%3A33%3A04Z&sr=b&rscd=inline%3B+filename%3D%22avatar.png%22%3B+filename*%3DUTF-8%27%27avatar.png&rsct=image%2Fjpeg&sig=qV4CAnHfNASQc8PVg64H8zVYUN53n%2FItnsRe0M58K%2Bk%3D"
```

Try visiting the link here:

~~https://s189t01ittmsrvsa.blob.core.windows.net/storage/r1whzefgb4yj39dwedtwfhfsjt85?sp=r&sv=2024-05-04&se=2024-10-16T12%3A33%3A04Z&sr=b&rscd=inline%3B+filename%3D%22avatar.png%22%3B+filename*%3DUTF-8%27%27avatar.png&rsct=image%2Fjpeg&sig=qV4CAnHfNASQc8PVg64H8zVYUN53n%2FItnsRe0M58K%2Bk%3D~~

Link has expired. Good.

### DIFF
--- a/terraform/application/storage_account.tf
+++ b/terraform/application/storage_account.tf
@@ -1,0 +1,31 @@
+resource "azurerm_storage_account" "general_storage_account" {
+  name                             = local.azure_storage_account_name
+  resource_group_name              = local.app_resource_group_name
+  location                         = "UK South"
+  account_tier                     = "Standard"
+  account_replication_type         = "LRS"
+  allow_nested_items_to_be_public  = false
+  cross_tenant_replication_enabled = false
+
+  dynamic "blob_properties" {
+    for_each = var.environment == "production" ? [1] : []
+
+    content {
+      delete_retention_policy {
+        days = 7
+      }
+
+      container_delete_retention_policy {
+        days = 7
+      }
+    }
+  }
+
+  lifecycle { ignore_changes = [tags] }
+}
+
+resource "azurerm_storage_container" "general_container" {
+  name                  = "storage"
+  storage_account_name  = azurerm_storage_account.general_storage_account.name
+  container_access_type = "private"
+}
--- a/Gemfile
+++ b/Gemfile
@@ -116,6 +116,9 @@ gem "okcomputer"
 # OpenStruct [https://github.com/ruby/ostruct]
 gem "ostruct", "~> 0.6.0"
 
+# Azure Storage adapter for Active Storage
+gem "azure-blob"
+
 group :development do
   gem "annotate", require: false
   gem "prettier_print", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -121,6 +121,8 @@ GEM
     audited (5.7.0)
       activerecord (>= 5.2, < 8.0)
       activesupport (>= 5.2, < 8.0)
+    azure-blob (0.5.2)
+      rexml
     base64 (0.2.0)
     better_html (2.1.1)
       actionview (>= 6.0)
@@ -684,6 +686,7 @@ DEPENDENCIES
   amazing_print
   annotate
   audited
+  azure-blob
   better_html
   bootsnap
   brakeman
--- a/config/analytics_blocklist.yml
+++ b/config/analytics_blocklist.yml
@@ -22,6 +22,27 @@ shared:
     - remote_address
     - request_uuid
     - created_at
+  :active_storage_variant_records:
+    - id
+    - blob_id
+    - variation_digest
+  :active_storage_blobs:
+    - id
+    - key
+    - filename
+    - content_type
+    - metadata
+    - service_name
+    - byte_size
+    - checksum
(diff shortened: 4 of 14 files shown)


terraform/application/config/review.tfvars.json
```diff
@@ -2,5 +2,7 @@
   "cluster": "test",
   "namespace": "bat-qa",
   "deploy_azure_backing_services": false,
-  "enable_postgres_ssl": false
+  "enable_postgres_ssl": false,
+  "create_storage_account": true,
+  "storage_account_name": "s189t01ittmsrvsa"
```
RMcVelia, line 7: using the same name for review apps must it will already exist, so can't be created on later rv apps. You'll need to use a local var with the pr number in the name e.g. https://github.com/DFE-Digital/access-your-teaching-qualifications/blob/main/terraform/application/storage.tf#L6

terraform/application/storage_account.tf
```diff
@@ -0,0 +1,37 @@
+resource "azurerm_storage_account" "general_storage_account" {
+  count                            = var.create_storage_account ? 1 : 0
```
RMcVelia, line 2: you could replace this with 
count  = var.storage_account_name != null ? 1 : 0
and then drop the create_storage_account variable.
But as it looks like you also create the storage account for every env, you could just drop the count and var completely?

terraform/application/storage_account.tf
```diff
@@ -0,0 +1,37 @@
...
+
+      container_delete_retention_policy {
+        days = 7
+      }
+    }
+  }
+
+  tags = {
+    environment = var.environment
+  }
```
RMcVelia, line 27: Shouldn't need this tag as it should be added automatically from the resource group

terraform/application/config/dv_review.tfvars.json
```diff
@@ -1,5 +1,7 @@
 {
   "namespace": "development",
   "deploy_azure_backing_services": false,
-  "enable_postgres_ssl": false
+  "enable_postgres_ssl": false,
+  "create_storage_account": true,
+  "storage_account_name": "s189d01ittmsrvsa"
```
RMcVelia, line 6: we usually add something descriptive to the sa name, e.g. s189d01ittmscsvrvsa

terraform/application/storage_account.tf
Nitemaeric: I used https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/terraform/aks/storage_account.tf as my reference. I don't actually know what this does 😅 

Can I drop lines 25:27?

terraform/application/config/dv_review.tfvars.json
Nitemaeric: This won't be specifically for CSVs, it will be the apps main/generic storage account. I've opted for a similarly generic name.

That being said, I'm unbiased about the infrastructure naming conventions.

What do you think?

terraform/application/storage_account.tf
RMcVelia: yes you can drop 25:27

terraform/application/config/dv_review.tfvars.json
RMcVelia: maybe using 'app' to identify it? so ittmsapprvsa

terraform/application/config/review.tfvars.json
Nitemaeric: I've removed the json configuration for the storage account and added in some code to compute what the account name should be for each environment.

RMcVelia (approved):

JamieCleare2525 (approved):
LGTM!

PR #1117, opened 2024-10-16
Update rubocop.yml to allow shared system spec to use scenario

JamieCleare2525:
## Context

- After some exploration, `shared_examples` aren't ideally suited to handle `system`/`feature` specs.
- As a result, the rubocop rules for `RSpec/Dialect` have been ignored for files in the `spec/support/shared_examples/system` directory.

- Rubocop rule for `RSpec/Dialect` is now ignored for files in the `spec/support/shared_examples/system` directory.
- Shared system tests now use `scenario` instead of `it`

## Link to Trello card

https://trello.com/c/15g93JQU/870-configure-rubocop-so-shared-examples-can-use-system-spec-syntax

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -28,6 +28,8 @@ Lint/MissingSuper:
 
 RSpec/Dialect:
   Enabled: true
+  Exclude:
+    - spec/support/shared_examples/system/**/*.rb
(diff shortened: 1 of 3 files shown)


ollietreend (approved):
Nice work

PR #1121, opened 2024-10-21
Add view component previews to settings page

Nitemaeric:
## Context

To rapidly iterate on development of view components, we want to be able to see multiple variants.

- Add links to view components to the Settings pages for ease of access.
- Fix a view component preview bug.

## Guidance to review

- You can visit the Settings page and see "View components".
- Visiting the "View components" link should show a list of service-specific component previews and shared component previews.
  - \* Right now, some components are not under the `Claims` or `Placements` namespaces, hence some undesired results. This can be fixed by namespacing the components correctly.
- Preview pages should only be visible to Support Users.
- Components that use `FactoryBot` break right now.
  - Many solutions here, but offloading those to be considered in the future.

### DIFF
--- a/app/controllers/claims/support/view_components_controller.rb
+++ b/app/controllers/claims/support/view_components_controller.rb
@@ -1,0 +1,7 @@
+class Claims::Support::ViewComponentsController < Claims::Support::ApplicationController
+  before_action :skip_authorization
+
+  def index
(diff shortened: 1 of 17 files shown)


ollietreend (commented):
@Nitemaeric I don't think we need to implement this. The View Component gem already comes with a built in preview feature which we can access in development at:

http://placements.localhost:3000/rails/view_components

View Component previews

Nitemaeric:
@ollietreend Happy for this to be closed.

The next step I had planned for this was groups of related components, but capacity and time is fleeting 😅

PR #1129, opened 2024-10-23
Add claims_private_beta flag to schools

Nitemaeric:
## Context

We need a flag on Schools to determine the starting academic year of claims.

Private beta schools will see 2023-2024 as the starting academic year of claims, others will see 2024-2025.

We need to add in the flag now, rather than later, to ensure that the current Schools in production have been marked as private beta schools, before we add any additional schools to the list.

- Add the `claims_private_beta` boolean attribute to the `schools` table.

## Guidance to review

- This is intended as additional information, for now. Changes to claims lists and their starting years will be added later down the line.

### DIFF
--- a/app/models/claims/school.rb
+++ b/app/models/claims/school.rb
@@ -8,6 +8,7 @@
 #  address3                               :string
 #  admissions_policy                      :string
 #  claims_grant_conditions_accepted_at    :datetime
+#  claims_private_beta                    :boolean          default(FALSE)
 #  claims_service                         :boolean          default(FALSE)
 #  district_admin_code                    :string
 #  district_admin_name                    :string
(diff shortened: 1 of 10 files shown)


Nitemaeric:
@ollietreend As you called out in Slack, it's possible for you to implement this properly.

The idea for this PR was to mark schools as part of the private beta for 2 reasons:

1. We want to show the first academic year of "2023 to 2024" for Private Beta schools and "2024 to 2025" for Public Beta schools.
2. We want to highlight in the support console that a school was part of the Private Beta.

CleanShot 2024-11-04 at 17 09 14@2x

Ultimately, both of those goals can be achieved with the `School` to `AcademicYear` join table we've discussed before.

PR #1136, opened 2024-10-24
Organisation switcher component

JamieCleare2525:
## Context

- Create an Organisation Switcher component to replace the duplicated code.

- Create `OrganisationSwitcherComponent`
- Replace duplicated code with component.

## Guidance to review

1. Fetch this branch locally
- visit http://placements.localhost:3000/rails/view_components/placement/organisation_switcher_component/default to view the preview of the component.

--- The other review scenarios can be performed on the review app ---

2. Sign in as Colin (Support User)
- Select any organisation
- The organisation switcher component should show at the top of the page.

3. Sign in as Mary (Multi-org User)
- Select any organisation
- The organisation switcher component should show at the top of the page.

4. Sign in as Anne (School User)
- The organisation switcher component should NOT show at the top of the page.

5. Sign in as Patricia (Provider User)
- The organisation switcher component should NOT show at the top of the page.

## Link to Trello card

https://trello.com/c/KQVvkeY1/887-create-a-placements-organisation-switcher-view-component

## Screenshots

Screenshot 2024-10-24 at 14 48 03


app/components/placement/organisation_switcher_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= content_for(:header_content) do %>
+  <%= render(ContentHeaderComponent.new(
+    title: organisation.name,
+    actions: [govuk_link_to(
+      t(".change_organisation"),
+      change_organisation_link,
+      no_visited_state: true,
+    )],
+  )) %>
```
ollietreend, line 9: Since this component just renders another component, it might be simpler to skip the `.html.erb` file altogether and just [define a `#call` method](https://viewcomponent.org/guide/templates.html#call).

For example, that's what's been done here:

https://github.com/DFE-Digital/itt-mentor-services/blob/99466916b2091d346122b94afaece52e90d63782/app/components/claim/status_tag_component.rb#L10-L12

spec/components/previews/placement/status_tag_component_preview.rb
```diff
@@ -1,4 +1,4 @@
-class Placement::SummaryComponentPreview < ApplicationComponentPreview
+class Placement::StatusTagComponentPreview < ApplicationComponentPreview
```
ollietreend, line 1: Thanks for fixing this

spec/components/placement/organisation_switcher_component_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    it "renders the organisation switcher" do
+      render_inline(component)
+
+      within(".content-header") do
+        expect(page).to have_content(organisation.name)
+        expect(page).to have_link(
+          text: "Change organisation",
+          href: public_send("placements_support_organisations_path"),
+        )
+      end
```
ollietreend, line 22: It definitely doesn't feel good to be using `within(".content-header")` here. I think the component should stand alone from the layout section it gets injected into (as commented above).

ollietreend (commented):
This is looking good, but I have a few thoughts.

ollietreend (approved):

PR #1138, opened 2024-10-25
Optimise travel time performance 

dp-daly:
## Context

Previously we called Placements::TravelTime in `summary_component.rb`, meaning we performed 3x API calls for every single listing shown on the page. At 25 listings per page, that’s 75 external calls to the Google Routes API.

This PR moves the service call to the placements controller so that `calculate_travel_time` is called once for the entire collection of results.

- Move `calculate_travel_time` (which calls the Placements::TravelTime service object) to the placements controller
- Remove logic from `summary_component.rb`
- Add additional response hashes to `routes_api` travel time stubs to reflect that a whole collection of results is now decorated and populated, as opposed to a single result

## Guidance to review

- Review placements controller and `calculate_travel_time` method
- Run test `spec/system/placements/providers/placements/searching_for_a_placement_spec.rb`
- Run test `spec/system/placements/providers/placements/view_placements_list_spec.rb`

## Link to Trello card

https://trello.com/c/0NEkMKfM/888-optimise-performance-of-travel-time-calculations

### DIFF
--- a/app/controllers/placements/providers/placements_controller.rb
+++ b/app/controllers/placements/providers/placements_controller.rb
@@ -12,6 +12,12 @@ class Placements::Providers::PlacementsController < Placements::ApplicationContr
     @terms = Placements::Term.order_by_term.select(:id, :name)
     query = Placements::PlacementsQuery.call(params: query_params)
     @pagy, @placements = pagy(placements.merge(query))
+    @placements = @placements.decorate
+
+    Placements::TravelTime.call(
+      origin_address: search_location,
+      destinations: @placements.map(&:school),
+    )
   end
 
   def show
--- a/app/controllers/placements/providers/placements_controller.rb
+++ b/app/controllers/placements/providers/placements_controller.rb
@@ -65,7 +65,7 @@ class Placements::Providers::PlacementsController < Placements::ApplicationContr
 
     Placements::TravelTime.call(
       origin_address: search_location,
-      destinations: @placements.map(&:school),
+      destinations: @placements.map(&:school).uniq, # travel times are attributes on decorated schools
     )
   end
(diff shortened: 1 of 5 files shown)


app/controllers/placements/providers/placements_controller.rb
```diff
@@ -57,6 +60,15 @@ def location_coordinates
     end
   end
 
+  def calculate_travel_time
+    return if search_location.blank?
+
+    Placements::TravelTime.call(
+      origin_address: search_location,
+      destinations: @placements.map(&:school),
```
JamieCleare2525, line 68: Using a `map` is an expensive process.
It should be more efficient to use:
```suggestion
      destinations: Placements::School.where(id: @placements.select(:school_id)),
```
`map` will also create duplicates of the school record, you would need to do `map(&:school).uniq` to prevent the duplicates. Whereas `where` will only return the uniq matching records.

JamieCleare2525: ACTUALLY... I've just tested this myself, and this is a faff if you do it that way. Because `@placements.map(&:school)` passes in the decorated schools associated with the already decorated placements, which is why this works, because the travel times are attributes on the decorated school, not the school model itself. 
At the very least add `.uniq` to the end of the `.map(...)` to remove any duplicated schools.

JamieCleare2525 (approved):
I'm happy with this. I'm annoyed that we can't be more efficient with the queries, but given that the `Placements::TravelTime` will only be `mapping` a maximum of 25 "decorated" school records, I don't think it will be that costly.

ollietreend (approved):
This looks good to me! 👍🏻

PR #1148, opened 2024-10-29
Fix issue deleting placements with windows

JamieCleare2525:
## Context

- Fix issue when deleting placements which are associated with terms through placement windows.

- Add destroy dependancy on placement windows in placement and term models.

## Guidance to review

- Sign in as Anne (School user)
- Find or make a placement with an associated term. (Summer, Autumn or Spring)
- Delete the placement.
- This should be deleted without any issues.

## Link to Trello card

https://trello.com/c/nqbJPqFn/898-investigate-and-resolve-placement-deletion-issue

## Screenshots

https://github.com/user-attachments/assets/8543d211-150b-4db1-b23e-d38e20a23079

### DIFF
--- a/app/models/placement.rb
+++ b/app/models/placement.rb
@@ -32,7 +32,7 @@ class Placement < ApplicationRecord
 
   has_many :placement_additional_subjects, class_name: "Placements::PlacementAdditionalSubject", dependent: :destroy
   has_many :additional_subjects, through: :placement_additional_subjects, source: :subject
-  has_many :placement_windows, class_name: "Placements::PlacementWindow"
+  has_many :placement_windows, class_name: "Placements::PlacementWindow", dependent: :destroy
   has_many :terms, class_name: "Placements::Term", through: :placement_windows
 
   belongs_to :academic_year, class_name: "Placements::AcademicYear"
--- a/app/models/placements/term.rb
+++ b/app/models/placements/term.rb
@@ -25,6 +25,6 @@ class Placements::Term < ApplicationRecord
 
   validates :name, presence: true, inclusion: { in: VALID_NAMES }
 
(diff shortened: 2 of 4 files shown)


app/models/placements/term.rb
```diff
@@ -25,6 +25,6 @@ class Placements::Term < ApplicationRecord
 
   validates :name, presence: true, inclusion: { in: VALID_NAMES }
 
-  has_many :placement_windows, class_name: "Placements::PlacementWindow"
+  has_many :placement_windows, class_name: "Placements::PlacementWindow", dependent: :destroy
```
ollietreend, line 28: How about using `dependent: :restrict_with_exception` here?

**We should never delete terms.** But if we ever accidentally do (e.g. somebody types a dodgy Rails console command), this would give an extra layer of protection so we don't inadvertently cascade delete a large chunk of placement records.

I like the built-in protection this offers. So I think it'd be sensible to make use of it here.

```suggestion
  has_many :placement_windows, class_name: "Placements::PlacementWindow", dependent: :restrict_with_exception
```

app/models/placement.rb
```diff
@@ -32,7 +32,7 @@ class Placement < ApplicationRecord
 
   has_many :placement_additional_subjects, class_name: "Placements::PlacementAdditionalSubject", dependent: :destroy
   has_many :additional_subjects, through: :placement_additional_subjects, source: :subject
-  has_many :placement_windows, class_name: "Placements::PlacementWindow"
+  has_many :placement_windows, class_name: "Placements::PlacementWindow", dependent: :destroy
```
ollietreend, line 35: Whereas for this association, `dependent: :destroy` makes sense 👍🏻

ollietreend (approved):

PR #1150, opened 2024-10-31
Changes to allow wizards to be reuseable across both services

JamieCleare2525:
## Context

- Changes to make wizards reusable across both claims and placements services

- Renaming/moving files so `Wizard` is the top directory
- Changes to the `to_partial_path` on Wizard
- Helper method to render partials for wizards.

## Guidance to review

- Test the placements app and ensure nothing is broken

## Link to Trello card

https://trello.com/c/uKrFn3Kj/907-make-wizard-pattern-reusable-throughout-app


app/controllers/concerns/wizard_controller.rb
```diff
@@ -0,0 +1,31 @@
+module WizardController
+  extend ActiveSupport::Concern
+
+  included do
+    helper_method :current_step_path, :back_link_path
```
ollietreend, line 5: Let's see if we can include `:step_path` here, too. That would tidy up the individual controllers even more.

```suggestion
    helper_method :step_path, :current_step_path, :back_link_path
```

ollietreend (commented):
Really nice work @JamieCleare2525! This massively DRYs up the wizard controllers.

ollietreend (approved):
Chef kiss emoji

PR #1153, opened 2024-11-01
Remove redundant comment from Terraform config

ollietreend:
We dropped the data_migrate gem in PR #1022, but it looks like we forgot to tidy up this reference to the gem.

This comment is no longer helpful since we no longer use the gem.


dp-daly (approved):
I'm happy with this! ;)

PR #1188, opened 2024-11-19
Jc/refactor claims to wizard pattern

JamieCleare2525:
## Context

- Refactor so the multi-step form used to add a claim, now uses the AddClaimWizard

- Add an AddClaimWizard, including steps for each part of the multi-step process.
- Refactor necessary components and views to use the AddClaimWizard

## Guidance to review

- Sign in as Anne (School User)
- Run through the process of adding a claim. (Make sure everything works as expected)
--------
- Sign in as Colin (Support User)
- Run through the process of adding a claim. (Make sure everything works as expected)

## Link to Trello card

https://trello.com/c/MhC5trwI/922-refactor-the-add-claim-form-to-use-the-wizard-pattern?filter=label:Development


app/wizards/claims/add_claim_wizard/mentor_training_step.rb
```diff
@@ -0,0 +1,64 @@
...
+    unless: :custom_hours_selected?,
+  )
+
+  validates(
+    :custom_hours_completed,
+    presence: true,
+    numericality: { only_integer: true },
+    between: { min: 1, max: :max_hours },
+    if: :custom_hours_selected?,
+  )
```
ollietreend, line 25: I wonder if it'd simplify things to model it like this:

```ruby
attribute :hours_to_claim, :string
attribute :custom_hours, :integer
```

Where `hours_to_claim` can either be `"maximum"` or `"custom"`.
And `custom_hours` is either `nil` or an integer.

And then you can have a method:

```ruby
def hours_completed
  hours_to_claim == "maximum" ? max_hours : custom_hours
end
```

That way, you'll only need one `numericality` validation rule rather than duplicating it across two fields, and needing conditional based on `custom_hours_selected?`.

Does that make sense? 🤔 I'm not sure how well I've explained it.

config/locales/en/activemodel.yml
```diff
@@ -135,6 +135,30 @@ en:
...
+            mentor_id:
+              blank: Select a mentor
+            custom_hours_completed:
+              blank: Enter the number of hours
+              not_an_integer: Enter whole numbers only
+              between: Enter the number of hours between %{min} and %{max}
+            hours_completed:
+              blank: Select the number of hours
+              not_an_integer: Enter whole numbers only
+              between: Enter the number of hours between %{min} and %{max}
```
ollietreend, line 157: It would mean you wouldn't need to duplicate the validation messages here, either.

app/wizards/claims/add_claim_wizard/mentor_training_step.rb
JamieCleare2525: Whilst a nice concept, I'm not sure how we would implement this with the form.

app/wizards/claims/add_claim_wizard/mentor_step.rb
```diff
@@ -0,0 +1,21 @@
+class Claims::AddClaimWizard::MentorStep < BaseStep
+  attribute :mentor_ids, default: []
+
+  validates :mentor_ids, presence: true, inclusion: { in: ->(step) { step.mentors_with_claimable_hours.unscoped.ids } }
```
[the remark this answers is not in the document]
JamieCleare2525: Without `unscoped` you get the following error:
`ActiveRecord::StatementInvalid: PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list`
This is because ` Claims::MentorsWithRemainingClaimableHoursQuery` returns an ordered collection and `ids` does a `select` which doesn't include the `order` condition. `unscoped` just removes the ordering.

app/wizards/claims/add_claim_wizard/mentor_training_step.rb
JamieCleare2525: So most of this works, but I can't use `attribute :custom_hours, :integer` because the `govuk_number_field` allows for floats, and `attribute :custom_hours, :integer` converts it immediately to an integer. So `2.5` becomes `2`, with no warning. Which breaks validation.

ollietreend (approved):
Nice work. Looks good to me! 👍🏻

app/services/concerns/claims/claim/referencable.rb
```diff
@@ -0,0 +1,13 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    def generate_reference
+      loop do
+        reference = SecureRandom.random_number(99_999_999)
+
+        break reference unless Claims::Claim.exists?(reference:)
+      end
+    end
```
Kizr, line 11: This is makes me very uncomfortable for many reasons. I know this isn't new logic you've implemented but we should probably look into auto-incrementing the reference on claim creation via the database rather than this.

I'd be interesting to find out why we add a reference to a claim during creation rather than automatically assigning one in general.

app/views/wizards/claims/add_claim_wizard/_mentor_step.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  error: current_step.errors.any?,
+) %>
+
+<%= form_for(current_step, url: current_step_path, method: :put) do |f| %>
+  <%= f.govuk_error_summary %>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <span class="govuk-caption-l"><%= contextual_text %></span>
+      <h1 class="govuk-heading-l"><%= t(".heading", provider_name: @wizard.provider.name) %></h1>
```
Kizr, line 12:
```suggestion
      <h1 class="govuk-heading-l"><%= t(".heading", provider_name: @wizard.provider_name) %></h1>
```

app/views/wizards/claims/add_claim_wizard/_mentor_training_step.html.erb
```diff
@@ -0,0 +1,42 @@
...
+  error: current_step.errors.any?,
+) %>
+
+<%= form_for(current_step, url: current_step_path, method: :put) do |f| %>
+  <%= f.govuk_error_summary %>
+
+   <div class="govuk-grid-row">
+      <div class="govuk-grid-column-two-thirds">
+        <span class="govuk-caption-l"><%= contextual_text %> - <%= @wizard.steps[:provider].provider.name %></span>
+        <h1 class="govuk-heading-l"><%= t(".hours_of_training_for_mentor", mentor: current_step.mentor.full_name) %> </h1>
```
Kizr, line 12: We have some opportunities for delegation here ☺️ 

```suggestion
        <span class="govuk-caption-l"><%= contextual_text %> - <%= @wizard.steps[:provider].provider_name %></span>
        <h1 class="govuk-heading-l"><%= t(".hours_of_training_for_mentor", mentor: current_step.mentor_full_name) %> </h1>
```

app/views/wizards/claims/add_claim_wizard/_no_mentors_step.html.erb
```diff
@@ -0,0 +1,18 @@
...
+  t(".page_title", contextual_text:, provider_name: @wizard.provider.name),
+  error: current_step.errors.any?,
+) %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-l"><%= t(".heading_empty", provider_name: @wizard.provider.name) %></h1>
+
+    <p class="govuk-body">
+      <%= t(".no_mentors_with_claimable_hours", provider_name: @wizard.provider.name) %>
```
Kizr, line 11:
```suggestion
    <h1 class="govuk-heading-l"><%= t(".heading_empty", provider_name: @wizard.provider_name) %></h1>

    <p class="govuk-body">
      <%= t(".no_mentors_with_claimable_hours", provider_name: @wizard.provider_name) %>
```

app/views/wizards/claims/edit_claim_wizard/_declaration_step.html.erb
```diff
@@ -0,0 +1,91 @@
...
+        <% end %>
+      <% end %>
+
+      <h2 class="govuk-heading-m"><%= t("wizards.claims.add_claim_wizard.check_your_answers_step.grant_funding") %></h2>
+
+      <%= govuk_summary_list(actions: false) do |summary_list| %>
+        <% summary_list.with_row do |row| %>
+          <% row.with_key(text: t("wizards.claims.add_claim_wizard.check_your_answers_step.total_hours")) %>
+          <% row.with_value(
+            text: pluralize(@wizard.claim.mentor_trainings.sum(:hours_completed),
```
Kizr, line 61: Is there a method we can use for this given our converstion regarding this logic?

app/views/wizards/claims/edit_claim_wizard/_declaration_step.html.erb
```diff
@@ -0,0 +1,91 @@
...
+        <% end %>
+
+        <% summary_list.with_row do |row| %>
+          <% row.with_key(text: t("wizards.claims.add_claim_wizard.check_your_answers_step.hourly_rate")) %>
+          <% row.with_value(text: humanized_money_with_symbol(@wizard.school.region_funding_available_per_hour)) %>
+        <% end %>
+
+        <% summary_list.with_row do |row| %>
+          <% row.with_key(text: Claims::Claim.human_attribute_name(:claim_amount)) %>
+          <% row.with_value(text: humanized_money_with_symbol(@wizard.claim.amount)) %>
```
Kizr, line 73:
```suggestion
          <% row.with_value(text: humanized_money_with_symbol(@wizard.school_region_funding_available_per_hour)) %>
        <% end %>

        <% summary_list.with_row do |row| %>
          <% row.with_key(text: Claims::Claim.human_attribute_name(:claim_amount)) %>
          <% row.with_value(text: humanized_money_with_symbol(@wizard.claim_amount)) %>
```

Another couple of candidates for delegation

config/locales/en/wizards/claims/add_claim_wizard.yml
```diff
@@ -0,0 +1,51 @@
...
+            remaining: The remaining amount of hours for standard training
+          other_amount: Another amount
+          custom_hours_completed_hint:
+            one: Enter whole numbers up to a maximum of %{count} hour
+            other: Enter whole numbers up to a maximum of %{count} hours
+          number_of_hours: Number of hours
+        check_your_answers_step:
+          page_title: Check your answers - %{contextual_text}
+          title: Check your answers
+          warning: You will not be able to change any of the claim details once you have submitted it.
```
Kizr, line 33:
```suggestion
          warning: You will not be able to change any of the claim details once you have submitted the claim.
```

Personal opinion but I think we should avoid using "it" (I know this is likely pre-existing copy)

Kizr (approved):
dancing-dog

PR #1191, opened 2024-11-20
Remove "rspec-retry" gem and remove its use in specs

dp-daly:
## Context

The rspec-retry gem was added to the codebase in [PR#223](https://github.com/DFE-Digital/itt-mentor-services/pull/223).

The original intent behind this PR was to reduce the flakiness of JS-enabled system specs. In particular, it seems like the autocomplete component used on some pages (for example, when adding a provider) was resulting in intermittent test failures.

However brute-force retrying failed tests is a [code smell](https://en.wikipedia.org/wiki/Code_smell). The gem serves as a sticking plaster to cover over the underlying issue. In this case, the underlying issue is that Capybara isn’t waiting long enough for JS-enabled features to appear asynchronously on the page.

Rather than addressing the symptom, we should fix the underlying problem. This will remove our reliance on the "rspec-retry "gem and resolve the core issue.

Refactor the JS-enabled system specs to replace the third-party rspec-retry gem with built-in Capybara functionality:
- Remove the rspec-retry gem
- Remove [configuration associated with the gem](https://github.com/DFE-Digital/itt-mentor-services/blob/0e3c0c7c893a604f4ce2ade3876a8ee01d5e6f81/spec/spec_helper.rb#L132-L137)
- Remove the retry: tag [from system specs](https://github.com/search?q=repo:DFE-Digital/itt-mentor-services%20%22retry:%22&type=code)
- Add wait: times to individual selectors which we know to be slow

## Guidance to review

- Run `bundle install` to update gems
- Run`rails parallel:spec` to test that all specs still work without the retry gem

## Link to Trello card

https://trello.com/c/grkKmMAM/932-refactor-system-specs-to-remove-their-dependency-on-the-rspec-retry-gem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -150,7 +150,6 @@ group :test do
   gem "capybara-screenshot"
   gem "climate_control"
   gem "rails-controller-testing"
-  gem "rspec-retry", git: "https://github.com/DFE-Digital/rspec-retry.git", branch: "main"
   gem "selenium-webdriver"
   gem "shoulda-matchers"
   gem "simplecov", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,14 +10,6 @@ GIT
       request_store_rails (~> 2)
 
 GIT
-  remote: https://github.com/DFE-Digital/rspec-retry.git
-  revision: 306b42b8d5853303982e1a165e82d04744c20ea3
-  branch: main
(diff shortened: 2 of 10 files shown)


ollietreend (approved):
Great stuff @dp-daly! ⭐ Thanks for doing this.

PR #1194, opened 2024-11-20
Lean into namespaced policies per controller

Nitemaeric:
## Context

While developing the Claim support workflows, all workflows have a sub `ClaimsController`, e.g. `Payments::ClaimsController`, `Samplings::ClaimsController`. There are actions that can be taken on claims which share controller action names, such as `index`. The permissions per each controller differ but they all share the same model. By default, Pundit will always reference the same policy.

To work around this, we need to introduce namespaced policies that mirror controller namespaces. This will allow us to differentiate controller action permissions with shared models without explicitly setting a custom policy class or policy scope class.

- Add `DemodularizesPolicyClass` model concern and include it in `ApplicationRecord`.
- Override Pundit controller methods by [injecting namespaces](https://github.com/varvet/pundit?tab=readme-ov-file#policy-namespacing).
- Rename the `GrantConditions::Claims::SchoolPolicy` to `Claims::GrantConditions::SchoolPolicy` to hoist the service namespace to the outer-most level.

## Guidance to review

> [Lean into namespaced policies per controller](https://github.com/DFE-Digital/itt-mentor-services/commit/ac62f62d8aff7a5b3874253ac462d2430993f65e)
>
> Demodulize model policy class names as string and rebuild the classes by following the guidance in Pundit's README: https://github.com/varvet/pundit?tab=readme-ov-file#policy-namespacing.
> 
> We have service namespaces and then the "Support" namespace within each service. To avoid duplicate service namespaces - as by default `authorize [:claims, :support, Claims::Claim]` would attempt to lookup a `Claims::Support::Claims::ClaimPolicy`. As both the model and controller both have the service namespaces, we need to de-dup them.
> 
> By demodulizing policy class names from models, The `Claims::Claim` model will return a `ClaimPolicy`. This policy does not exist by itself, but we override each of Pundit's controller methods by injecting namespaces.

- Tests should pass.
- Review policies. Non-support policies should only handle non-support actions.


ollietreend (commented):
I understand the problem you're trying to solve, but I'm a little concerned that the approach you've chosen is slightly over-engineered.

The added code to 'unwrap' pundit scopes and 'demodulize' models seems quite complicated to me. It's taken me some time to read through and try and understand what this new code is doing (and if I'm honest, I'm still not 100% sure I can visualise what it's doing).

From looking through the code, and reading the PR description you've written up, I get the impression you're trying to introduce 'automagic' behaviour so the controller implicitly works out which policy scope class to use. It's important that future developers working on this codebase can easily understand which policy scope is being used, and it feels like the approach proposed here would add a layer of indirection to that.

This proposed approach is also a broad change which would apply across the entire app – for example, with the changes you've made to the base Application Controller and Application Record classes – which means we need to consider it in the wider context.

We've had the same need over in the Placements service, but we took a comparatively simpler approach. Rather than building in implicit behaviour, we chose to be more explicit so it's really clear to see which policy is in use.

For example, here are two different controllers which need to display Placement records. But they're both used in different contexts, so warrant standalone policy scopes (similar to the need you've described).

https://github.com/DFE-Digital/itt-mentor-services/blob/78a87a93dc348b7a91fb2c7d1189f32e3e686cb1/app/controllers/placements/schools/placements_controller.rb#L41-L43

https://github.com/DFE-Digital/itt-mentor-services/blob/dbe8f045b382699cf60d5453d0da5fbdf690feca/app/controllers/placements/providers/placements_controller.rb#L27-L32

This makes it really clear when we're asking pundit to deviate from its default behaviour. In this case, I think the added explicitness both helps code readability and makes the overall behaviour of the app much easier to understand for other developers.

Please correct me if I've misunderstood the intention behind this PR. But if the approach taken on Placements could also meet your needs, then I'd ask that we use that. After all, life is always better when we [KISS](https://www.geeksforgeeks.org/kiss-principle-in-software-development/).

Kizr (approved):
Happy for this to be merged when CI is passing 👍

app/controllers/application_controller.rb
```diff
@@ -74,4 +74,8 @@ def user_not_authorized
       redirect_back(fallback_location: root_path)
     end
   end
+
+  def unwrap_pundit_scope(scope)
+    scope.is_a?(Array) ? scope : [scope]
+  end
```
Nitemaeric, line 80: When assigning namespaces to pundit via records and scopes, we are expected to either pass an array of symbols and then the record or scope:

```ruby
policy([namespace, record])
```

The Pundit policy finder uses the last element in the case an array is given as the record: https://github.com/varvet/pundit/blob/main/lib/pundit/context.rb#L173-L175

In the case above, and following Pundit's guidance, this is true. https://github.com/varvet/pundit?tab=readme-ov-file#policy-namespacing

The issue we have occurs with multiple layers of recursion:

```ruby
class Claims::ApplicationController
  def policy(record)
    super([:claims, record])
  end
end
```

This calls the top-level `pundit.policy` with `[:claims, record]`.

But then we do:

```ruby
class Claims::Support::ApplicationController < Claims::ApplicationController
  def policy(record)
    super([:support, record])
  end
end
```

This is to insert the next namespace level into the namespaces, but what we end up calling would be `pundit.policy` with `[:claims, [:support, record]]`. This makes the last element of the array an array itself.

We want to flatten this hierarchy of namespaces. To do this, the `unwrap_pundit_scope` method is introduced.

This might be better named `wrap_pundit_scope` actually...?

Then we flatten the given array or wrapped non-array resource to pass it through.

e.g.

`Claims::Support::ApplicationController#policy` unfolds it's defined `policy` method as follows:


```ruby
policy([:something, claim]) # `claim` is of type `Claims::Claim`
=> policy([:claims, :support, :something, claim]) # Typically `Claims::Claim` would become `Claims::ClaimPolicy`, but we already define the service namespace at the controller, we don't need to use the `Claims` namespace from the model.
=> Claims::Support::Something::ClaimPolicy
```

app/controllers/claims/support/schools/claims/mentors_controller.rb
```diff
@@ -1,20 +1,24 @@
-class Claims::Support::Schools::Claims::MentorsController < Claims::ApplicationController
+class Claims::Support::Schools::Claims::MentorsController < Claims::Support::ApplicationController
```
Nitemaeric, line 1: Found a bug from this work. Support controller inheriting from the non-support application controller.

app/models/concerns/demodulizes_policy_class.rb
```diff
@@ -0,0 +1,9 @@
+module DemodulizesPolicyClass
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def policy_class
+      "#{name.demodulize}Policy"
+    end
+  end
+end
```
Nitemaeric, line 9: This is the logic to demodulize all of the model policies.

This could likely be improve to retain non-service namespaces.

```ruby
def policy_class
  return "#{name.delete_prefix("Claims::")}Policy" if name.starts_with("Claims::")
  return "#{name.delete_prefix("Placements::")}Policy" if name.starts_with("Placements::")
end
```

This would allow automatic lookup for policies of variants of models, like `School::Location` vs `Provider::Location`, if the need arised.

app/views/claims/support/schools/claims/mentors/_form.html.erb
```diff
@@ -33,7 +33,7 @@
         </p>
 
         <p class="govuk-body">
-          <%= govuk_link_to t(".change_provider"), new_claims_school_claim_path(school, id: claim_mentors_form.claim.id) %>
+          <%= govuk_link_to t(".change_provider"), new_claims_support_school_claim_path(school, id: claim_mentors_form.claim.id) %>
```
Nitemaeric, line 36: Support page linking to non-support page.

config/locales/en/placements/support/schools.yml
```diff
@@ -2,8 +2,6 @@ en:
   placements:
     support:
       schools:
-        school_options:
-          cancel: Cancel
```
Nitemaeric, line 6: This key is redefined later in this file.

PR #1195, opened 2024-11-21
Add error details to internal server error page when local

Nitemaeric:
## Context

It's slightly inconvenient to switch to the server logs to see what is causing an internal server error on our local machines - namely when translations are missing.

I want to be told what the error is so I can immediately address it.

- Add error details to internal server error page when in `development`.

## Guidance to review

- Any concerns?

## Screenshots

CleanShot 2024-11-21 at 00 44 14@2x


app/views/errors/internal_server_error.html.erb
```diff
@@ -18,4 +18,18 @@
       </p>
     </div>
   </div>
+
+  <% if Rails.env.development? %>
```
ollietreend, line 22: Just to note that you may have some trouble getting past RSpec + Undercover with your changes to the controller action.

You might have more success leaving the controller untouched and doing this in the view:

```suggestion
  <% if Rails.env.development? && exception = request.env["action_dispatch.exception"]&.cause %>
```

ollietreend (approved):
I'm not entirely convinced this change is necessary, but if you think it'd be helpful then feel free to add it.

PR #1210, opened 2024-11-27
Fix for organisation options step

JamieCleare2525:
## Context

- When the session expires, the `organisations` on the `organisation_options_step` returns `nil` which breaks `.count`

- Change to the `organisations` method to return an empty array `[]` if `wizard.organisation_model.blank?`

## Link to Trello card

https://trello.com/c/YXrsqbXO/955-fix-organisation-count-bug


ollietreend (approved):
This looks good as a quick fix to the specific problem.

I think it could be good – as a separate task – to consider whether there's a more generalised fix which could be applied to Wizard Controllers.

Without thinking it through too much, I think it's fair to say that the `edit` action of a wizard controller _should not render_ if `session[state_key]` is `nil`. It should probably redirect back to the `new` action instead?

PR #1214, opened 2024-11-29
Remove an unused Stimulus JS controller

ollietreend:
## Context

The "select-autocomplete" Stimulus JS controller is no longer in use.

I think it was used in previous iterations of the 'Claim funding for mentor training' designs to enhance `<select>` inputs into autocomplete components. Presumably this JS controller was just never tidied up when the final use of it stopped being used.

Thankfully this change also removes a potential source of confusion for future developers, since there is now only one Stimulus JS controller called "autocomplete" rather than two similarly named controllers.

Remove the unused `"select-autocomplete"` Stimulus JS controller.

## Guidance to review

Do the tests still pass? Every use of autocomplete functionality is already covered by JS-enabled system specs.

## Link to Trello card

https://trello.com/c/nVqhFI01/948-refactor-support-user-onboards-an-organisation-specs


JamieCleare2525 (approved):

Kizr (approved):
😭

PR #1226, opened 2024-12-05
Demonstrate postgres module fix for server version 

saliceti:


saliceti:
[Postgres default container image does not follow var.server_version](https://trello.com/c/IJqJlNga/2178-postgres-default-container-image-does-not-follow-varserverversion)

PR #1240, opened 2024-12-13
Add conditional clawback details on show page

dp-daly:
## Context

We are making progressive enhancements to the support console to assist in the processing of Claims.

- Add the Clawback details to the clawbacks show page (an initial show page was added in `dpd/request-clawback-wizard`)
- Use change links on 'number of hours' and 'reason for clawback' which link to the edit request clawback wizard. This section should only be visible if a clawback has been requested. 
- If the clawback is in progress the Clawback details should be visible, but the change links should not be available.

## Guidance to review

- Navigate to the show page of claims with all three clawback statuses (`sampling_not_approved`, `clawback_requested`, and `clawback_in_progress`.
- Check them against the logic outlined above.
- Note that the values for clawback hours and reason are not yet in place. Changes to the Claims table have been postponed until a further design iteration has been approved and handed over.

## Link to Trello card

https://trello.com/c/eyaECIcP/999-clawbacks-add-clawback-show-page

### DIFF
--- a/app/views/claims/support/claims/_details.html.erb
+++ b/app/views/claims/support/claims/_details.html.erb
@@ -86,7 +86,7 @@
   <% end %>
 <% end %>
 
-  <% if claim.status == "clawback_requested" || claim.status == "clawback_in_progress" %>
+  <% if claim.clawback_requested? || claim.clawback_in_progress? %>
     <h2 class="govuk-heading-m"><%= t(".clawback_details") %></h2>
     <%= govuk_summary_list do |summary_list| %>
       <% summary_list.with_row do |row| %>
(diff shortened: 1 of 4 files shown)


app/views/claims/support/claims/_details.html.erb
```diff
@@ -85,3 +85,41 @@
     <% row.with_value(text: humanized_money_with_symbol(claim.amount)) %>
   <% end %>
 <% end %>
+
+  <% if claim.status == "clawback_requested" || claim.status == "clawback_in_progress" %>
```
Kizr, line 89:
```suggestion
  <% if claim.clawback_requested? || claim.clawback_in_progress? %>
```

When we have enums we gain some useful helper methods for these sorts of checks! What do you think about updating the status logic with these methods?

PR #1290, opened 2025-01-08
Validate presence of action in activity log and conditionally show CSV

JamieCleare2525:
## Context

- Claim Activity records require an action, otherwise the activity log page breaks. Currently this is not validated.

- Add validation to `Claims::ClaimActivity` to ensure `action` is present.

## Guidance to review

(Assuming you have a number of `Claims::ClaimActivity` records in your database.
- Sign in as Colin (Support user)
- Navigate to Claims -> Activity log in the nav bar.
- The page should not break whether or not you have a `csv_file` attached to the record.

_In the rails console_
- Try to create a `Claims::ClaimActivity` record without an `action`
- It should error due to the validation.

## Link to Trello card

https://trello.com/c/viXc3m6A/338-fix-activity-log-when-csv-files-are-not-present

## Screenshots
screencapture-claims-localhost-3000-support-claims-activity-2025-01-08-13_33_41

### DIFF
--- a/app/models/claims/claim_activity.rb
+++ b/app/models/claims/claim_activity.rb
@@ -32,7 +32,7 @@ class Claims::ClaimActivity < ApplicationRecord
     sampling_response_uploaded: "sampling_response_uploaded",
     clawback_request_delivered: "clawback_request_delivered",
     clawback_response_uploaded: "clawback_response_uploaded",
-  }
+  }, validate: true
 
   delegate :full_name, to: :user, prefix: true, allow_nil: true
 end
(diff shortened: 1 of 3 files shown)


app/models/claims/claim_activity.rb
```diff
@@ -23,6 +23,8 @@ class Claims::ClaimActivity < ApplicationRecord
   belongs_to :user, class_name: "Claims::SupportUser"
   belongs_to :record, polymorphic: true
 
+  validates :action, presence: true
```
Nitemaeric, line 26: Alternatively, you could use the enum validation too.

CleanShot 2025-01-08 at 16 03 17@2x

Kizr (approved):

PR #1309, opened 2025-01-15
Migrate provider emails to ProviderEmailAddress records

JamieCleare2525:
## Context

- Migrate provider emails to new ProviderEmailAddress records

- Add model ProviderEmailAddress
- Change provider importer to create ProviderEmailAddress records

## Guidance to review

- run migrations
- ProviderEmailAddress table is added
- Provider emails are migrated to ProviderEmailAddress records

-------

- run `bundle exec rake db:drop db:create db:setup`
- Seed should use the import and create ProviderEmailAddress records

## Link to Trello card

https://trello.com/c/FbmyXt7z/345-allow-providers-to-have-multiple-email-addresses

### DIFF
--- a/db/migrate/20250114161953_migrate_provider_email_addresses.rb
+++ b/db/migrate/20250114161953_migrate_provider_email_addresses.rb
@@ -1,5 +1,7 @@
 class MigrateProviderEmailAddresses < ActiveRecord::Migration[7.2]
   def up
+    return unless Provider.column_names.include?("email_address")
+
     email_attributes = Provider.select(:email_address, :id)
       .map do |provider|
         { provider_id: provider.id, email_address: provider.email_address }
@@ -11,6 +13,6 @@ class MigrateProviderEmailAddresses < ActiveRecord::Migration[7.2]
   end
 
   def down
-    ProviderEmailAddress.destroy_all
+    raise ActiveRecord::IrreversibleMigration
   end
 end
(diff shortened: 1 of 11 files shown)


db/migrate/20250114161953_migrate_provider_email_addresses.rb
```diff
@@ -0,0 +1,16 @@
...
+        { provider_id: provider.id, email_address: provider.email_address }
+      end
+    ProviderEmailAddress.upsert_all(
+      email_attributes,
+      unique_by: :unique_provider_email,
+    )
+  end
+
+  def down
+    ProviderEmailAddress.destroy_all
```
Kizr, line 14: ⚠️ This would leave us without any record of provider email addresses after a successful migration!

Please make this an irreversible migration 😄

Kizr (changes requested):

db/migrate/20250114161953_migrate_provider_email_addresses.rb
JamieCleare2525: That's not correct. The email address is still an attribute in the provider records.

Kizr: For now, but not after we remove it. My concern is that we could end up losing data in the future, this feels like it should be irreversible to avoid any mistakes

Kizr (approved):

PR #1366, opened 2025-02-10
Allow contextual actions from all claims

Kizr:
## Context

Support users need to be able to take contextual payments/auditing/clawbacks actions from the All Claims page

- [x] Move all contextual actions into partials
- [x] Create actions component to condinionally render the correct partial based off of the claim status

## Guidance to review

- Log in as Colin
- Create a range of claim statuses
- Verify that the actions from the context areas match the actions from all claims for each status

## Link to Trello card

[Allow contextual actions from All claims page
](https://trello.com/c/6NO1tHB2/403-allow-contextual-actions-from-all-claims-page)

## Screenshots

image
image
image
image
image


app/views/claims/support/claims/payments/claims/_actions.html.erb
```diff
@@ -0,0 +1,22 @@
+<%# locals: (claim:) %>
+
+<% if claim.unpaid_reason.present? %>
+  <%= govuk_inset_text do %>
+    <h2 class="govuk-heading-s"><%= t(".unpaid_reason_header") %></h2>
+    <p class="govuk-body"><%= claim.unpaid_reason %></p>
+  <% end %>
+<% end %>
```
JamieCleare2525, line 8: Does this belong here? It's not related to the actions. Perhaps this should be in a responses partial?

app/views/claims/support/claims/payments/claims/_actions.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <%= govuk_button_link_to t(".buttons.reject"), confirm_reject_claims_support_claims_payments_claim_path(claim), secondary: true %>
+  </div>
+<% end %>
+
+<% if claim.payment_information_sent? %>
+  <div class="govuk-button-group">
+    <%= govuk_button_link_to t(".buttons.paid"), confirm_paid_claims_support_claims_payments_claim_path(claim) %>
+    <%= govuk_button_link_to t(".buttons.reject"), confirm_reject_claims_support_claims_payments_claim_path(claim), secondary: true %>
+  </div>
+<% end %>
```
JamieCleare2525, line 22: I would separate these out into different partials. That way you minimise the `if` checks and only render the actions suitable to that status.

app/views/claims/support/claims/samplings/_actions.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <div class="govuk-button-group">
+    <%= govuk_button_link_to t(".buttons.approve"), confirm_approval_claims_support_claims_sampling_path(claim) %>
+    <%= govuk_button_link_to t(".buttons.confirm_provider_reject"), new_provider_rejected_claims_support_claims_sampling_path(claim), secondary: true %>
+  </div>
+<% elsif claim.sampling_provider_not_approved? %>
+  <div class="govuk-button-group">
+    <%= govuk_button_link_to t(".buttons.approve"), confirm_approval_claims_support_claims_sampling_path(claim) %>
+    <%= govuk_button_link_to t(".buttons.reject_claim"), new_rejected_claims_support_claims_sampling_path(claim), secondary: true %>
+  </div>
+<% end %>
```
JamieCleare2525, line 27: There is a lot of `if` logic being handled in this view, which could be minimisedd by separating the logic into separate partials per status.

app/views/claims/support/claims/payments/claims/_actions.html.erb
Kizr: Originally I thought it didn't, but after I read your comment I refactored the logic and moved this out into its own partial.

Once I'd done this I realised that due to the way the controller is structured this information is scoped to the same statuses as the partial is, it's just not explicitly scoped.

As a result I think that we should leave this alone and tackle this issue when we address the tech debt aspect of removing the additional views.

https://trello.com/c/yCKEtagO/412-consolidate-actions-component-logic-and-remove-duplicate-views will track this work.

app/views/claims/support/claims/payments/claims/_actions.html.erb
Kizr: Will address in https://trello.com/c/yCKEtagO/412-consolidate-actions-component-logic-and-remove-duplicate-views

app/views/claims/support/claims/samplings/_actions.html.erb
Kizr: Will address in https://trello.com/c/yCKEtagO/412-consolidate-actions-component-logic-and-remove-duplicate-views

JamieCleare2525 (approved):
Approved, with the intention of revisiting this later.

PR #1384, opened 2025-02-13
Consolidate actions component

JamieCleare2525:
## Context

- Consolidate the actions component to use 1 HTML file.

- Make the ActionsComponent use 1 HTML file.
- Add ResponsesComponent to show claim/mentor training responses.

## Guidance to review

- Sign in as Colin (Support user)
- Navigate to "Claims"
- Visit each show page to ensure the correct buttons and responses are visible.

## Link to Trello card

https://trello.com/c/yCKEtagO/412-consolidate-actions-component-logic-and-remove-duplicate-views

## Screenshots

Screenshot 2025-02-14 at 09 53 02
Screenshot 2025-02-14 at 09 59 11
Screenshot 2025-02-14 at 09 58 02
Screenshot 2025-02-14 at 09 57 30
Screenshot 2025-02-14 at 09 56 58
Screenshot 2025-02-14 at 09 56 23
Screenshot 2025-02-14 at 09 55 59
Screenshot 2025-02-14 at 09 53 45
Screenshot 2025-02-14 at 09 53 02


app/components/claims/support/claim/responses_component.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def initialize(claim:, classes: [], html_attributes: {})
+    super(classes:, html_attributes:)
+
+    @claim = claim
+  end
+
+  def provider_response
+    not_assured_mentor_trainings = mentor_trainings.not_assured
+    return "" if not_assured_mentor_trainings.blank?
```
Kizr, line 12: If I recall correctly, I think that we wanted to show nothing rather than a blank box here? It might be worth changing this to `nil` and then rendering the details based on that value, possibly go a step further and create a method that's responsible for the check?

We probably want to confirm this with Nao, at the very least I'd expect a message for the user to tell that there is no response if we're showing the inset text regardless?

app/components/claims/support/claim/responses_component.rb
```diff
@@ -0,0 +1,47 @@
...
+            "#{mentor_training.mentor_full_name}: #{mentor_training.reason_not_assured}",
+          ),
+        )
+      end
+    end
+  end
+
+  def school_response
+    rejected_mentor_trainings = mentor_trainings.rejected
+    return "" if rejected_mentor_trainings.blank?
```
Kizr, line 28: If I recall correctly, I think that we wanted to show nothing rather than a blank box here? It might be worth changing this to `nil` and then rendering the details based on that value, possibly go a step further and create a method that's responsible for the check?

We probably want to confirm this with Nao, at the very least I'd expect a message for the user to tell that there is no response if we're showing the inset text regardless?

Kizr (approved):

PR #1387, opened 2025-02-17
Refactor spec/system/placements/support/users/support_user_views_a_user_spec

JamieCleare2525:
## Context

- Refactor `spec/system/placements/support/users/support_user_views_a_user_spec.rb` to use the new style guide.

## Link to Trello card

https://trello.com/c/qlsDGgKQ/414-refactor-spec-system-placements-support-users-supportuserviewsauserspecrb-to-new-style


spec/system/placements/support/users/school_users/support_user_views_the_details_of_a_school_user_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+      "Name" => "John Smith",
+      "Email address" => @user_john_smith.email,
+    })
+  end
+
+  def when_i_click_on_joe_bloggs
+    click_on "Joe Bloggs"
+  end
+
+  def then_i_see_the_user_details_for_joe_bloggs
```
Kizr, line 96: Could we also test for the presence of the title, h1 and delete link for this user?

spec/system/placements/support/users/school_users/support_user_views_the_details_of_a_school_user_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    click_on "Ashford School"
+  end
+
+  def and_i_click_users_in_the_primary_navigation
+    within(primary_navigation) do
+      click_on "Users"
+    end
+  end
+
+  def then_i_see_the_users_index_page_for_ashford_school
```
Kizr, line 64: There is also an "Add user" button and some help text regarding what a user is, but I think it's okay to _not_ test these in this test as it feels like we're really testing the show page. At the moment we're testing the index page more than the show page though.

Kizr (changes requested):
Same comment for both specs really, please could we increase the scrutiny of the test for the show pages?

Kizr (approved):

PR #1510, opened 2025-04-23
Ba/redirect to eoi for new schools

Kizr:


JamieCleare2525 (approved):
I've voiced my concerns with @Kizr regarding what the behaviour should be when the next academic year begins.

PR #1514, opened 2025-04-23
Remove placements page from provider interface

richardpattinson:
Add show_provider_placements feature flag so that this functionality may be re-enabled in future

## Context

When accessing the service as a provider the landing page is ‘Find a placement’

Note that we are not removing the logic for the time being but a fair few system specs are going to break as they’ll use this page as a “jumping off” point. Replace the physical “click on navigation” action for these specs with the path instead.

Once we’ve launched the private beta we’ll assess if we can chop this logic entirely. We should also add a policy that blocks users accessing these pages via URL’s (that can be stubbed in the specs)

Add feature flag to show provider placements. When it is set to false; the placements page will not be visible.

## Guidance to review

Run the review app and try to navigate the pages with/without the feature flag.

## Link to Trello card

(https://trello.com/c/Ck4VbN9n/524-remove-find-a-placement-page-from-the-provider-view)

### DIFF
--- a/app/controllers/placements/organisations_controller.rb
+++ b/app/controllers/placements/organisations_controller.rb
@@ -32,7 +32,7 @@ class Placements::OrganisationsController < Placements::ApplicationController
     if organisation.is_a?(School)
       placements_school_placements_path(organisation)
     else # Provider
-      placements_provider_placements_path(organisation)
+      placements_provider_find_index_path(organisation)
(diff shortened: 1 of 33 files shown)


Kizr:
@richardpattinson this one may not be necessary due to changes from design, leaving open for now but I may close later today, sorry!

PR #1517, opened 2025-04-28
Ba/show all schools find a school

Kizr:

### DIFF
--- a/app/components/placements/schools/interest_tag_component.rb
+++ b/app/components/placements/schools/interest_tag_component.rb
@@ -17,10 +17,11 @@ class Placements::Schools::InterestTagComponent < ApplicationComponent
 
   private_constant :INTEREST_COLOURS, :INTEREST_TEXT
 
-  def initialize(school:, classes: [], html_attributes: {})
+  def initialize(school:, academic_year:, classes: [], html_attributes: {})
     super(classes:, html_attributes:)
 
     @school = school
+    @academic_year = academic_year
   end
 
   def call
@@ -29,7 +30,7 @@ class Placements::Schools::InterestTagComponent < ApplicationComponent
 
   private
 
-  attr_reader :school
+  attr_reader :school, :academic_year
 
   def interest_colour
     INTEREST_COLOURS[calculated_status]
@@ -44,7 +45,7 @@ class Placements::Schools::InterestTagComponent < ApplicationComponent
       "filled_placements"
     elsif actively_looking? && has_available_placements?
       "unfilled_placements"
-    elsif actively_looking?
+    elsif open?
       "open"
     elsif not_looking?
       "not_open"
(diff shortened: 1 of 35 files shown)


app/components/placements/schools/summary_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <li class="app-search-results__item">
   <h2 class="govuk-heading-m">
-    <% if !open_to_hosting? %>
+    <% if !open_to_hosting? && !interested_in_hosting? %>
```
JamieCleare2525, line 3: Why not `if actively_looking?`?

app/components/placements/schools/interest_tag_component.rb
```diff
@@ -54,18 +55,22 @@ def calculated_status
...
+  end
+
+  def open?
+    school.current_hosting_interest(academic_year:)&.appetite == "interested"
   end
 
   def not_looking?
-    school.current_hosting_interest_appetite == "not_open"
+    school.current_hosting_interest(academic_year:)&.appetite == "not_open"
   end
```
JamieCleare2525, line 67: Might be something we can cleanup later, but I think methods on the `HostingInterest` model for `actively_looking?`, `interested?` and `not_open?` might be worth doing.

app/components/placements/schools/summary_component.html.erb
```diff
@@ -51,26 +51,32 @@
...
+              <%= t(".historically_offered_placements") %>
+            <% else %>
+              <%= t(".last_offered_count", count: previous_academic_year_placements_count, academic_year_name: academic_year.previous.name) %>
+            <% end %>
           </dd>
         </div>
       <% else %>
-        <% unless school.current_hosting_interest_appetite == "actively_looking" && (school.unavailable_placements.empty? && school.available_placements.empty?) %>
-          <% unless school.current_hosting_interest_appetite == "actively_looking" && (school.unavailable_placements.exists? && school.available_placements.empty?) %>
+        <% unless (current_hosting_interest_appetite == "actively_looking" && school.unavailable_placements(academic_year:).empty? && school.available_placements(academic_year:).empty?) || interested_in_hosting? %>
```
JamieCleare2525, line 78: no `open_to_hosting?` method? Makes would make sense given `interested_in_hosting?` and `not_open_to_hosting?`. To would then replace `current_hosting_interest_appetite == "actively_looking"`

app/components/placements/schools/summary_component.html.erb
```diff
@@ -94,7 +100,7 @@
             <%= t(".last_offered") %>
           </dt>
           <dd class="app-result-detail__value">
-            <%= t(".last_offered_count", count: previous_academic_year_placements_count, academic_year_name: Placements::AcademicYear.current.previous.name) %>
+            <%= t(".last_offered_count", count: previous_academic_year_placements_count, academic_year_name: @academic_year.previous.name) %>
```
JamieCleare2525, line 103:
```suggestion
            <%= t(".last_offered_count", count: previous_academic_year_placements_count, academic_year_name: academic_year.previous.name) %>
```

app/components/placements/schools/summary_component.html.erb
Kizr: Ah yeah this gets messy because this isn't the same as `open_to_hosting?` I need to go back through all of this and method it up better after launch :/

JamieCleare2525 (approved):
Previously mentioned comments will be addressed in Tech Debt sweep.

PR #1570, opened 2025-05-15
Change back link href on schools placements show page

richardpattinson:
## Context

After updating an individual placements, the back button doesn't take user to placement summary page but asks the user to complete the action again. This is incorrect as the user should use the change links should a detail need to be amended again. The school placements show page backlink should always point to the placements index.  The system spec for editing a placement has been extended to cover this behaviou.

## Guidance to review

Test the backlink. Run the specs. 

## Link to Trello card

https://trello.com/c/Nfp4pzS5/602-back-return-location-after-adding-or-updating-an-individual-placement


Kizr:
@richardpattinson please can you add yourself as an assignee and put a "deploy" tag on this PR so we can test it in a review environment?

Kizr (approved):

PR #1573, opened 2025-05-16
Add categories to secondary subjects on the Add placement flow

richardpattinson:
## Context

Subject categories should be added to the add placment flow in order to keep them consistent with the multi-placment flow, and to align with Find (and their user research on readability). This was accomplished by splitting the subject step of the add placements wizard into two, one for primary subjects and one for secondary subjects.  

Add placement subject_step split into primary_subject_selection_step and secondary_subject_selection_step. The subject category divisions were added to the secondary_subject_selection_step. Additionally content changes were made to the secondary_subject_selection_step to bring it in line with the information presented in the multi-placement wizard.

## Guidance to review

Go through the add a placment flow with schools in primary, secondary and all through phases. 
 
## Link to Trello card

https://trello.com/c/agS8POxd/587-sweep-add-categories-to-secondary-subjects-on-the-single-add-placement-journey
## Screenshots

Add Placement secondary_school_selection_page should look as below: 
Screenshot 2025-05-19 at 16 38 28


app/views/wizards/placements/add_placement_wizard/_primary_subject_selection_step.html.erb
richardpattinson, line 1: This is a copy of the current subject_step. I have left the original in currently but will remove it once I have fixed the specs.

app/wizards/placements/add_placement_wizard/primary_subject_selection_step.rb
```diff
@@ -0,0 +1,9 @@
+class Placements::AddPlacementWizard::PrimarySubjectSelectionStep < Placements::AddPlacementWizard::SubjectSelectionStep
+  def subjects_for_selection
+    @subjects_for_selection = Subject.parent_subjects.primary
+  end
+
+  def step_name
+    :primary_subject_selection
+  end
```
JamieCleare2525, line 8: This is unnecessary as `step_name` if determined within the `app/wizards/base_wizard.rb` model.

app/wizards/placements/add_placement_wizard/secondary_subject_selection_step.rb
```diff
@@ -0,0 +1,13 @@
...
+           :art_humanities_social_subjects, :health_physical_education_subjects,
+           to: :subjects_for_selection
+
+  def subjects_for_selection
+    @subjects_for_selection = Subject.parent_subjects.secondary
+  end
+
+  def step_name
+    :secondary_subject_selection
+  end
```
JamieCleare2525, line 12: This is unnecessary as step_name if determined within the `app/wizards/base_wizard.rb` model.

app/wizards/placements/add_placement_wizard/primary_subject_selection_step.rb
richardpattinson: I think that method is private, rather than deprivatising it isnt it better to have a separate method that can be used only on this class?

richardpattinson: If there is no issue with deprivatising it then im happy to do that

richardpattinson: Additionally; if the primary_subject_selection_step is soon to be removed then this method will be taken out anyway; as there will be no need to conditionally pull out the step_name for the step_path method in the check_your_answers_step. Happy to deprivatise step_name, but we will probably end up re-privatising it when the primary_subject_selection_step is removed.

JamieCleare2525: So, because in the `app/wizards/placements/add_placement_wizard.rb` methods `year_group_steps` and `secondary_subject_steps`, you are specifically adding `PrimarySubjectSelectionStep` and `SecondarySubjectSelectionStep`. Therefore, the `step_name` generated is `:primary_subject_selection` and `:secondary_subject_selection`, respectively.

As an experiment to prove this, remove the `step_name` methods from the `app/wizards/placements/add_placement_wizard/secondary_subject_selection_step.rb` and `app/wizards/placements/add_placement_wizard/primary_subject_selection_step.rb` classes and run `bundle exec rspec spec/wizards/placements/add_placement_wizard_spec.rb`. You'll see that the `steps` returned are what you expected, without the need to specify each `step_name` in each class.

JamieCleare2525: Whilst we will be removing the primary subject selection step from the wizard, it's still worth doing as separate PRs. Otherwise, it will muddy the waters of what has been requested and resolved.
Obviously, feel free to branch off of this PR should it not be in `main` by the time the ticket comes around.

richardpattinson: > So, because in the `app/wizards/placements/add_placement_wizard.rb` methods `year_group_steps` and `secondary_subject_steps`, you are specifically adding `PrimarySubjectSelectionStep` and `SecondarySubjectSelectionStep`. Therefore, the `step_name` generated is `:primary_subject_selection` and `:secondary_subject_selection`, respectively.
> 
> As an experiment to prove this, remove the `step_name` methods from the `app/wizards/placements/add_placement_wizard/secondary_subject_selection_step.rb` and `app/wizards/placements/add_placement_wizard/primary_subject_selection_step.rb` classes and run `bundle exec rspec spec/wizards/placements/add_placement_wizard_spec.rb`. You'll see that the `steps` returned are what you expected, without the need to specify each `step_name` in each class.

I understand that the step_name isn't needed to assign the step_name to the wizard steps or for the class to function. Its inclusion is that it allows me to remove logic from the view in the _check_your_answers_step.html.erb when a step name as a symbol is required for the step_path method.  I could do something similar by replacing 
`wizard.subject_step.step_name` 
with: 
`wizard.steps[:primary_subject_selection_step] || wizard.steps[:secondary_subject_selection_step]`
But i thought it better to reuse the logic used in every other case to determine which step is applicable to the current wizard. 


Screenshot 2025-05-20 at 16 11 31

JamieCleare2525: `wizard.subject_step.step_name` will still work without the methods in the `PrimarySubjectSelectionStep` and `SecondarySubjectSelectionStep` classes.
```
def subject_step
   case placement_phase
    when School::PRIMARY_PHASE
      steps.fetch(:primary_subject_selection)
    when School::SECONDARY_PHASE
      steps.fetch(:secondary_subject_selection)
    end
  end
```

Your implementation of `subject_step` will return the appropriate step. `step_name` (without your additional implementation) will return the same results you expect

richardpattinson: My mistake then. Thanks for the clarification

app/wizards/placements/add_placement_wizard/secondary_subject_selection_step.rb
richardpattinson: Should be resolved as above

app/wizards/placements/add_placement_wizard.rb
```diff
@@ -53,5 +74,15 @@ def build_placement
...
+
+    def year_group_steps
+      add_step(PrimarySubjectSelectionStep)
+      add_step(YearGroupStep)
+    end
+
+    def secondary_subject_steps
+      add_step(SecondarySubjectSelectionStep)
+      add_step(AdditionalSubjectsStep) if steps.fetch(:secondary_subject_selection).subject_has_child_subjects?
+    end
```
JamieCleare2525, line 86: Not a blocker, but these method names are misleading. Given that for example `year_group_steps` does more than assign year groups.

`primary_steps` and `secondary_steps` would be more appropriate, as the methods determine the steps used after selection `primary` or `secondary`

app/wizards/placements/add_placement_wizard.rb
```diff
@@ -24,6 +27,24 @@ def placement_phase
...
+
+    def subject_step_name
+      case placement_phase
+      when School::PRIMARY_PHASE
+        :primary_subject_selection
+      when School::SECONDARY_PHASE
+        :secondary_subject_selection
+      end
+    end
```
JamieCleare2525, line 46:
```suggestion
    def subject_step_name
     step_name(subject_step)
    end
```

app/wizards/placements/add_placement_wizard.rb
richardpattinson: Yeah that makes a lot of sense. I've updated the method names.

app/wizards/placements/add_placement_wizard.rb
richardpattinson: Another good change. I've put it in

JamieCleare2525 (approved):

PR #1585, opened 2025-05-21
639 remove primary subjects step

richardpattinson:
## Context

There isn’t really a concept of primary subjects so far as a primary schools are concerned. These primary subjects were taken from the register subject list but that doesn’t necessarily reflect the reality of how schools use this data. We should remove the unnecessary step from the flow and default all Primary phase placements to the "Primary" subject. 

- Remove the primary subjects step from the the AddPlacementWizard

- Default the subject for primary schools (or schools that have selected the primary path) to be the “Primary”* subject

- Remove the subject row in the check_your_answers_step for primary phase placements.
After conferring with design as there is no choice of subject for a primary placement anymore there should not be a row on the check your answers page with subject listed. There is also no need for the resulting change link that would have been displayed on this page

- Fix any tests that break as a result of these changes

## Changes proposed in the ticket but NOT included in this pull request

Create a migration that converts all existing “Primary with X”** subjects to be “Primary”

^ The above step should be completed in a seperate PR after both the multi-add-placement flow and the add placement flow have been completed. We dont want to turn all the records from Primary with X -> Primary only to have more Primary with X added afterwards and have the migration run again. 

## Guidance to review

Run through the add placement flow as a primary, secondary and all through school. 

## Link to Trello card

https://trello.com/c/ZZ22hpNz/639-remove-primary-subjects-step


JamieCleare2525 (commented):
Was the PR branched off of https://github.com/DFE-Digital/itt-mentor-services/pull/1573? 
If yes, could you change the base on this PR?

Kizr:
@richardpattinson I've looked into the primary subjects on the multi-placement subject and we don't actually display them there, so it's safe for the migration to be included in this PR

richardpattinson:
> @richardpattinson I've looked into the primary subjects on the multi-placement subject and we don't actually display them there, so it's safe for the migration to be included in this PR

I've added the migration.

db/migrate/20250523154716_set_placements_subject_to_primary.rb
```diff
@@ -0,0 +1,16 @@
+class SetPlacementsSubjectToPrimary < ActiveRecord::Migration[7.2]
+  def up
+    primary_subject = Subject.find_by(name: "Primary", subject_area: "primary")
+    unless primary_subject
+      raise ActiveRecord::IrreversibleMigration, "No subject found with name 'Primary' and subject_area 'primary'"
+    end
+
+    other_primary_subjects = Subject.where(subject_area: "primary").where.not(id: primary_subject.id)
```
Kizr, line 8: I left some information on the Trello card on how to safely query the relevant subjects, please can you use these queries?

We may also want to make use of a `subject_scope` variable which would be `Subject.primary` and then we can chain our queries off of that and avoid duplication, e.g. `subject_scope.find_by(name: "Primary", code: "00")`

See my comment here:

> Historically we’ve allowed schools to choose their primary subject which means we may have historic data that needs to be migrated.
> 
> *Primary subject: Subject.primary.find_by(name: "Primary", code: "00")
> 
> **Primary with X subjects: Subject.primary.where.not(name: "Primary", code: "00")
 
I don't think it makes sense to raise an `ActiveRecord::IrreversibleMigration` error in the `unless` clause, the error doesn't make sense in this context, instead this would probably be an `ActiveRecord::RecordNotFound` error instead. 

I think that it's okay to raise an error here, however we should check that the record exists via the production console before rolling this out to be sure as a failure at this point will require a bit of work to rollback should it fail.

db/schema.rb
```diff
@@ -555,10 +555,10 @@
     t.string "local_authority_code"
     t.datetime "claims_grant_conditions_accepted_at"
     t.uuid "claims_grant_conditions_accepted_by_id"
-    t.jsonb "potential_placement_details"
     t.string "vendor_number"
     t.boolean "expression_of_interest_completed", default: false
     t.boolean "previously_offered_placements", default: false
+    t.jsonb "potential_placement_details"
```
Kizr, line 561: I don't think change beongs in this PR?

Kizr (changes requested):

db/migrate/20250523154716_set_placements_subject_to_primary.rb
```diff
@@ -0,0 +1,11 @@
+class SetPlacementsSubjectToPrimary < ActiveRecord::Migration[7.2]
+  def up
+    other_primary_subjects = Subject.primary.where.not(id: Subject.primary_subject.id)
+
+    Placement.where(subject_id: other_primary_subjects.pluck(:id)).update_all(subject_id: Subject.primary_subject.id)
```
Kizr, line 5: Last minor tweak here before I can approve, please can we use `.select(:id)` instead of `.pluck(:id)` as we don't need any of the other data!

richardpattinson: That makes sense, ive put select in

PR #1589, opened 2025-05-22
Fix claim guidance warning

Kizr:
## Context

The message was using the start and end years from the claim window, rather than the academic year which was causing confusion for users.

- Replaced the claim window `starts_on` and `ends_on` year for the academic year equivalent
- Delegates `starts_on` and `ends_on` to the academic year with prefix

## Guidance to review

- Log in as Anne
- Check that the warning message references the academic year start and end date

## Link to Trello card

[Typo on CFMT claim journey](https://trello.com/c/fbxy3PZs/628-typo-on-cfmt-claim-journey)

## Screenshots

image


JamieCleare2525 (approved):
Thanks for adding delegations

PR #1638, opened 2025-06-11
Ba/do not show niot orgs

Kizr:


richardpattinson (approved):
Looks good! The NIoT providers should be reviewed from the support find page, but that can happen in a seperate PR.

JamieCleare2525 (approved):

Kizr:
> Looks good! The NIoT providers should be reviewed from the claims support find page, but that can happen in a seperate PR.

I'll sort a card out, but it's fairly low priority as support won't use these filters for a few months yet. Good catch though

PR #1642, opened 2025-06-16
Add page to display manually onboarded schools

Kizr:
## Context

Claims needs a way to view which schools have been manually onboarded so that they can be cross referenced against whether they should have been.

- Adds the manually onboarded schools page to the support settings page

## Guidance to review

- Log in as Colin
- Click on settings
- View the Manually onboarded schools page

## Link to Trello card

[Add support console page to view manually onboarded schools](https://trello.com/c/bv50tUj7/710-add-support-console-page-to-view-manually-onboarded-schools)

## Screenshots

image
image

### DIFF
--- a/app/views/claims/support/manually_onboarded_schools/index.html.erb
+++ b/app/views/claims/support/manually_onboarded_schools/index.html.erb
@@ -5,39 +5,45 @@
 <% end %>
 
 <div class="govuk-width-container">
-  <h1 class="govuk-heading-l"><%= t(".heading") %></h1>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-full">
+      <h1 class="govuk-heading-l"><%= t(".heading") %></h1>
 
-  <% if @manually_onboarded_schools.exists? %>
-    <%= govuk_table do |table| %>
-      <% table.with_head do |head| %>
-        <% head.with_row do |row| %>
-          <% row.with_cell text: t(".table.headers.school_name") %>
-          <% row.with_cell text: t(".table.headers.onboarded_by") %>
-          <% row.with_cell text: t(".table.headers.eligible_for") %>
-        <% end %>
-      <% end %>
+      <% if @manually_onboarded_schools.exists? %>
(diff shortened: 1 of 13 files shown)


app/views/claims/support/manually_onboarded_schools/index.html.erb
```diff
@@ -0,0 +1,43 @@
+<% render "claims/support/primary_navigation", current: :settings %>
+
+<%= content_for(:before_content) do %>
+  <%= govuk_back_link(href: claims_support_settings_path) %>
+<% end %>
+
+<div class="govuk-width-container">
```
JamieCleare2525, line 7: Shouldn't this now be followed by:
```
<div class="govuk-grid-row">
  <div class="govuk-grid-column-full">
```

spec/system/claims/support/settings/support_user_views_manually_onboarded_schools_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def when_i_navigate_to_the_settings_index_page
+    within(primary_navigation) do
+      click_on "Settings"
+    end
+  end
+
+  def then_i_see_the_manually_onboarded_schools_link
+    expect(page).to have_link("Manually onboarded schools", href: "/support/manually_onboarded_schools")
```
JamieCleare2525, line 41: Not a blocker, but you can use the path

JamieCleare2525 (changes requested):
You've missed adding the `manually_onboarded_by` to the `app/wizards/claims/add_school_wizard.rb` wizard. You are only recording the user if they onboarding the user using the vendor ID.

spec/system/claims/support/settings/support_user_views_manually_onboarded_schools_spec.rb
Kizr: I could! Does this go against the "No logic in system specs unless necessary" rule though? 🤔

JamieCleare2525 (approved):

PR #1672, opened 2025-06-27
Support console performance improvements

Kizr:


app/policies/claims/support/provider_policy.rb
```diff
@@ -0,0 +1,5 @@
+class Claims::Support::ProviderPolicy < Claims::ApplicationPolicy
+  def search?
+    user.support_user?
+  end
+end
```
JamieCleare2525, line 5: Is this necessary given the `support_user?` check in `Claims::Support::ApplicationController`

JamieCleare2525: In truth, I have no issue with this. Just making sure the perspective of the previous development team is considered.

JamieCleare2525 (approved):

PR #1682, opened 2025-07-02
Update New session page to highlight different login methods

richardpattinson:
## Context

Quite a few users need to use both these services. Some users are getting confused and trying to access MSP using their Manage ECT credentials - causing user frustration and adding to our support queries.

Update content to reflect multiple login methods that exist and which is applicable.


app/views/sessions/new.html.erb
```diff
@@ -3,7 +3,9 @@
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l"><%= t(".heading", service_name:) %></h1>
 
-      <p class="govuk-body"><%= t(".description") %></p>
+      <p class="govuk-body"><%= t(".pilot_service") %></p>
+
+      <p class="govuk-body"><%= t(".access_your_account") %></p>
```
JamieCleare2525, line 8: This needs to be within conditional logic to only show for the claims service.

```suggestion
<% if current_service == :placements %>
      <p class="govuk-body"><%= t(".pilot_service") %></p>

      <p class="govuk-body"><%= t(".access_your_account") %></p>
<% end %>
```

Otherwise it leaks into the claims service:
Screenshot 2025-07-02 at 16 21 49

richardpattinson: Good spot; cheers!

JamieCleare2525 (approved):

PR #1695, opened 2025-07-08
Fix claim window banner spec

richardpattinson:
## Context

Claim window banner spec needs a start date as well as an end date; otherwise validation error is created when instantiating the claim_window object.


spec/components/claims/claim_window_warning_banner_component_spec.rb
JamieCleare2525, line 1: Could you swap these out for specific dates?
`let(:ends_on) { 2.months.from_now }`

JamieCleare2525 (approved):

PR #1720, opened 2025-07-24
Show related activity logs history on the show claim page for support users

richardpattinson:
## Context

Add activity log to claims so that a support user can see the history related to a claim.

When a claim is viewed in the “All Claims”, “Payments”, “Auditing” or “Clawback” tabs we should add a new “history” section at the bottom of the claim details page to display all of the audit logs associated to that claim so that there is an easy way to view the history of that claim and the related files


app/controllers/claims/support/claims/payments/claims_controller.rb
```diff
@@ -4,7 +4,12 @@ class Claims::Support::Claims::Payments::ClaimsController < Claims::Support::App
   before_action :set_claim, only: %i[show confirm_information_sent information_sent confirm_paid paid confirm_reject reject]
   before_action :authorize_claim
 
-  def show; end
+  def show
+    claim_activities = Claims::Claim::FindActivitiesForClaim.call(claim: @claim).order(:created_at)
```
Kizr, line 8: Would the `.order` be better placed inside the service?

app/services/claims/claim/find_activities_for_claim.rb
```diff
@@ -0,0 +1,34 @@
+class Claims::Claim::FindActivitiesForClaim < ApplicationService
```
Kizr, line 1: Would this be more suited to a `Query` rather than a `Service`?

app/services/claims/claim/find_activities_for_claim.rb
```diff
@@ -0,0 +1,34 @@
...
+        record_id: Claims::Clawback.joins(:clawback_claims)
+                                    .where(clawback_claims: { claim_id: claim.id })
+                                    .select(:id),
+      ),
+    ).or(
+      Claims::ClaimActivity.where(
+        record_type: "Claims::Payment",
+        record_id: Claims::Payment.joins(:payment_claims)
+                                  .where(payment_claims: { claim_id: claim.id })
+                                  .select(:id),
```
Kizr, line 30: Could move each of these scopes into a method that can provider a bit of context as to what we're fetching with the queries?

app/views/claims/support/claims/_history.html.erb
```diff
@@ -0,0 +1,31 @@
...
+        </div>
+        <% end %>
+    </div>
+
+    <%= render PaginationComponent.new(pagy: @pagy) %>
+    <% else %>
+    <p class="govuk-body"><%= t(".no_activity") %></p>
+    <% end %>
+</div>
+</div>
```
Kizr, line 31: Is the indentation off in this file?

config/locales/en/claims/support/claims.yml
```diff
@@ -42,6 +42,14 @@ en:
           mentor_with_index: Mentor %{index}
           mentor: Mentor
           submitted_by: Submitted by %{name} on %{date}.
+          no_activity: There is no activity.
```
Kizr, line 45: Should this be here?

spec/system/claims/support/claims/clawbacks/support_user_views_clawbacks_show_page_spec.rb
```diff
@@ -169,7 +182,7 @@ def then_i_see_the_show_page_for_the_clawback_in_progress_claim
     expect(primary_navigation).to have_current_item("Claims")
     expect(page).to have_element(:span, text: "Clawbacks - Claim 33333333", class: "govuk-caption-l")
     expect(page).to have_h1(@clawback_in_progress_claim.school_name)
-    expect(page).to have_element(:strong, text: "Clawback in progress", class: "govuk-tag govuk-tag--orange")
+    # expect(page).to have_element(:strong, text: "Clawback in progress", class: "govuk-tag govuk-tag--orange")
```
Kizr, line 185: Should this be commented out?

Gemfile.lock
```diff
@@ -708,6 +708,7 @@ PLATFORMS
   arm64-darwin
   x86_64-darwin-22
   x86_64-darwin-23
+  x86_64-darwin-24
```
Kizr, line 711: Does this belong in this PR?

spec/system/claims/support/claims/payments/view_a_claim_spec.rb
```diff
@@ -42,5 +42,6 @@ def then_i_can_see_the_details_of_claim(claim)
     within(".govuk-inset-text") do
       expect(page).to have_content("Some reason")
     end
+    expect(page).to have_content("History")
```
Kizr, line 45: Can we use a more specific matcher here?

config/locales/en/claims/support/claims/payments/claims.yml
```diff
@@ -12,6 +12,12 @@ en:
               mentor_with_index: Mentor %{index}
               mentor: Mentor
               submitted_by: Submitted by %{name} on %{date}.
+            history:
+              no_activity: There is no activity.
```
Kizr, line 16: Could we give the user a bit more information about what to expect? Maybe something along the lines of "Claim activity will be shown here after the claim has been processed" possibly check with product?

Kizr (changes requested):
A few things to consider on this one, could we also add some screenshots please?

Kizr (changes requested):

Kizr (approved):

PR #1728, opened 2025-07-30
Enforce non-root user

vipin-dfe:
## Context
Enforce containers to run as non-root user

Update the application and worker pods to run as non-root users

## Guidance to review
- Application should run without any issues.
- Application runs as non-root user

## Link to Trello card
https://trello.com/c/YekUhaCn

### Checklist
- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Tested by running locally

### DIFF
--- a/terraform/application/application.tf
+++ b/terraform/application/application.tf
@@ -47,6 +47,8 @@ module "web_application" {
   docker_image               = var.docker_image
   enable_logit               = true
 
+  run_as_non_root = var.run_as_non_root
+
   send_traffic_to_maintenance_page = false
 }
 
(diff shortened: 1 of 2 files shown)


terraform/application/application.tf
```diff
@@ -47,6 +47,10 @@ module "web_application" {
   docker_image               = var.docker_image
   enable_logit               = true
 
+  run_as_user     = var.run_as_user
```
RMcVelia, line 50: we just need the run_as_non_root setting, not the run_as_user and run_as_group

RMcVelia (approved):

PR #1729, opened 2025-07-30
Remove email_address field from schools

richardpattinson:
## Context

The data is only by request and we can’t reasonably update it, we do not use this field for anything and as such it can be removed as it is creating support requests due to outdated informaton.

We store an email address against the School for the Claims/Placements services, remove this column. The email address for the school contact can be used in place. 

## Guidance to review

Check that school contact is the email that a school is using.

## Link to Trello card

https://trello.com/c/kElByRGT/60-remove-school-email-address


app/models/school.rb
```diff
@@ -154,9 +151,7 @@ def primary?
   end
 
   def email_addresses
-    return [] if email_address.nil?
-
-    [email_address]
+    Array(school_contact&.email_address).compact
   end
```
Kizr, line 155: I'd probably remove this as school contact is a placement specific function

db/migrate/20250730130850_remove_email_address_from_schools.rb
```diff
@@ -0,0 +1,10 @@
+class RemoveEmailAddressFromSchools < ActiveRecord::Migration[8.0]
+  def change
+    # This column must be ignored in the School model before deploying this migration
+    # class School < ApplicationRecord
+    #   self.ignored_columns += ["email_address"]
+    # end
+
+    safety_assured { remove_column :schools, :email_address, :string }
```
Kizr, line 8: We should always follow strong migration's advice for a production environment, you can do the previous migration in another PR before this one is deployed

db/schema.rb
```diff
@@ -334,6 +334,20 @@
...
+    t.integer "status", default: 0, null: false
+    t.datetime "sent_at"
+    t.text "error"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["recipient"], name: "index_notify_email_queues_on_recipient"
+    t.index ["sent_at"], name: "index_notify_email_queues_on_sent_at"
+    t.index ["status"], name: "index_notify_email_queues_on_status"
+  end
+
```
Kizr, line 350: Does this belong here?

app/decorators/school_decorator.rb
```diff
@@ -1,4 +1,5 @@
 class SchoolDecorator < OrganisationDecorator
+  delegate :email_address, to: :school_contact, allow_nil: true
```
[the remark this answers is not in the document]
richardpattinson: This is already in the school model - I couldn't work out why It needed to be in the decorator to work, but if it is removed we get a nomethod error.

app/models/school.rb
Kizr: Then I think we need to change the details, this is a completely different value that we're returning and not one that we should be returning here.

app/models/school.rb
```diff
@@ -72,6 +72,7 @@
 #  fk_rails_...  (trust_id => trusts.id)
 #
 class School < ApplicationRecord
+  self.ignored_columns += %w[email_address]
```
Kizr, line 75: This needs to be added in a PR that precedes this one

Kizr (changes requested):

Kizr:
@richardpattinson this will be good to review after it's been rebased against main

PR #1731, opened 2025-08-04
Ignore email address column on schools

richardpattinson:
## Context

In order to remove the email address column we must first ignore it as per rails migration safety guidlines.


Kizr (commented):
As mentioned in the previous PR, the school email address and placement contact email address are not interchangable fields, we shoudn't substitute one for the other here.

The school email address should be removed with no replacement please

Kizr (approved):

PR #1732, opened 2025-08-07
Apply Dfe rebrand styling to Claim + Placements

richardpattinson:
## Context

The header component for government service has been updated and we need to update our services in line with this change by December.

Update the primary navigation component + header + images to adhere to new design.

## Guidance to review

Navigate around in the review app - check that the nav bar + logout links work as expected. Check that styling is appropriate.

## Link to Trello card

https://trello.com/c/Y4vPsoCh/37-update-claim-funding-header-to-align-to-new-design-system-standards

## Screenshots

Screenshot 2025-08-11 at 17 07 45


richardpattinson:
> I don't have any qualms with the general approach of the component but the setup is a bit complicated, if we can simplify the logic then I think it's a good approach.
> 
> image
> Given this screenshot, we need to remove the service name from the top blue section of the nav and probably flip the organisation switcher and service navigation around as well.
> 
> I think stick with the approach but refine it a bit more!

Added screenshots with the updated changes.

Kizr (approved):

PR #1734, opened 2025-08-11
Add claim window navigation

richardpattinson:
## Context

# School users

Currently when a school submits a claim for payment they lose sight of the claim after it has been sent off for payment to the ESFA and then all future corresponance happens via email.

Some of these emails direct the user back to the service but the claims they’re directed to are no longer visible due to the specific scopes we use on the claims index page.

The claims can be viewed when they’re linked to directly but there is no intuitive way for a user to find them naturally.

We need to add a way for our school users to view their historic claims and also monitor the state of their claims after they have been paid as sometimes they will need to be audited and clawed back. It would be useful for the school to have visibility of all of these steps as they are involved in these processes.

We should separate the historic claims from the current claims whilst taking the academic year into account.

# Support users

Support users can technically view historic claims via the “All claims” section of the support console, however all of the claims are bundled in together.

There is a desire to see all claims that are related to an organisation via their details as well as on the all claims page, however with the standardisation of the support console we will get this functonality for free once the school users needs have been addressed. The standardisation of the support console is tracked here:

Add toggle to claim page to show claims by claim window.

## Guidance to review

Use the claims page, view previous cycle claims.

## Link to Trello card

https://trello.com/c/2KFJeBCB/80-see-claims-from-previous-cycles


Kizr:
Please can you add screenshots?

richardpattinson:
> I think we need to change the message when claims are found as it may confusing now.
> 
> image
> This isn't quite true, there are no claims for the selected academic year, but there _are_ claims. What does the page look like when there are claims in the previous academic year?

Thats a good catch; I'll update the wording. Here is what it looks like when a previous claim window has a claim in it.

Screenshot 2025-08-15 at 12 56 43

app/views/claims/support/schools/claims/index.html.erb
```diff
@@ -0,0 +1,73 @@
...
+  <% elsif Claims::ClaimWindow.current.present? && !@school.eligible_for_claim_window?(Claims::ClaimWindow.current) %>
+    <%= govuk_warning_text do %>
+      <p class="govuk-body"><%= t(".not_eligible_html", support_email: mail_to(t("claims.support_email"), t("claims.support_email_html"), class: "govuk-link")) %>
+    <% end %>
+  <% else %>
+    <%= govuk_inset_text do %>
+      <p class="govuk-body"><%= t(".window_closed", start_year: Claims::ClaimWindow.previous.academic_year.starts_on.year, end_year: Claims::ClaimWindow.previous.academic_year.ends_on.year) %></p>
+      <p class="govuk-body"><%= sanitize t(".window_closed_guidance", link_to: govuk_link_to(t(".window_closed_link_text"))) %></p>
+    <% end %>
+  <% end %>
```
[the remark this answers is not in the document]
richardpattinson: Added

app/views/claims/schools/claims/index.html.erb
```diff
@@ -69,10 +68,14 @@
...
-  <% else %>
-    <p>
+  <% elsif @school.claims.blank? %>
+    <p class="govuk-body">
       <%= t("no_records_for", records: "claims", for: @school.name) %>
     </p>
+  <% else %>
+    <p class="govuk-body">
+      <%= t(".no_claims_for_selected_academic_year", school: @school.name, academic_year: @academic_year.name) %>
+    </p>
```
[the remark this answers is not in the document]
richardpattinson: Flipped

app/decorators/academic_year_decorator.rb
```diff
@@ -11,4 +11,13 @@ def display_name
       I18n.t("placements.academic_year.previous_academic_year", academic_year: academic_year.name)
     end
   end
+
+  def card_name
+    case academic_year
+    when AcademicYear.for_date(Date.current)
+      "#{academic_year.name} (current)"
```
Kizr, line 18:
```suggestion
      "#{academic_year.name} #{I18n.t("placements.academic_year.current")}"
```

Could we add a translation here?

app/views/claims/schools/claims/_claims_index_table.html.erb
```diff
@@ -0,0 +1,34 @@
...
+<%= app_table do |table| %>
+  <% table.with_caption(size: "m", text: caption) %>
+  <% table.with_head do |head| %>
+    <% head.with_row do |row| %>
+      <% row.with_cell(header: true, text: t(".table.headers.reference")) %>
+      <% row.with_cell(header: true, text: Claims::Claim.human_attribute_name(:accredited_provider)) %>
+      <% row.with_cell(header: true, text: Claims::Claim.human_attribute_name(:mentors)) %>
+      <% row.with_cell(header: true, text: t(".table.headers.claim_amount")) %>
+      <% row.with_cell(header: true, text: Claims::Claim.human_attribute_name(:submitted_at)) %>
+      <% row.with_cell(header: true, text: Claims::Claim.human_attribute_name(:status)) %>
```
Kizr, line 12: I've seen this `human_attribute_name` a few times, but I think it'd be nice to standardise these to use translations instead?

config/locales/en/claims/schools/claims.yml
```diff
@@ -40,6 +46,13 @@ en:
           window_closed_guidance_link_text: add a mentor
           window_closed_support: Email %{link_to} if you have a query about your claims.
           upcoming_window_guidance: A second claim window will open %{starts_on} and will close %{ends_on} to enable you to submit claims for any remaining mentor training for the %{academic_year_name} academic year.
+          no_claims_for_selected_academic_year: There are no claims for %{school} in the %{academic_year} academic year.
```
Kizr, line 49:
```suggestion
          no_claims_for_selected_academic_year: There are no claims for %{school_name} in the %{academic_year} academic year.
```

I was worried that this might output an object, could we rename this to match the input please?

spec/system/claims/schools/claims/claims_user_creates_a_claim_for_a_mentor_with_a_previous_year_claim_spec.rb
```diff
@@ -38,6 +41,8 @@ def given_a_school_exists_with_a_previous_year_claim
     end
     @current_claim_window = build(:claim_window, :current)
     @historic_claim_window = build(:claim_window, :historic)
+    @historic_academic_year_name = @historic_claim_window.academic_year.decorate.card_name
+    @current_academic_year_name = @current_claim_window.academic_year.name
```
Kizr, line 45: Should both of these use `card_name` for consistency?

spec/system/claims/schools/claims/claims_user_creates_a_claim_for_a_mentor_with_a_previous_year_claim_spec.rb
```diff
@@ -66,6 +71,18 @@ def and_i_am_signed_in
     sign_in_as(@user_anne)
   end
 
+  def then_i_see_the_claims_index_page_with_no_submitted_claims_for_this_year
+    expect(page).to have_title("Claims - Claim funding for mentor training - GOV.UK")
+    expect(primary_navigation).to have_current_item("Claims")
+    expect(page).to have_h1("Claims")
+    expect(page).to have_link("Add claim", href: "/schools/#{@shelbyville_school.id}/claims/new")
+    expect(page).to have_paragraph("There are no claims for Shelbyville Elementary in the #{@current_academic_year_name} academic year.")
```
Kizr, line 79: We could specifically set the name here to expect a string here instead of `@current_academic_year_name`?

spec/system/claims/schools/claims/claims_user_creates_a_claim_for_a_mentor_with_a_previous_year_claim_spec.rb
```diff
@@ -66,6 +71,18 @@ def and_i_am_signed_in
...
+  end
+
+  def when_i_click_on_previous_academic_year
+    click_on(@historic_academic_year_name)
```
Kizr, line 83: We could specifically set the name here to expect a string here instead of @historic_academic_year_name?

Kizr (changes requested):

app/controllers/claims/schools/claims_controller.rb
```diff
@@ -3,14 +3,19 @@ class Claims::Schools::ClaimsController < Claims::ApplicationController
...
+  before_action :set_academic_year, only: :index
   before_action :authorize_claim
 
   helper_method :edit_attribute_path
 
   def index
-    @pagy, @claims = pagy(
-      @school.claims.includes(:provider, :mentor_trainings).active.order_created_at_desc,
-    )
+    @current_academic_year_selected = set_academic_year == current_academic_year
```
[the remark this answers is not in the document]
richardpattinson: I dont think there is a current? method on academic year anymore, I could add it back in - but I think in this case it is probably ok to leave a current_academic_year method in the controller (and we can use it in the set_academic_year method).

Kizr (approved):

PR #1741, opened 2025-08-14
Jc/118/prototype clawback approval flow

JamieCleare2525:

### DIFF
--- a/config/locales/en/components/claims/support/claim/actions_component.yml
+++ b/config/locales/en/components/claims/support/claim/actions_component.yml
@@ -12,4 +12,4 @@ en:
               information_sent: Confirm information sent
               paid: Confirm claim paid
               reject: Reject claim
-              request_approval: Request approval
+              approve_clawback: Approve clawback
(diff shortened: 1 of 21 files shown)


config/locales/en/components/claims/support/claim/actions_component.yml
```diff
@@ -12,3 +12,4 @@ en:
               information_sent: Confirm information sent
               paid: Confirm claim paid
               reject: Reject claim
+              review_clawback: Review clawback
```
Kizr, line 15: The only thing I spotted in this beautiful PR is that button says "Review clawback" and the page says "Approve clawback" which goes against what we normally do with matching button actions to page headings, it's a minor thing though!

Kizr (approved):

config/locales/en/components/claims/support/claim/actions_component.yml
JamieCleare2525: Resolved in https://github.com/DFE-Digital/itt-mentor-services/pull/1741/commits/caff456ac967fb2a612c67f6fddc32433220a6c3

PR #1766, opened 2025-09-08
Fix academic year display

Kizr:
## Context

The service calculates the current academic year by date, however we're still processing claims from the previous academic year leading to support and end users not being able to see the claims they've made without changing the tab/selecting a filter.

- Adds a new method to calculate the academic year based off of the current or previous claim window and if neither exists, falls back to the current academic year by date.

## Guidance to review

- Log in as Anne and create a claim, you should see the claim on the index page
- Log in as Colin and view the claims tab, you should see Anne's claim

## Link to Trello card

https://trello.com/c/30SRrWPX/190-fix-academic-year-display-issue-for-users-and-support

## Screenshots

image
image

### DIFF
--- a/spec/models/academic_year_spec.rb
+++ b/spec/models/academic_year_spec.rb
@@ -180,6 +180,8 @@ RSpec.describe AcademicYear, type: :model do
                name: "2024 to 2025")
       end
 
+      before { previous_academic_year.destroy }
+
       it "creates a new academic year for the previous year" do
         expect { current_academic_year.previous }.to change(described_class, :count).by(1)
 


app/models/academic_year.rb
```diff
@@ -20,7 +20,7 @@ class AcademicYear < ApplicationRecord
 
   scope :order_by_date, -> { order(starts_on: :asc) }
   scope :order_by_date_desc, -> { order(starts_on: :desc) }
-  scope :before_date, ->(date) { where("starts_on <= ?", for_date(date).starts_on) }
+  scope :before_date, ->(date) { where("academic_years.starts_on <= ?", for_date(date).starts_on) }
```
JamieCleare2525, line 23: Isn't the `academic_years` redundant here?

JamieCleare2525 (approved):

PR #1888, opened 2026-03-12
Update claim hours

Kizr:
## Context

With the upcoming academic year the claim hours are changing, however we still need to be able to calculate claims for previous years, therefore we must add variable claim hours.

- Abstracts the training hours into their own class
- Updates the training allowance class to utilise the new training hours logic
- Updates the grant funding guidance to match the upcoming academic year
- Updates user facing text to use dynamic data from the new training hours class

## Link to Trello card

https://trello.com/c/8Ntbut3o/321-updating-logic-for-hours-you-can-claim-for-initial-and-refresher-training

### DIFF
--- a/app/models/claims/training_allowance.rb
+++ b/app/models/claims/training_allowance.rb
@@ -11,7 +11,7 @@ class Claims::TrainingAllowance
   end
 
   def total_hours
-    training_type == INITIAL_TRAINING_TYPE ? INITIAL_TRAINING_HOURS : REFRESHER_TRAINING_HOURS
+    training_type == INITIAL_TRAINING_TYPE ? initial_training_hours : refresher_training_hours
   end
 
   def remaining_hours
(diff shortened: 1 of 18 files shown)


app/models/claims/training_allowance.rb
```diff
@@ -27,12 +27,9 @@ def available?
   attr_reader :mentor, :provider, :academic_year, :claim_to_exclude
 
   INITIAL_TRAINING_TYPE = :initial
-  INITIAL_TRAINING_HOURS = 20
   REFRESHER_TRAINING_TYPE = :refresher
```
JamieCleare2525: These are replicated in your TrainingHours model. Do you need them both or can one be removed?

app/models/claims/training_hours.rb
```diff
@@ -0,0 +1,45 @@
...
+    hours_for(training_type: INITIAL_TRAINING_TYPE)
+  end
+
+  def refresher
+    hours_for(training_type: REFRESHER_TRAINING_TYPE)
+  end
+
+  def hours_for(training_type:)
+    HOURS_BY_PERIOD.fetch(training_type.to_sym).fetch(period)
+  end
```
JamieCleare2525, line 36: Could this be a private method?

Kizr: I use this method outside of this class :)

JamieCleare2525 (approved):

